PR #31, opened 2018-03-02
Run integration tests in kokoro

priyawadhwa:
I put the kokoro stuff in an if statement so that this could still be run locally with 'make integration-test'

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -13,7 +13,14 @@
 # limitations under the License.
 
 #!/bin/bash
-set -e
+set -ex
+
+if [ -f "$KOKORO_GFILE_DIR"/common.sh ]; then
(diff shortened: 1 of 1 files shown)


integration-test.sh
```diff
@@ -13,7 +13,14 @@
 # limitations under the License.
 
 #!/bin/bash
-set -e
+set -ex
+
+if [ -f "$KOKORO_GFILE_DIR"/common.sh ]; then
+    echo "Installing dependencies..."
+    source "$KOKORO_GFILE_DIR/common.sh"
+    pushd github/k8s-container-builder
```
sharifelgamal: switch this to github/kaniko and everything should be good to go

PR #94, opened 2018-04-16
Allow snapshot of kaniko certs so kaniko can build kaniko

priyawadhwa:
also fixed quick bug from adding a default path to the build context

### DIFF
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -57,8 +57,10 @@ func ExtractFileSystemFromImage(img string) error {
 
 // PathInWhitelist returns true if the path is whitelisted
 func PathInWhitelist(path, directory string) bool {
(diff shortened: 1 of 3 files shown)


pkg/util/fs_util.go
```diff
@@ -57,7 +57,7 @@ func ExtractFileSystemFromImage(img string) error {
 
 // PathInWhitelist returns true if the path is whitelisted
 func PathInWhitelist(path, directory string) bool {
-	if path == constants.KanikoExecutor {
+	if path == constants.KanikoExecutor || path == constants.KanikoCerts {
```
sharifelgamal: Do we want to just have a list of "privileged paths" in constants, in case we ever have to add any more?

priyawadhwa: for sure, just added it!

PR #97, opened 2018-04-16
Omit debugging symbols for maximum small bins

dekkagaijin:


dlorenc:
@dekkagaijin could you do an amend/force push to retrigger tests?

dekkagaijin:
think that creating the patch in GitHub before the repo rename broke my ability to rebase. I'll add another commit to hopefully fix things...

PR #113, opened 2018-04-17
Set the docker default seccomp profile in the executor process.

dlorenc:


r2d4:
Does this technically prevent users from specifying a more restrictive seccomp profile at the pod level? Since this would technically be trying to "elevate" permissions then

dlorenc:
> Does this technically prevent users from specifying a more restrictive seccomp profile at the pod level? Since this would technically be trying to "elevate" permissions then

Yeah, I think that would fail. That's why I added the --use-seccomp flag.

sanmai-NL:
@dlorenc: What prevents this from being merged at this time?

tejal29:
@dlorenc Are you still interested in merging this in?

tejal29:
closing due to inactivity.

PR #126, opened 2018-04-20
Integration test refactoring

sharifelgamal:


integration/integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func TestRun(t *testing.T) {
+	dockerfiles, err := ioutil.ReadDir(dockerfilesPath)
+	if err != nil {
+		fmt.Printf("err=%s", err)
+		t.FailNow()
+	}
+
+	_, ex, _, _ := runtime.Caller(0)
```
dlorenc: This is a bit weird, could you use os.Args[0] instead?
https://golang.org/pkg/os/

integration/integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+	cwd := filepath.Dir(ex)
+
+	// Grab the latest container-diff binary
+	getContainerDiff := exec.Command("gsutil", "cp", "gs://container-diff/latest/container-diff-linux-amd64", ".")
```
dlorenc: Let's just require that container-diff is on the path.

integration/integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	containerDiffPerms := exec.Command("chmod", "+x", "container-diff-linux-amd64")
+	err = containerDiffPerms.Run()
+	if err != nil {
+		t.Error(err)
+		t.FailNow()
+	}
+
+
+	for _, dockerfile := range dockerfiles {
```
dlorenc: filepath.Glob might be easier than ls'ing everything and filtering here.

integration/integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+			kanikoCmd := exec.Command("docker", "run",
+				"-v", os.Getenv("HOME")+"/.config/gcloud:/root/.config/gcloud",
+				"-v", cwd + ":/workspace",
+				executorImage,
+				"-f", path.Join(buildContextPath, dockerfilesPath, dockerfile.Name()),
+				"-d", kanikoImage,
+				"-c", buildContextPath,
+			)
+
+			err = kanikoCmd.Run()
```
dlorenc: nit: a small helper that takes the cmd and t and does this could be useful.

integration/integration_test.go
sharifelgamal: os.Args gives me where the test file is actually being run, which is in a tmp folder somewhere. I actually want to access the directory where the go file actually lives, which is what runtime.Caller gives.

integration/integration_test.go
sharifelgamal: Done

integration/integration_test.go
sharifelgamal: yep. done.

integration/integration_test.go
sharifelgamal: done.

dlorenc:
Looks like we'll have to move the container-diff installation piece to the kokoro setup (or the container that will run in GCB)

sharifelgamal:
This is ready for a look.

integration/integration_test.go
```diff
@@ -0,0 +1,194 @@
...
+	buildKaniko := exec.Command("docker", "build", "-t", executorImage, "-f", "../deploy/Dockerfile", "..")
+	err := buildKaniko.Run()
+	if err != nil {
+		fmt.Print(err)
+		fmt.Print("Building kaniko failed.")
+		os.Exit(1)
+	}
+
+	// Make sure container-diff is on user's PATH
+	err = exec.Command("container-diff").Run()
```
dlorenc: nit: exec.LookPath would probably be a bit easier here.

integration/integration_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	bucketContextTests := []string{"Dockerfile_test_copy_bucket"}
+
+	_, ex, _, _ := runtime.Caller(0)
+	cwd := filepath.Dir(ex)
+
+	for _, dockerfile := range dockerfiles {
+		t.Run("test_"+dockerfile, func(t *testing.T) {
+			dockerfile = dockerfile[len("dockerfile/")+1:]
+			fmt.Printf("%s\n", dockerfile)
```
dlorenc: nit: t.Log here instead of fmt.Print

integration/integration_test.go
```diff
@@ -0,0 +1,194 @@
...
+					"-d", kanikoImage,
+					contextFlag, contextPath},
+					buildArgs...)...,
+			)
+
+			RunCommand(kanikoCmd, t)
+
+			// container-diff
+			daemonDockerImage := daemonPrefix + dockerImage
+			//daemonKanikoImage := daemonPrefix + kanikoImage
```
dlorenc: remove this

integration/integration_test.go
sharifelgamal: Done

PR #128, opened 2018-04-20
Refactor copy and add

priyawadhwa:
Refactored copy/add to pull out the code for copying files and directories into helper functions, which will be useful for copying directories/files when we start using chroot for executing commands


priyawadhwa:
For sure, thanks @AkihiroSuda!

Issue #109, opened 2018-04-17
How to provide login credentials for a registry?


damoon:

dlorenc:
We should document this more clearly. Kaniko uses docker credential helpers right now, so anything that gets configured in it's container should work. Right now we only bundle the one for GCR, though.

carlossg:
I have some work in progress in #129

Issue #124, opened 2018-04-18
Document how to provide credentials


errm:
The README says
> configuring another credential helper should allow pushing to a different registry

It would be great to include an example

Boolman:
this works for me:

sudo docker run -v ${PWD}:${PWD} -v ${HOME}/.docker/config.json:/root/.docker/config.json boolman/kaniko:ubuntu -d gitlab.domain.local:4567/org/containername:tagname -c ${PWD}

cat $HOME/.docker/config.json
{
	"auths": {
		"gitlab.domain.local:4567": {
			"auth": "dXNlcm5hbWU6cGFzc3dvcmQ="
		},
	},
	"HttpHeaders": {
		"User-Agent": "Docker-Client/18.03.0-ce (linux)"
	}
}

carlossg:
Documented in #129

dlorenc:
Closing as a dup of #109.

PR #129, opened 2018-04-20
Use kaniko with docker config.json password

carlossg:
Allow standard docker `config.json` authentication using username and password, ie. for `hub.docker.com`.

Opening for reference and to start the discussion on how to add more Docker credential helpers

# Configuration

Get your docker registry user and password encoded in base64

    echo USER:PASSWORD | base64

Create a `config.json` file with your docker registry url and the previous generated base64 string

```
{
	"auths": {
		"https://index.docker.io/v1/": {
			"auth": "xxxxxxxxxxxxxxx"
		}
	}
}
```

Run kaniko

    docker run -ti --rm -v `pwd`:/workspace -v config.json:/root/.docker/config.json:ro csanchez/kaniko --dockerfile=Dockerfile --destination=yourimagename


deploy/Dockerfile
```diff
@@ -17,9 +17,7 @@
 FROM scratch
 ADD out/executor /kaniko/executor
 ADD files/ca-certificates.crt /kaniko/ssl/certs/
-ADD files/docker-credential-gcr /usr/local/bin/
-ADD files/config.json /root/.docker/
-RUN ["docker-credential-gcr", "config", "--token-source=env"]
+ADD files/docker-credential-pass /usr/local/bin/
```
dlorenc: Nice! We're actually only adding the gcr helper from a file here because it's still a special build we had to make to get it to run in this container. We're going to move it to a real release build soon. Would you mind using the ADD https:// format to add it without needing to check it in here?

carlossg:
Fixes #109 #124

carlossg:
I have an example running on Kubernetes with Jenkins using standard [`docker-registry` credentials](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-in-the-cluster-that-holds-your-authorization-token) at https://github.com/jenkinsci/kubernetes-plugin/pull/312

It would be something like 

``
kind: Pod
 metadata:
   name: kaniko
 spec:
   containers:
   - name: kaniko
     args: ["--dockerfile=<path to Dockerfile>",
            "--destination=<https://myregistry:50000/$PROJECT/$IMAGE:$TAG>"]
     imagePullPolicy: Always
     volumeMounts:
       - name: jenkins-docker-cfg
         mountPath: /root
   volumes:
     - name: jenkins-docker-cfg
       secret:
         secretName: regcred
```

**Comment by seeekr** [comment]
Hey! Wondering what the status is here, having kaniko push to a plain old docker registry is something I'd like to be able to do!

**Comment by everpeace** [comment]
public `gcr.io/kaniko-project/executor` doesn't support `config.json` ???

I confirmed my private repository works fine and my k8s cluster can pull my private image by `imagePullSecrets`.

```shell
$ kubectl create secret docker-registry myregistry-creds \
  --docker-username=___YOUR_USERNAME___ \
  --docker-password=___YOUR_PASSWORD___  \
  --docker-email=___YOUR_EMAIL___ \
  --docker-server=myregistry
 
$ kubectl get secret harbor-creds -ojsonpath="{.data['\.dockerconfigjson']}" | base64 -D | jq .
{
  "auths": {
    "myregistry": {
      "username": "___YOUR_USERNAME___",
      "password": "___YOUR_PASSWORD___",
      "email": "___YOUR_EMAIL___",
      "auth": "____base64_encoded_string_of_username:password____"
    }
  }
}

$ cat << EOT  > manifest.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kaniko-context
data:
  Dockerfile: |
    FROM ubuntu:latest
    RUN echo hello
---
apiVersion: v1
kind: Pod
metadata:
  name: kaniko
spec:
  containers:
  - name: kaniko
    args: ["--dockerfile=/context/Dockerfile",
           "--context=/context",
           "--destination=myregistry/my-user/test-kaniko:latest"]
    volumeMounts:
      - name: context
        mountPath: /context
      - name:  myregistry-creds
        mountPath: /root/.docker/
  restartPolicy: Never
  volumes:
    - name: harbor-creds
      secret:
        secretName: myregistry-creds
        items:
         - key: .dockerconfigjson
           path: config.json
    - name: context
      configMap:
        name: kaniko-context
EOT

$ kubectl create -f manifest.yaml

$ kubecl logs kaniko
time="2018-07-03T10:05:04Z" level=info msg="Not adding /var/lib/dpkg because it was added by a prior layer"
time="2018-07-03T10:05:04Z" level=info msg="Not adding /var/lib/dpkg/diversions because it was added by a prior layer"
time="2018-07-03T10:05:04Z" level=info msg="Not adding /var/run because it is whitelisted"
time="2018-07-03T10:05:04Z" level=info msg="Taking snapshot of full filesystem..."
time="2018-07-03T10:05:05Z" level=info msg="cmd: /bin/sh"
time="2018-07-03T10:05:05Z" level=info msg="args: [-c echo hello]"
hello
time="2018-07-03T10:05:05Z" level=info msg="Taking snapshot of full filesystem..."
time="2018-07-03T10:05:06Z" level=info msg="No files were changed, appending empty layer to config."
time="2018-07-03T10:05:06Z" level=error msg="UNAUTHORIZED: \"authentication required\""
```

carlossg:
Kaniko does support docker credentials, see documentation in this PR on how to use it, and example with Jenkins at https://github.com/jenkinsci/kubernetes-plugin/blob/master/examples/kaniko.groovy

everpeace:
@carlossg Thank you for the info.  After I looked the `kaniko.groovy`, I understood that mounting image pull secret which is created by `kubectl create secret docker-registry` command is the key.  Am I right??

Then, I can't understand why [my example above](https://github.com/GoogleContainerTools/kaniko/pull/129#issuecomment-402095349) doesn't work...

Would you mind guding me how to fix my example??? 🙇

everpeace:
@carlossg Figured it out.  As you mentioned, kaniko supports loading `config.json` via `go-containerregistry`.  However, `go-containerregistry` can't push a container image including layers which come from the different registry.  Currently, it works only for GCR, ECR or DockerHub unfortunately....

For example,

```dockerfile
FROM ubuntu:latest  # this layer comes from 'index.docker.io'
RUN echo hello
```

When trying to push an image built from the above `Dockerfile` to your private registry (say `myregistry`), then `go-containerregistry` fails because it tries to use [Cross Repository Blob Mount API](https://docs.docker.com/registry/spec/api/#cross-repository-blob-mount) even though the private registry doesn't have `libarary/ubuntu` repository.  Moreover,  I think it's dangerous because the `FROM` layer doesn't come from the private registry.

Thus, I created an issue https://github.com/google/go-containerregistry/pull/219 so that kaniko supports any private registries.

jonjohnsonjr:
> However, go-containerregistry can't push a container image including layers which come from the different registry. Currently, it probably works only for GCR, ECR or DockerHub unfortunately.

What makes you think that? We target schema 2 images for the registry API, so in theory it should work for any (modern) compliant registry. 😅

andreykaipov:
This is an old PR but I wanted to mention that I had to find it and https://docs.gitlab.com/ce/ci/docker/using_kaniko.html#building-a-docker-image-with-kaniko to figure out how to get Kaniko to successfully auth against our private registry.

Issue https://github.com/GoogleContainerTools/kaniko/issues/109 was closed by https://github.com/GoogleContainerTools/kaniko/pull/167, but that only added instructions for Amazon ECR, so I feel this PR is still appreciated. :-)

README.md
```diff
@@ -139,7 +139,29 @@ To run kaniko in Docker, run the following command:
...
+{
+	"auths": {
+		"https://index.docker.io/v1/": {
+			"auth": "xxxxxxxxxxxxxxx"
+		}
+	}
+}
+```
+
+Run kaniko with the `config.json` inside `/root/.docker/config.json`
```
andreykaipov: This should now be `/kaniko/.docker/config.json` because of https://github.com/GoogleContainerTools/kaniko/commit/e3f4dc479d03075d4eed7beb67be82c6e2e69f75

ghoben: Thanks for this comment @andrewrynhard . I had the same issue and pointing it to /kaniko/.docker/config.json solved NOT AUTHENTICATED issue.

tejal29: thanks!! done!!

PR #140, opened 2018-04-23
Switch from containers/image to go-containerregistry

dlorenc:


dlorenc:
This should be RFAL now.

PR #141, opened 2018-04-24
Support multi stage builds

### LINKED ISSUE #4: Allow multi-stage builds

priyawadhwa:
Fixes #4  and partially #9


pkg/dockerfile/dockerfile.go
```diff
@@ -37,6 +42,25 @@ func Parse(b []byte) ([]instructions.Stage, error) {
 	return stages, err
 }
 
+// ResolveStages resolves any calls to previous stages with names to indices
+// Ex. --from=second_stage should be --from=1 for easier processing later on
+func ResolveStages(stages []instructions.Stage) {
+	nameToIndex := make(map[string]string)
+	for i, stage := range stages {
+		index := strconv.Itoa(i)
+		nameToIndex[stage.Name] = index
```
dlorenc: it looks like the second one here overwrites the first one. Do we need both?

pkg/dockerfile/dockerfile.go
```diff
@@ -54,3 +78,47 @@ func ParseCommands(cmdArray []string) ([]instructions.Command, error) {
...
 }
+
+// Dependencies returns a list of files in this stage that will be needed in later stages
+func Dependencies(index int, stages []instructions.Stage) ([]string, error) {
+	var dependencies []string
+	for stageIndex, stage := range stages {
+		if stageIndex <= index {
+			continue
+		}
+		ms, err := image.NewSourceImage(stage.BaseName)
```
dlorenc: This will have to get rewritten a bit after my go-containerregistry PR goes in.

pkg/commands/copy.go
```diff
@@ -39,6 +40,11 @@ func (c *CopyCommand) ExecuteCommand(config *manifest.Schema2Config) error {
 	logrus.Infof("cmd: copy %s", srcs)
 	logrus.Infof("dest: %s", dest)
 
+	// Resolve from
+	if c.cmd.From != "" {
```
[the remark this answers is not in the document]
priyawadhwa: Sorry, what do you mean by randomness?

Right now, if "from" is specified as a previous stage, we change the build context to be the path to the directory where files from that stage are saved. That directory should remain unchanged even if multiple copy commands refer back to it

pkg/dockerfile/dockerfile.go
priyawadhwa: so if the stage has a name associated the second won't override the first, so I added both in just to make it easier to resolve '--from' from names to indexes later on

cmd/executor/cmd/root.go
```diff
@@ -23,23 +23,21 @@ import (
...
-	bucket                      string
-	dockerInsecureSkipTLSVerify bool
-	logLevel                    string
-	force                       bool
+	dockerfilePath string
+	destination    string
+	srcContext     string
+	snapshotMode   string
+	bucket         string
+	logLevel       string
```
dlorenc: looks like you dropped the dockerinsecuretlsverify in a merge conflict.

pkg/commands/copy.go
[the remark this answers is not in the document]
dlorenc: I was worried about a case like this:

```
FROM base1
RUN something

FROM base1
RUN somethingelse

FROM base2
COPY --from=0
```

But I guess we can't have conflicts here, because the users have to refer to them by index (or alias)

pkg/dockerfile/dockerfile.go
dlorenc: Ah ok. Maybe it would be more clear to have two separate maps?

dlorenc: I guess it would get confusing if someone tried to name a stage with an index...

pkg/dockerfile/dockerfile.go
```diff
@@ -54,3 +83,66 @@ func ParseCommands(cmdArray []string) ([]instructions.Command, error) {
...
+		}
+		imageConfig, err := sourceImage.ConfigFile()
+		if err != nil {
+			return nil, err
+		}
+		for _, cmd := range stage.Commands {
+			switch c := cmd.(type) {
+			case *instructions.EnvCommand:
+				envCommand := commands.NewEnvCommand(c)
+				if err := envCommand.ExecuteCommand(&imageConfig.Config); err != nil {
```
dlorenc: Why are we executing ENV commands in here? It seems like this shouldn't have any side effects.

cmd/executor/cmd/root.go
priyawadhwa: yah, it didn't look like it was doing anything now that we switched to go-containerregistry

pkg/dockerfile/dockerfile.go
priyawadhwa: so we need to keep track of which ENVs were declared before a copy command for a case like this:

```
FROM base1
RUN something

FROM base2
ENV file file
COPY --from=0 $file ./
```
So here I'm executing any EnvCommands on a temporary config for that stage, and that updated config is used to evaluate dependencies in any subsequent COPY commands

cmd/executor/cmd/root.go
priyawadhwa: (probably should have done that in a different commit)

pkg/dockerfile/dockerfile.go
priyawadhwa: Oh I see, that makes sense --  I changed it a bit to be less confusing with one map

priyawadhwa:
PTAL

PR #145, opened 2018-04-25
feat: Allow building an image without pushing

### LINKED ISSUE #144: Allow building without pushing
# Context
When using tooling like Gitlab-CI, we want to be able to run the build process pre-merge, in order to validate changes

# Problem
kaniko doesn't appear to allow you to run just the build without also needing credentials to push to a registry. 

# Solution
Allow for just the build process to run, without specifying a destination to push to.

mikebryant:
Fixes #144

### DIFF
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -109,6 +109,12 @@ func DoBuild(dockerfilePath, srcContext, destination, snapshotMode string, docke
 			}
 		}
 	}
+
+	if destination == "" {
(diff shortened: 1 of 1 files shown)


dlorenc:
Thanks for this! What do you think about adding a mode to output the image as a tarball on disk, instead of just skipping the push. That would serve the same use-case for you (you can use /dev/null or just ignore the tarball), and is something we'll probably need anyway.

PR #155, opened 2018-04-27
Switch to multi-stage dockerfile for building kaniko

priyawadhwa:


deploy/Dockerfile
```diff
@@ -14,8 +14,14 @@
 
 # Builds the static Go image to execute in a Kubernetes job
 
+FROM golang
+WORKDIR /go/src/github.com/GoogleContainerTools/kaniko
+RUN apt-get update && apt-get install -y make
```
r2d4: nit: pretty sure golang has make already?

deploy/Dockerfile
```diff
@@ -14,8 +14,14 @@
...
+COPY . .
+RUN make -B
```
r2d4: why `make -B`? that rebuilds everytime regardless of changes

deploy/Dockerfile
```diff
@@ -14,8 +14,14 @@
 
 # Builds the static Go image to execute in a Kubernetes job
 
+FROM golang
```
r2d4: nit: pin to a version here

PR #162, opened 2018-05-01
ARG Command

### LINKED ISSUE #14: Implement ARG command
See #1 for details

priyawadhwa:
Begins to fix #14 

To implement ARG, I modified the code [here](https://github.com/moby/moby/blob/3a633a712c8bbb863fe7e57ec132dd87a9c4eff7/builder/dockerfile/buildargs.go). I also created a BuildArgs type and passed it in to every command along with the config. Then, in each command, we get a list of ARG variables that aren't already defined in config.Env, and add them as values that can be used for environment replacement.

I didn't add functionality for specifying ARGs before FROM and replacing environment variables in FROM since this PR was getting big, so I was thinking I'd add that after this one goes in.


dlorenc:
> To implement ARG, I modified the code here. I also created a BuildArgs type and passed it in to every command along with the config. Then, in each command, we get a list of ARG variables that aren't already defined in config.Env, and add them as values that can be used for environment replacement.

Can you explain why you had to modify that code? It will be difficult to carry a modification to it long term, and the diff is a bit hard to read.

priyawadhwa:
None of the buildargs methods are exported, so I pretty much just copied over the parts I needed. Not great but I couldn't think of a better way to do it...

priyawadhwa:
For sure, it's exported now so I switched to vendoring it in.

The changes are still kind of hard to read, but basically I changed
`ExecuteCommand(config *v1.Config)`
to 
`ExecuteCommand(config *v1.Config, buildArgs *dockerfile.BuildArgs)`
for all the commands, and then wrote [ReplacementEnvs](https://github.com/priyawadhwa/kaniko/blob/6972bdbb165e1ac0206d78a776bf89efc4bde842/pkg/util/command_util.go#L230) to get the combined ENV/ARG keys for substitution.

It should be RFAL now

cmd/executor/cmd/root.go
```diff
@@ -77,13 +79,33 @@ var RootCmd = &cobra.Command{
...
 		}
-		if err := executor.DoBuild(dockerfilePath, srcContext, destination, snapshotMode, dockerInsecureSkipTLSVerify); err != nil {
+		if err := executor.DoBuild(dockerfilePath, srcContext, destination, snapshotMode, dockerInsecureSkipTLSVerify, buildArgs); err != nil {
 			logrus.Error(err)
 			os.Exit(1)
 		}
 	},
 }
 
+type buildArg []string
```
dlorenc: nit: could you move this to another file in the same package?

pkg/commands/copy.go
```diff
@@ -33,15 +35,16 @@ type CopyCommand struct {
 	snapshotFiles []string
 }
 
-func (c *CopyCommand) ExecuteCommand(config *v1.Config) error {
+func (c *CopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.BuildArgs) error {
```
dlorenc: nit: it's generally a best practice to not expose types like dockerfile.BuildArgs that come from a package dependency. Could we wrap this type with one of our own, and pass that around everywhere?

priyawadhwa: Done!

PR #166, opened 2018-05-02
Add demo to README

priyawadhwa:


kittydoor:
Out of curiosity, what tools were used to produce this demo? It would be great to know so I can do the same for my projects whenever relevant! :)

PR #167, opened 2018-05-02
Add credential helper and documentation for pushing to Amazon ECR

### LINKED ISSUE #109: How to provide login credentials for a registry?

priyawadhwa:
Begins to fix #109, since there are probably more credential helpers we will want to add

Built off of #165


deploy/Dockerfile
```diff
@@ -18,11 +18,19 @@ FROM golang:1.10
 WORKDIR /go/src/github.com/GoogleContainerTools/kaniko
 COPY . .
 RUN make
+WORKDIR /usr/local/bin
```
dlorenc: instead of changing workdir repeatedly can we ADD these into the right place?

priyawadhwa: done!

deploy/Dockerfile
```diff
@@ -18,13 +18,12 @@ FROM golang:1.10
 WORKDIR /go/src/github.com/GoogleContainerTools/kaniko
 COPY . .
 RUN make
-WORKDIR /usr/local/bin
-ADD https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v1.4.3-static/docker-credential-gcr_linux_amd64-1.4.3.tar.gz .
-RUN tar -xvzf /usr/local/bin/docker-credential-gcr_linux_amd64-1.4.3.tar.gz
+# Get GCR credential helper
+ADD https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v1.4.3-static/docker-credential-gcr_linux_amd64-1.4.3.tar.gz /usr/local/bin/
```
dlorenc: Hmm, shouldn't adding the tar also handle the unpacking below? It would be nice to unpack in the same step we add, to avoid having the .tar.gz in the final image.

priyawadhwa: [ADD](https://docs.docker.com/engine/reference/builder/#add) only unpacks local tar archives, so we need to unpack in the next step. The .tar.gz shouldn't end up in the final image though, because I only copy over the unpacked binary in the next stage

dlorenc: lol... thanks for explaining

PR #171, opened 2018-05-07
Create debug image with busybox shell

### LINKED ISSUE #156: Use busybox for kaniko debug image

priyawadhwa:
Fixes #156

### DIFF
--- a/deploy/executor-release.yaml
+++ b/deploy/executor-release.yaml
@@ -3,15 +3,7 @@ steps:
   - name: "gcr.io/cloud-builders/docker"
     args: ["build", "-f", "deploy/Dockerfile",
            "-t", "gcr.io/kaniko-project/executor:${COMMIT_SHA}", "."]
-  # Then, we want to clone the distroless repo and build busybox.tar
-  - name: "gcr.io/cloud-builders/git"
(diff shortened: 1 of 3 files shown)


deploy/executor-release.yaml
```diff
@@ -3,4 +3,17 @@ steps:
   - name: "gcr.io/cloud-builders/docker"
     args: ["build", "-f", "deploy/Dockerfile",
            "-t", "gcr.io/kaniko-project/executor:${COMMIT_SHA}", "."]
-images: ["gcr.io/kaniko-project/executor:${COMMIT_SHA}"]
+  # Then, we want to clone the distroless repo and build busybox.tar
+  - name: "gcr.io/cloud-builders/git"
+    args: ["clone", "https://github.com/GoogleContainerTools/distroless.git"]
+  - name: "gcr.io/cloud-builders/bazel"
+    args: ["build", "busybox:busybox_tar"]
```
dlorenc: Hmm, could we do this in another stage of the dockerfile? otherwise we won't be able to build locally.

priyawadhwa: Oh yah that makes sense, just added it!

PR #173, opened 2018-05-08
make destination required to fix "could not parse reference" errors

### LINKED ISSUE #170: Build of Images fails with unclear message
I tried to build the attached Dockerfile using kaniko, but the build failed with the following error.

```
time="2018-05-07T13:41:41Z" level=error msg="could not parse reference"
```

Attached you'll find the [build.log](https://github.com/GoogleContainerTools/kaniko/files/1980062/build.log) and the [Dockerfile](https://github.com/GoogleContainerTools/kaniko/files/1980065/Dockerfile.txt).

nbyl:
When `--destination` is missing on the command line, an empty string will be passed to the parser. This will result a non-descrtive error message. This PR make the argument required which fixes #170.


nbyl:
I signed it!

Issue #17, opened 2018-02-15
Implement HEALTHCHECK command


priyawadhwa:
See #1 for details

chrisz100:
Healthcheck requires a different config struct:
https://github.com/moby/moby/blob/master/api/types/container/config.go

It seems to be not included in the open container specs - thus implementing this would require completely replacing the base for configs. 

After that - if I get the code right - it would simply mean passing the already prepared HealthConfig struct to the config to persist it. Which shouldn't be hard at all - but that requires HealthCheck to be existent in the container config. 

Ideas?

chrisz100:
@dlorenc @priyawadhwa any idea?

priyawadhwa:
Hey, sorry for the late response! We use the [go-containerregistry](https://github.com/google/go-containerregistry) library for working with our images, and right now the [config](https://github.com/google/go-containerregistry/blob/master/v1/config.go#L75) doesn't have a field for Healthcheck. You'd probably have to create the healthcheck config struct in go-containerregistry and add it to the config before starting to implement this command in kaniko.

chrisz100:
So you suggest contributing to go-containerregistry and using the merged version for proceeding here, correct?

dlorenc:
@chrisz100 correct! I can help you get that PR merged into go-containerregistry as well :)

PR #177, opened 2018-05-14
WIP: Feature/healthcheck signal

chrisz100:
Solves https://github.com/GoogleContainerTools/kaniko/issues/17


pkg/commands/commands.go
```diff
@@ -68,6 +68,8 @@ func GetCommand(cmd instructions.Command, buildcontext string) (DockerCommand, e
 		return &ArgCommand{cmd: c}, nil
 	case *instructions.ShellCommand:
 		return &ShellCommand{cmd: c}, nil
+	case *instructions.HealthCheckCommand:
+		return &HealthCheckCommand{cmd: c}, nil
```
dlorenc: From the build error, it looks like you'll have to translate from the dockerfile HealthCheck type to the go-containerregistry one.

chrisz100: exactly. Just looking into how to do this the most efficiently. 

Just creating a v1.HealthConfig and passing all values would work, but I wouldn't really regard this as future proof or clean tbh. Any alternatives you could think of?

chrisz100:
So... I don't really like this way of doing it but it works and there is no way to convert struct to struct more fail-safe/ autonomous that I am aware of.

dlorenc:
> So... I don't really like this way of doing it but it works and there is no way to convert struct to struct more fail-safe/ autonomous that I am aware of.

Hmm, if the structs are actually equivalent you might be able to cast them: https://beta.golang.org/doc/go1.8#language

That would then error if the types don't match which would probably be a good thing.

chrisz100:
This is the case as of now as well sadly: 
`config.Healthcheck = v1.HealthConfig(h.cmd.Health)`
```
pkg/commands/healthcheck.go:44:38: cannot convert h.cmd.Health (type *container.HealthConfig) to type "github.com/GoogleContainerTools/kaniko/vendor/github.com/google/go-containerregistry/v1".HealthConfig
make: *** [out/executor] Error 2
```

On go 1.10

dlorenc:
Hmm, that might be because one is a pointer and the other isn't. Could you give it one more try by dereferencing the h.cmd.Health before converting?

chrisz100:
Same thing: 
```
test := &h.cmd.Health
converted := v1.HealthConfig(test)

--> 
pkg/commands/healthcheck.go:45:30: cannot convert test (type **container.HealthConfig) to type "github.com/GoogleContainerTools/kaniko/vendor/github.com/google/go-containerregistry/v1".HealthConfig
make: *** [out/executor] Error 2
```

dlorenc:
Nice!

PR #184, opened 2018-05-16
Allow multiple destinations

jesusofsuburbia:
Hello everyone,

I had the requirement to push an image to multiple container registries.
Since the image is already built it would make sense to push it to more than one registry (instead of building the image again with a different destination).

I recycled and existing type for multiple args and added a simple for-loop.
The missing `os.Exit` inside the loop made sense to me (why stop when only one of multiple pushes fail)
Tested this in my fork.

You can now specify multiple `--destination` flags.
If the `tarpath` is set, multiple tar-files (one for each destination) will be generated.

Please note that I didn't touch the `README.md` in this pull request.

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -85,10 +85,9 @@ var RootCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		for _, destination := range destinations {
-			if err := executor.DoPush(ref, image, destination, tarPath); err != nil {
-				logrus.Error(err)
-			}
+		if err := executor.DoPush(ref, image, destinations, tarPath); err != nil {
+			logrus.Error(err)
+			os.Exit(1)
 		}
 
 	},
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -31,14 +31,14 @@ import (
 
 var (
 	dockerfilePath              string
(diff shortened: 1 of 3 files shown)


jesusofsuburbia:
I signed it!

cmd/executor/cmd/root.go
```diff
@@ -84,10 +84,13 @@ var RootCmd = &cobra.Command{
 			logrus.Error(err)
 			os.Exit(1)
 		}
-		if err := executor.DoPush(ref, image, destination, tarPath); err != nil {
-			logrus.Error(err)
-			os.Exit(1)
+
+		for _, destination := range destinations {
```
priyawadhwa: Could we instead pass the destinations into DoPush, and have the for loop there?

cmd/executor/cmd/root.go
```diff
@@ -84,10 +84,13 @@ var RootCmd = &cobra.Command{
...
+			if err := executor.DoPush(ref, image, destination, tarPath); err != nil {
+				logrus.Error(err)
+			}
```
priyawadhwa: We probably want to add the `os.Exit(1)` here!

jesusofsuburbia:
Tested this, should be fine

priyawadhwa:
Great, thanks!

PR #185, opened 2018-05-16
adding metadata tests back to integration tests

sharifelgamal:


integration/integration_test.go
```diff
@@ -93,16 +102,24 @@ func TestRun(t *testing.T) {
 			"file=context/foo",
 			"file3=context/b*",
 		},
+		"Dockerfile_test_multistage": {"file=/foo2"},
 	}
 
 	bucketContextTests := []string{"Dockerfile_test_copy_bucket"}
 
+	testsToIgnore := []string{"Dockerfile_test_user_run"}
```
priyawadhwa: Why are we ignoring this test?

pkg/executor/executor.go
```diff
@@ -99,6 +99,9 @@ func DoBuild(dockerfilePath, srcContext, snapshotMode string, args []string) (na
 			return nil, nil, err
 		}
 		imageConfig, err := sourceImage.ConfigFile()
+		if baseImage == constants.NoBaseImage {
+			imageConfig.Config.Env = []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}
```
priyawadhwa: Could we add this to the constants package?

priyawadhwa: Could we add this to the constants package?

integration/dockerfiles/Dockerfile_test_expose
```diff
@@ -0,0 +1,8 @@
+FROM gcr.io/google-appengine/debian9
+EXPOSE 80
+EXPOSE 81/udp
+ENV protocol tcp
+EXPOSE 82/$protocol
+ENV port 83
+EXPOSE $port/udp
+EXPOSE $port/$protocol
```
priyawadhwa: nit: newline

integration/dockerfiles/Dockerfile_test_user_run
```diff
@@ -17,3 +17,5 @@ RUN useradd testuser
 RUN groupadd testgroup
 USER testuser:testgroup
 RUN echo "hey" > /tmp/foo
+USER testuser:1001
+RUN echo "hey2" >> /tmp/foo
```
priyawadhwa: nit: newline

integration/integration_test.go
tejal29: just add the bug as a todo comment.

pkg/commands/user.go
```diff
@@ -47,43 +46,15 @@ func (r *UserCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
 		}
 	}
 
-	// Lookup by username
-	userObj, err := user.Lookup(userStr)
+	_, _, err = util.GetUserFromUsername(userStr, groupStr)
```
tejal29: Little bit confused here.. 
Not sure if its pertaining to this PR. 
But we are ignoring the resolved user and group and returning the input userStr:grpStr correct?
Is this command to just see if user and group can be resolved?

pkg/commands/user.go
```diff
@@ -47,43 +46,15 @@ func (r *UserCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
...
-
-	uid := userObj.Uid
-	if group != nil {
-		uid = uid + ":" + group.Gid
+		userStr = userStr + ":" + groupStr
 	}
-
-	logrus.Infof("Setting user to %s", uid)
-	config.User = uid
+	config.User = userStr
```
tejal29: Initially this was set to uid:gid.
Now its being set to userStr:groupStr.
Is this intended ?

pkg/executor/executor.go
sharifelgamal: Done

pkg/commands/user.go
sharifelgamal: Yes, that's how docker does it and we want to mimic what docker does as closely as possible.

Issue #191, opened 2018-05-21
ECR credential helper being removed with multi-stage dockerfiles


Rishka:
When running kaniko in gitlab ci, the ECR credential helper is not available after the first stage in a multi stage docker build. 

The image I am using for kaniko is slightly modified to get it to run in gitlab-ci:
```docker
FROM gcr.io/kaniko-project/executor:debug-v0.1.0
RUN ["ln", "-s", "/busybox/sh", "/usr/local/bin/sh"]
RUN ["mv", "/usr/local/bin/docker-credential-ecr-login", "/busybox/docker-credential-ecr-login"]
ENTRYPOINT [ "/busybox/sh" ]
```

then run kaniko like
`/kaniko/executor --dockerfile /path/to/Dockerfile --context /path/to/  --destination reg_id.dkr.ecr.us-east-1.amazonaws.com/other:latest`

where the Dockerfile is

```docker
FROM  reg_id.dkr.ecr.us-east-1.amazonaws.com/repo:latest as precompile

FROM reg_id.dkr.ecr.us-east-1.amazonaws.com/second-repo:latest
```

exits with 
```
time="2018-05-21T15:49:56Z" level=error msg="Get https://reg_id.dkr.ecr.us-east-1.amazonaws.com/second-repo:latest: exec: \"docker-credential-ecr-login\": executable file not found in $PATH"
OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "exec: \"sh\": executable file not found in $PATH": unknown
ERROR: Job failed: error executing remote command: command terminated with non-zero exit code: Error executing in Docker Container: 1
```

Both of these images are alpine based, I can provide more logs if needed

edit: The credential helper is also not present at the end of the build

priyawadhwa:
Now that #192 is merged, this should be fixed so I'll close the issue.

PR #192, opened 2018-05-21
Make sure necessary files aren't deleted for multistage builds

priyawadhwa:


Rishka:
Tested this out in my ci, this fixes #191

PR #195, opened 2018-05-27
Feature/contextsources

chrisz100:
Implementing a modular build context retrieval. 
This will allow to define various sources, be it file, gs, s3 or any other. 

Feel free to review on a commit-basis as this might be a bigger thing and correcting it all at once would be double the effort in the end.

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -18,7 +18,6 @@ package cmd
 
 import (
 	"errors"
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -135,7 +134,7 @@ func resolveSourceContext() error {
 	}
 	if bucket != "" {
 		if !strings.Contains(bucket, "://") {
-			srcContext = fmt.Sprintf("gs://%s", bucket)
+			srcContext = constants.GCSBuildContextPrefix + bucket
 		} else {
 			srcContext = bucket
 		}
@@ -145,12 +144,11 @@ func resolveSourceContext() error {
 	if err != nil {
 		return err
 	}
-
-	buildContextPath := constants.BuildContextDir
-	srcContext, err = contextExecutor.UnpackTarFromBuildContext(buildContextPath)
+	logrus.Debugf("Getting source context from %s", srcContext)
+	srcContext, err = contextExecutor.UnpackTarFromBuildContext()
 	if err != nil {
 		return err
 	}
-	logrus.Debugf("Unpacked tar from %s to path %s", srcContext, buildContextPath)
+	logrus.Debugf("Build context located at %s", srcContext)
 	return nil
 }
--- a/files/config.json
+++ b/files/config.json
@@ -8,6 +8,5 @@
 		"gcr.io": "gcr",
 		"gcr.kubernetes.io": "gcr",
 		"us.gcr.io": "gcr"
-	},
-	"credsStore": "ecr-login"
+	}
 }
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -19,6 +19,9 @@ package integration
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/GoogleContainerTools/kaniko/pkg/constants"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/daemon"
 	"math"
 	"os"
 	"os/exec"
@@ -27,9 +30,6 @@ import (
 	"runtime"
 	"strings"
 	"testing"
-
-	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/google/go-containerregistry/pkg/v1/daemon"
 
 	"github.com/GoogleContainerTools/kaniko/testutil"
 )
@@ -159,7 +159,7 @@ func TestRun(t *testing.T) {
 			for _, d := range bucketContextTests {
 				if d == dockerfile {
 					contextFlag = "-b"
-					contextPath = "gs://" + kanikoTestBucket
+					contextPath = constants.GCSBuildContextPrefix + kanikoTestBucket
 					break
 				}
 			}
--- a/pkg/buildcontext/buildcontext.go
+++ b/pkg/buildcontext/buildcontext.go
@@ -18,30 +18,29 @@ package buildcontext
 
 import (
 	"errors"
+	"github.com/GoogleContainerTools/kaniko/pkg/constants"
 	"strings"
 )
 
-var buildContextMap = map[string]BuildContext{
-	"gs://":  &GCS{},
-	"s3://":  &S3{},
-	"dir://": &Dir{},
-}
-
 // BuildContext unifies calls to download and unpack the build context.
 type BuildContext interface {
-	// Gets context.tar.gz from the build context and unpacks to the directory
-	UnpackTarFromBuildContext(directory string) (string, error)
-	SetContext(srcContext string)
+	// Unpacks a build context and returns the directory where it resides
+	UnpackTarFromBuildContext() (string, error)
 }
 
 // GetBuildContext parses srcContext for the prefix and returns related buildcontext
 // parser
 func GetBuildContext(srcContext string) (BuildContext, error) {
-	for prefix, bc := range buildContextMap {
-		if strings.HasPrefix(srcContext, prefix) {
-			bc.SetContext(strings.TrimPrefix(srcContext, prefix))
-			return bc, nil
-		}
+	split := strings.SplitAfter(srcContext, "://")
+	prefix := split[0]
+	context := split[1]
+	switch prefix {
+	case constants.GCSBuildContextPrefix:
+		return &GCS{context: context}, nil
+	case constants.S3BuildContextPrefix:
+		return &S3{context: context}, nil
+	case constants.LocalDirBuildContextPrefix:
+		return &Dir{context: context}, nil
 	}
-	return nil, errors.New("unknown build context prefix provided, please use on of the following: gs://, dir://, s3://")
+	return nil, errors.New("unknown build context prefix provided, please use one of the following: gs://, dir://, s3://")
 }
--- a/pkg/buildcontext/dir.go
+++ b/pkg/buildcontext/dir.go
@@ -22,10 +22,6 @@ type Dir struct {
 }
 
 // UnpackTarFromBuildContext just provides a directory with already extracted content
-func (f *Dir) UnpackTarFromBuildContext(directory string) (string, error) {
+func (f *Dir) UnpackTarFromBuildContext() (string, error) {
 	return f.context, nil
-}
-
-func (f *Dir) SetContext(srcContext string) {
-	f.context = srcContext
 }
--- a/pkg/buildcontext/gcs.go
+++ b/pkg/buildcontext/gcs.go
@@ -19,7 +19,6 @@ package buildcontext
 import (
 	"os"
 	"path/filepath"
-	"strings"
 
 	"cloud.google.com/go/storage"
 	"github.com/GoogleContainerTools/kaniko/pkg/constants"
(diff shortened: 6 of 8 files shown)


AkihiroSuda:
WDYT about using [rclone](https://rclone.org/)?
It supports Amazon Drive, Amazon S3, Backblaze B2, Box, Ceph, DigitalOcean Spaces, Dreamhost, Dropbox, FTP, Google Cloud Storage, Google Drive, HTTP, Hubic, IBM COS S3, Memset Memstore, Microsoft Azure Blob Storage, Microsoft OneDrive, Minio, Nextloud, OVH, Openstack Swift, Oracle Cloud Storage, OwnCloud, pCloud, put.io, QingStor, Rackspace Cloud Files, SFTP, Wasabi, WebDAV, and Yandex Disk.

chrisz100:
rclone seems to support a lot but wouldn’t that be a little much stuff to „just“ download and untar a file? 

In any case we’d probably try to compile it into the executor as kaniko tries to operate on a almost scratch-like container.

cmd/executor/cmd/root.go
```diff
@@ -110,21 +116,66 @@ func checkDockerfilePath() error {
...
+		response, err := svc.GetObject(input)
+		if err != nil {
+			return err
+		}
+		defer response.Body.Close()
+		body, err := ioutil.ReadAll(response.Body)
+		if err != nil {
+			return err
+		}
+		ioutil.WriteFile(tarPath, body, 0600)
```
chrisz100: As we have a reader and can use this to process the unTar, when moving this to util replace the local storing with direct decompression.

dlorenc:
> Sounds good, I’ll put the interface into the changes.
> 
> I just put it all to root to get started so I planned to move the extract anyhow.
> 
> I’d also propose to move the part figuring out what build module to use to that package as this prefix parser is a little out of context there if we have a buildcontext package.

Is this ready for another look?

chrisz100:
Feel free to give your input! I’ve been blocked with other things so no changes there from my side I’m afraid. Will continue working on this this week.

chrisz100:
@priyawadhwa @dlorenc can you have a quick look onto interface and the GC draft?

cmd/executor/cmd/root.go
```diff
@@ -110,21 +112,45 @@ func checkDockerfilePath() error {
 	return errors.New("please provide a valid path to a Dockerfile within the build context")
 }
 
-// resolveSourceContext unpacks the source context if it is a tar in a GCS bucket
+// resolveSourceContext unpacks the source context if it is a tar in a bucket
 // it resets srcContext to be the path to the unpacked build context within the image
 func resolveSourceContext() error {
 	if srcContext == "" && bucket == "" {
```
chrisz100: We need to decide what flag to use here - or introduce a more generic wording here.

pkg/buildcontext/gc.go
```diff
@@ -0,0 +1,31 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package buildcontext
+
+import "github.com/GoogleContainerTools/kaniko/pkg/util"
+
+// BuildContext unifies calls to download and unpack the build context.
+type GC struct {
```
priyawadhwa: nit: this should probably be GCS instead of GC

cmd/executor/cmd/root.go
priyawadhwa: I think we should keep both flags so that this change is backwards compatible. Let's make sure that either 
`--context` or `--bucket` is specified (and return an error if both or neither are), and pass whichever one is specified into `GetBuildContext()` (mentioned below)

Also, if we pass in `--bucket` with no prefix, assume a GCS bucket so that it's still backwards compatible :)

cmd/executor/cmd/root.go
```diff
@@ -110,21 +112,45 @@ func checkDockerfilePath() error {
...
-	logrus.Infof("Using GCS bucket %s as source context", bucket)
-	buildContextPath := constants.BuildContextDir
-	if err := util.UnpackTarFromGCSBucket(bucket, buildContextPath); err != nil {
+
+	// if no context is set, add default file context.tar.gz
+	if !strings.HasSuffix(bucket, ".tar.gz") {
+		bucket += "/" + constants.ContextTar
+	}
+
+	if strings.HasPrefix(bucket, "file://") {
```
priyawadhwa: nit: could we make this dir:// since there may be multiple files in a directory being used as the build context

cmd/executor/cmd/root.go
```diff
@@ -110,21 +112,45 @@ func checkDockerfilePath() error {
...
+	var contextExecutor buildcontext.BuildContext
+
+	if !strings.Contains(bucket, "://") {
 		return nil
 	}
-	logrus.Infof("Using GCS bucket %s as source context", bucket)
-	buildContextPath := constants.BuildContextDir
-	if err := util.UnpackTarFromGCSBucket(bucket, buildContextPath); err != nil {
+
+	// if no context is set, add default file context.tar.gz
```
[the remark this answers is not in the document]
chrisz100: Moving it to a function is good. Just unsure what string function to use with a map for prefixes. Will look into that tomorrow.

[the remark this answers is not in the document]
priyawadhwa: I was thinking something like this:

```golang
func GetBuildContext(srcContext string) (*BuildContext, error) {
   for prefix, bc := range buildContextMap {
        if strings.HasPrefix(srcContext, prefix) {
              return bc, nil
        }
   }
   return nil, fmt.Error("invalid prefix")
}
```

cmd/executor/cmd/root.go
chrisz100: When you look for a super fancy solution and it really is as simple as this.

pkg/buildcontext/gc.go
chrisz100: I know kops is using gc but I’ll check with the docs and such to make it match

cmd/executor/cmd/root.go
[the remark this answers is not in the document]
chrisz100: When you look for a super fancy solution and it really is as simple as this.

priyawadhwa:
Is this ready for review?

chrisz100:
From my side yes. Further improvements can also go to a new PR to keep them clean and small.

cmd/executor/cmd/root.go
```diff
@@ -110,21 +113,40 @@ func checkDockerfilePath() error {
...
+		return errors.New("please specify a path to the build context with the --context flag or a bucket with the --bucket flag")
 	}
-	if bucket == "" {
-		return nil
-	}
-	logrus.Infof("Using GCS bucket %s as source context", bucket)
+
 	buildContextPath := constants.BuildContextDir
-	if err := util.UnpackTarFromGCSBucket(bucket, buildContextPath); err != nil {
+	var contextExecutor buildcontext.BuildContext
```
priyawadhwa: Does this still need to be declared here?

cmd/executor/cmd/root.go
```diff
@@ -110,21 +113,40 @@ func checkDockerfilePath() error {
...
+
+	// if no prefix use Google Cloud Storage as default for backwards compability
+	if !strings.Contains(srcContext, "://") {
+		if bucket != "" {
+			srcContext = fmt.Sprintf("gs://%s", bucket)
+		} else if bucket == "" {
```
priyawadhwa: nit: this can just be `else`

pkg/buildcontext/file.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/GoogleContainerTools/kaniko/pkg/util"
+)
+
+// BuildContext unifies calls to download and unpack the build context.
+type File struct {
+}
+
+func (f *File) UnpackTarFromBuildContext(buildContext string, directory string) error {
+	srcContext := strings.TrimPrefix(buildContext, "file://")
```
priyawadhwa: I think you changed this to `dir://`? Also, we should probably change the name of the struct to reflect that!

cmd/executor/cmd/root.go
```diff
@@ -110,21 +113,40 @@ func checkDockerfilePath() error {
...
+			srcContext = fmt.Sprintf("dir://%s", srcContext)
+		}
+	}
+
+	// if no context is set, add default file context.tar.gz
+	if !strings.HasSuffix(srcContext, ".tar.gz") {
```
priyawadhwa: Since this should only happen when pulling from a GCS bucket, we should probably move this logic into `gcs.go`

cmd/executor/cmd/root.go
```diff
@@ -110,21 +113,40 @@ func checkDockerfilePath() error {
...
+		srcContext += "/" + constants.ContextTar
+	}
+
+	contextExecutor, err := buildcontext.GetBuildContext(srcContext)
```
priyawadhwa: If a user specifies the --bucket flag instead of --srcContext, how will that get passed in here?

chrisz100: The idea is the following: 
If you specify —context that goes with priority. 
If you don’t put a prefix in there or it’s not specified: 
If bucket is set: use the bucket value with gcs. 
Bucket is empty but no prefix is set: use the context value as directory. 

That way we keep it backwards compatible.

cmd/executor/cmd/root.go
chrisz100: I’d leave this here. That way people can rely on if they don’t specify an object path we go with the default behavior: use context.tar.gz from the bucket root. If you do specify a full path: we use that.

cmd/executor/cmd/root.go
chrisz100: No :)

pkg/buildcontext/file.go
chrisz100: Missed that piece there and I’ll go forward to rename the struct and file as well.

chrisz100:
Any other remarks?

cmd/executor/cmd/root.go
```diff
@@ -110,21 +113,34 @@ func checkDockerfilePath() error {
...
+			srcContext = fmt.Sprintf("dir://%s", srcContext)
+		}
+	}
+
+	contextExecutor, err := buildcontext.GetBuildContext(srcContext)
+	if err != nil {
+		return err
+	}
+
+	if err := contextExecutor.UnpackTarFromBuildContext(srcContext, buildContextPath); err != nil {
```
priyawadhwa: We need to remove the prefix before passing in `srcContext`. We should probably do that in `GetBuildContext` and store the value in each of the build context structs

cmd/executor/cmd/root.go
```diff
@@ -110,21 +113,34 @@ func checkDockerfilePath() error {
...
-	}
-	logrus.Infof("Using GCS bucket %s as source context", bucket)
+
 	buildContextPath := constants.BuildContextDir
-	if err := util.UnpackTarFromGCSBucket(bucket, buildContextPath); err != nil {
+
+	// if no prefix use Google Cloud Storage as default for backwards compability
+	if !strings.Contains(srcContext, "://") {
+		if bucket != "" {
+			srcContext = fmt.Sprintf("gs://%s", bucket)
```
priyawadhwa: This makes sense if there is no prefix, but what if the bucket is passed in as `--bucket=gs://something`? The srcContext will need to be assigned to that value somewhere.

priyawadhwa:
Hey @chrisz100, just wondering how this was going? I wanted to get this in ASAP so it would be included in the next release.

chrisz100:
I was on a conference last week so I’ll get it fixed this weekend :-)

chrisz100:
@priyawadhwa what do you think of this way of solving it? I'm not super happy with the nested if's on bucket flag resolving...

cmd/executor/cmd/root.go
```diff
@@ -110,21 +113,38 @@ func checkDockerfilePath() error {
 	return errors.New("please provide a valid path to a Dockerfile within the build context")
 }
 
-// resolveSourceContext unpacks the source context if it is a tar in a GCS bucket
+// resolveSourceContext unpacks the source context if it is a tar in a bucket
 // it resets srcContext to be the path to the unpacked build context within the image
 func resolveSourceContext() error {
```
priyawadhwa: Hmm yah I see what you mean about the if statements, let's do it this way:

```golang
func resolveSourceContext() error {
	if srcContext == "" && bucket == "" {
		return errors.New("please specify a path to the build context with the --context flag or a bucket with the --bucket flag")
	}
	if srcContext != "" && !strings.Contains(srcContext, "://") {
		return nil
	}
	if bucket != "" {
		if !strings.Contains(bucket, "://") {
			srcContext = fmt.Sprintf("gs://%s", bucket)
		} else {
			srcContext = bucket
		}
	}
	// if no prefix use Google Cloud Storage as default for backwards compability
	contextExecutor, err := buildcontext.GetBuildContext(srcContext)
	if err != nil {
		return err
	}

	buildContextPath := constants.BuildContextDir
	if err := contextExecutor.UnpackTarFromBuildContext(buildContextPath); err != nil {
		return err
	}
	logrus.Debugf("Unpacked tar from %s to path %s", srcContext, buildContextPath)
	srcContext = buildContextPath
	return nil
}
```

chrisz100:
Rebased against master real quick just to make sure.

priyawadhwa:
Hey, the kokoro error was because your `dep ensure` updated go-containerregistry and the integration tests needed to be updated to match. I just merged a PR for that, so if you rebase one more time it should (hopefully) work :)

chrisz100:
Let's see!

chrisz100:
Almost forgot to add the s3 prefix for the s3 implementation again... That PR is getting way too big.  Should be all stable now!

priyawadhwa:
Hey @chrisz100, I wanted to refactor this a bit before merging and update the integration tests -- would you mind if I pushed a couple commits?

priyawadhwa:
Thank you!

chrisz100:
lgtm

chrisz100:
Busy times as it seems. @priyawadhwa any update or outlook on getting this merged?

PR #204, opened 2018-06-13
Add --single-snapshot flag to snapshot once after the build

### LINKED ISSUE #203: Add snapshot mode to only snapshot once at the end of the build

priyawadhwa:
Fixes #203


integration/integration_test.go
```diff
@@ -106,6 +106,12 @@ func TestRun(t *testing.T) {
 		"Dockerfile_test_multistage": {"file=/foo2"},
 	}
 
+	// Map for additional flags
```
dlorenc: In the test somewhere it would be good to make sure there is only one additional layer. Any ideas?

priyawadhwa: I added a test to make sure docker built and kaniko built images have the same number of layers (except for specified offsets for the --single-snapshot ones)

But it looks like it found errors in some of the other tests so I'll try and find those bugs as well

dlorenc:
```shell
--- FAIL: TestLayers (96.89s)
    --- FAIL: TestLayers/test_layer_dockerfiles/Dockerfile_test_user_run (1.70s)
    	integration_test.go:232: Error response from daemon: reference does not exist
    --- FAIL: TestLayers/test_layer_dockerfiles/Dockerfile_test_volume (9.76s)
    	integration_test.go:232: incorrect offset between layers of gcr.io/kaniko-test/docker-dockerfile_test_volume and gcr.io/kaniko-test/kaniko-dockerfile_test_volume: expected 0 but got 2
    --- FAIL: TestLayers/test_layer_dockerfiles/Dockerfile_test_workdir (9.78s)
    	integration_test.go:232: incorrect offset between layers of gcr.io/kaniko-test/docker-dockerfile_test_workdir and gcr.io/kaniko-test/kaniko-dockerfile_test_workdir: expected 0 but got 1

```

Does this need a rebase to pass?

priyawadhwa:
So these tests are failing because the number of layers is actually incorrect, but I found a bug for the workdir one which I'm going to submit a PR for now

PR #205, opened 2018-06-18
adding reproducible flag

### LINKED ISSUE #120: Add option to strip timestamps for reproducible image configs
Expose a (date or bool) flag to allow users to choose to strip timestamps or set a particular one. We should also make sure the rest of the metadata is deterministic. 

https://github.com/moby/moby/pull/30611

sharifelgamal:
Fixes #120


sharifelgamal:
There's a bug in go-containerregistry that is causing the kokoro failure. It's fixed in https://github.com/google/go-containerregistry/pull/210.

dlorenc:
This should be ready to go after a dep update.

sharifelgamal:
This is (finally) ready.

PR #209, opened 2018-06-20
Track file ownership and use file ownership from base images

### LINKED ISSUE #208: kaniko does not consider file ownership change
Using a simple test Dockerfile

```
FROM alpine:3.7
RUN touch testfile
RUN chown 1000:0 testfile
RUN ls -la testfile
```

the resulting Docker image has root:root ownership of testfile instead of 1000:0.
As the final `ls` command shows, the chown command itself worked in the build container.

The file ownership change is not recorded (`No files were changed` message):
```
time="2018-06-20T07:49:45Z" level=info msg="Taking snapshot of full filesystem..."
time="2018-06-20T07:49:49Z" level=info msg="cmd: /bin/sh"
time="2018-06-20T07:49:49Z" level=info msg="args: [-c touch testfile]"
time="2018-06-20T07:49:49Z" level=info msg="Taking snapshot of full filesystem..."
time="2018-06-20T07:49:54Z" level=info msg="cmd: /bin/sh"
time="2018-06-20T07:49:54Z" level=info msg="args: [-c chown 1000:0 testfile]"
time="2018-06-20T07:49:54Z" level=info msg="Taking snapshot of full filesystem..."
time="2018-06-20T07:49:57Z" level=info msg="No files were changed, appending empty layer to config."
time="2018-06-20T07:49:57Z" level=info msg="cmd: /bin/sh"
time="2018-06-20T07:49:57Z" level=info msg="args: [-c ls -la testfile]"
-rw-r--r--    1 1000     root             0 Jun 20 07:49 testfile
time="2018-06-20T07:49:57Z" level=info msg="Taking snapshot of full filesystem..."
time="2018-06-20T07:50:01Z" level=info msg="No files were changed, appending empty layer to config."
```

xanonid:
Closes #208

### DIFF
--- a/pkg/util/bucket_util.go
+++ b/pkg/util/bucket_util.go
@@ -58,7 +58,7 @@ func getTarFromBucket(bucketName, directory string) (string, error) {
 	}
 	defer reader.Close()
 	tarPath := filepath.Join(directory, constants.ContextTar)
-	if err := CreateFile(tarPath, reader, 0600); err != nil {
+	if err := CreateFile(tarPath, reader, 0600, 0, 0); err != nil {
 		return "", err
 	}
 	logrus.Debugf("Copied tarball %s from GCS bucket %s to %s", constants.ContextTar, bucketName, tarPath)
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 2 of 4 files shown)


dlorenc:
Hmm, do you think a better fix might be to make sure the uid and gid are included in the snapshot?

Maybe somewhere in: https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/util/util.go#L46

I think we'd need to cast the fi.Sys() to grab the gid and uid, then the file change should be detected.

xanonid:
@dlorenc Do you have reviewed the latest state of this pull request? It should already include the proposed snapshot change. The other changes might also be required to improve uid/gid handling. I have not yet open issues for every bug I found with file ownership, but can do this if necessary...

dlorenc:
> @dlorenc Do you have reviewed the latest state of this pull request? It should already include the proposed snapshot change. The other changes might also be required to improve uid/gid handling. I have not yet open issues for every bug I found with file ownership, but can do this if necessary...

Ah sorry, I must have looked at a stale snapshot or something... This looks like the right fix to me. Thanks!

PR #228, opened 2018-07-06
make path to Dockerfile an absolute path

priyawadhwa:
Fixes a bug in which the relative path `Dockerfile` was found since we were in the correct working directory, but once we tried to open the file from root it wasn't found. 

Saving the path to the Dockerfile as an absolute path should fix this.

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -113,6 +113,11 @@ func checkContained() bool {
 
 func checkDockerfilePath() error {
 	if util.FilepathExists(dockerfilePath) {
+		abs, err := filepath.Abs(dockerfilePath)
+		if err != nil {
+			return err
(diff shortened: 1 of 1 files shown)


cmd/executor/cmd/root.go
```diff
@@ -113,6 +113,11 @@ func checkContained() bool {
 
 func checkDockerfilePath() error {
 	if util.FilepathExists(dockerfilePath) {
+		abs, err := filepath.Abs(dockerfilePath)
+		if err != nil {
+			return err
+		}
+		dockerfilePath = abs
```
dlorenc: Hmm, could we refactor this a bit to avoid mutating the global variable in here?

priyawadhwa: done

PR #242, opened 2018-07-18
Switch the valid URL to https://google.com.

dlorenc:
It appears sometimes Github goes down :)

This led to a couple of flakes in @bobcatfish's recent PRs.


bobcatfish:
Oh did that happen too :O? The issue I noticed was with new behavior from the invalid URL (fixed in #237).

PR #243, opened 2018-07-18
[WIP] Look for usable on-cluster credentials using k8schain

ImJasonH:
This updates deps and uses [`k8schain`](https://godoc.org/github.com/google/go-containerregistry/pkg/authn/k8schain) and [`NewMultiKeychain`](https://godoc.org/github.com/google/go-containerregistry/pkg/authn#NewMultiKeychain) to attempt to authorize pushes using available image pull secrets.

This change isn't ready to merge, I just wanted to get some initial feedback about the plan and maybe ideas about how I should go about testing this.


priyawadhwa:
It looks like you changed the auth for getting base images, but not for [pushing](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/executor.go#L212) at the end of the build. That might be the issue?

ImJasonH:
Well that would certainly help, wouldn't it 😄 

Unfortunately, I don't think that's the problem, since the failure seems to happen long before the image is built or pushed. The full logs from the failing container are:

```
$ kubectl logs build-pod -c kaniko
time="2018-07-19T01:10:02Z" level=info msg="appending to multi args gcr.io/jasonhall-kube/built-with-kaniko-fixed"
time="2018-07-19T01:10:02Z" level=info msg="Unpacking filesystem of golang:latest..."
time="2018-07-19T01:10:02Z" level=error msg="serviceaccounts \"default\" is forbidden: User \"system:serviceaccount:default:default\" cannot get serviceaccounts in the namespace \"default\": Unknown user \"system:serviceaccount:default:default\""
```

pkg/executor/executor.go
```diff
@@ -91,11 +90,12 @@ func DoBuild(k KanikoBuildArgs) (name.Reference, v1.Image, error) {
 			if err != nil {
 				return nil, nil, err
 			}
-			auth, err := authn.DefaultKeychain.Resolve(ref.Context().Registry)
+			k8sc, err := k8schain.NewInCluster(k8schain.Options{})
```
mattmoor: `NewInCluster` will use the pod's serviceaccount identity to try and talk to the API server and read secrets, you want [this](https://github.com/google/go-containerregistry/blob/1fcd4c5f05065afeab866058b03c688d0ab89157/pkg/authn/k8schain/k8schain.go#L121-L131) which I now realize isn't documented.  Oops.

ImJasonH: Oh okay. I'll give that a try. Sounds like we need better docs about which is useful in which scenarios, and maybe a rename, since I figured `NewInCluster` would be the one I wanted while trying to auth from in the cluster.

ImJasonH: That worked! 🎉

pkg/executor/executor.go
```diff
@@ -91,11 +90,12 @@ func DoBuild(k KanikoBuildArgs) (name.Reference, v1.Image, error) {
...
 				return nil, nil, err
 			}
-			auth, err := authn.DefaultKeychain.Resolve(ref.Context().Registry)
+			k8sc, err := k8schain.NewNoClient()
 			if err != nil {
 				return nil, nil, err
 			}
-			sourceImage, err = remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport))
+			kc := authn.NewMultiKeychain(authn.DefaultKeychain, k8sc)
+			sourceImage, err = remote.Image(ref, remote.WithAuthFromKeychain(kc), remote.WithTransport(http.DefaultTransport))
```
mattmoor: FYI `remote.WithTransport(http.DefaultTransport)` is a nop

ImJasonH:
@priyawadhwa Rebased and updated `image_util.go` to use k8schain, this should be RFAL.

mattmoor:
Very eager to see this land :)

PR #249, opened 2018-07-24
Fix spelling mistake in README

johnaoss:


bobcatfish:
Well that certainly doesn't seem like it should cause a test failure :O https://github.com/GoogleContainerTools/kaniko/issues/251

PR #255, opened 2018-07-25
Add --target flag for multistage builds

### LINKED ISSUE #253: Ability to save save/upload stage/built targets.
Hello,

Just thinking if it makes sense to expose then, as some times we want to build an specific stage.
https://docs.docker.com/engine/reference/commandline/build/#specifying-target-build-stage---target

So it maybe nice to be able to save/upload those images as well.
I think the stage builds are already saved so it would be just an way to expose that.

If tarPath is set, save all stages with names maybe saved there.

priyawadhwa:
Should fix #253

### DIFF
--- a/pkg/dockerfile/dockerfile.go
+++ b/pkg/dockerfile/dockerfile.go
@@ -18,13 +18,15 @@ package dockerfile
 
 import (
 	"bytes"
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"strings"
+
 	"github.com/GoogleContainerTools/kaniko/pkg/constants"
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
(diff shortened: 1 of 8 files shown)


pkg/dockerfile/dockerfile.go
```diff
@@ -40,6 +42,15 @@ func Parse(b []byte) ([]instructions.Stage, error) {
 	return stages, err
 }
 
+func Validate(stages []instructions.Stage, target string) error {
```
dlorenc: nit: this makes it sound like you're going to validate the whole Dockerfile, when all we're really checking is that the target exists. Maybe rename it to that?

priyawadhwa: done

Issue #251, opened 2018-07-24
Number of layers inconsistent with kaniko builds that copy/copy the same files multiple times


bobcatfish:
The integration test run for https://github.com/GoogleContainerTools/kaniko/pull/249 failed even though the PR is fixing a typo in a README.

Seems like we either have a flakey test or the test is broken! :O

```
++ go test
--- FAIL: TestLayers (49.18s)
    --- FAIL: TestLayers/test_layer_dockerfiles/Dockerfile_test_copy_reproducible (2.70s)
    	integration_test.go:250: incorrect offset between layers of gcr.io/kaniko-test/docker-dockerfile_test_copy_reproducible and gcr.io/kaniko-test/kaniko-dockerfile_test_copy_reproducible: expected 0 but got 1
FAIL
```

bobcatfish:
Looking back through kokoro history I can see a bunch of similar failures:

https://github.com/GoogleContainerTools/kaniko/pull/249:
```
++ go test
--- FAIL: TestLayers (49.18s)
    --- FAIL: TestLayers/test_layer_dockerfiles/Dockerfile_test_copy_reproducible (2.70s)
    	integration_test.go:250: incorrect offset between layers of gcr.io/kaniko-test/docker-dockerfile_test_copy_reproducible and gcr.io/kaniko-test/kaniko-dockerfile_test_copy_reproducible: expected 0 but got 1
```

https://github.com/GoogleContainerTools/kaniko/pull/243 (twice, both with same commit):
```
++ go test
--- FAIL: TestLayers (49.67s)
    --- FAIL: TestLayers/test_layer_dockerfiles/Dockerfile_test_copy (2.74s)
    	integration_test.go:250: incorrect offset between layers of gcr.io/kaniko-test/docker-dockerfile_test_copy and gcr.io/kaniko-test/kaniko-dockerfile_test_copy: expected 0 but got 1
```
```
++ go test
--- FAIL: TestLayers (48.30s)
    --- FAIL: TestLayers/test_layer_dockerfiles/Dockerfile_test_copy_reproducible (2.55s)
    	integration_test.go:250: incorrect offset between layers of gcr.io/kaniko-test/docker-dockerfile_test_copy_reproducible and gcr.io/kaniko-test/kaniko-dockerfile_test_copy_reproducible: expected 0 but got 1
```

https://github.com/GoogleContainerTools/kaniko/pull/244:
```
++ go test
--- FAIL: TestLayers (53.94s)
[...]

PR #256, opened 2018-07-25
Local integration tests

bobcatfish:
When starting to look into #251 I wanted to try to reproduce the test flakes locally but I immediately ran into two problems:
1. The tests hardcode a GCP project and bucket, which means I have to have access to those to run the tests.
2. The tests have to be triggered via `integration_tests.sh` to work, b/c it contains important setup, and this makes it a bit tough to run tests individually

(1) Can be resolved by adding me to these projects, but it turns out the tests can work just fine using a different project and bucket, so I've updated the tests such that these values can be provided as arguments to the tests.
(2) Now the setup (i.e. the creation of the build context) is done as part of the tests themselves, so we can run the tests directly with `go test`.

Bonuses:
* The tests previously relied on `TestRun` being run before `TestLayers` because `TestRun` would actually build the images with docker and kaniko - this means running tests from `TestLayers` individually did not work. Now the tests will check if the images have been built before trying to use them, build them if they haven't been built, or use the previously built images if they have been.
* There will now be no possibility of a collision if the tests are run simultaneously using the same GCS bucket b/c the context tarball name will include a UUID.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -70,12 +70,32 @@ _These tests will not run correctly unless you have [checked out your fork into 
 The integration tests live in [`integration`](./integration) and can be run with:
 
 ```shell
+export GCS_BUCKET="gs://<your bucket>"
+export IMAGE_REPO="gcr.io/somerepo"
 make integration-test
 ```
 
-_These tests require push access to a project in GCP, and so can only be run
-by maintainers who have access. These tests will be kicked off by [reviewers](#reviews)
-for submitted PRs._
+If you want to run `make integration-test`, you must override the project using environment variables:
+
+* `GCS_BUCKET` - The name of your GCS bucket
+* `IMAGE_REPO` - The path to your docker image repo
+
+You can also run tests with `go test`, for example to run tests individually:
+
+```shell
+go test -v --bucket $GCS_BUCKET --repo $IMAGE_REPO -run TestLayers/test_layer_Dockerfile_test_copy_bucket
+```
+
+Requirements:
+
+* [`gcloud`](https://cloud.google.com/sdk/install)
+* [`gsutil`](https://cloud.google.com/storage/docs/gsutil_install)
+* [`container-diff`](https://github.com/GoogleContainerTools/container-diff#installation)
+* A bucket in [GCS](https://cloud.google.com/storage/) which you have write access to via
+  the user currently logged into `gcloud`
+* An image repo which you have write access to via the user currently logged into `gcloud`
+
+These tests will be kicked off by [reviewers](#reviews) for submitted PRs.
 
 ## Creating a PR
 
--- a/integration/gcs.go
+++ b/integration/gcs.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+)
+
+// CreateIntegrationTarball will take the contents of the integration directory and write
+// them to a tarball in a temmporary dir. It will return a path to the tarball.
+func CreateIntegrationTarball() (string, error) {
+	log.Println("Creating tarball of integration test files to use as build context")
+	dir, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("Failed find path to integration dir: %s", err)
+	}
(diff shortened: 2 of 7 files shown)


bobcatfish:
kokoro is going to love this, I can just feel it

DEVELOPMENT.md
```diff
@@ -70,12 +70,40 @@ _These tests will not run correctly unless you have [checked out your fork into
...
+If you want to run these tests yourself, you must override the project using environment variables:
+
+* `GCS_BUCKET` - The name of your GCS bucket
+* `IMAGE_REPO` - The path to your docker image repo
+
+These tests will be kicked off by [reviewers](#reviews) for submitted PRs.
+
+#### Troubleshooting
+
+If you see errors due to `container-diff` failing, try pulling the base images (specified in [the
```
priyawadhwa: Just realized we could also get rid of this problem by using fully qualified images for all the Dockerfiles, so once [this](https://github.com/GoogleContainerTools/kaniko/pull/257) goes through hopefully we won't need this section!

DEVELOPMENT.md
```diff
@@ -70,12 +70,40 @@ _These tests will not run correctly unless you have [checked out your fork into
...
 
-_These tests require push access to a project in GCP, and so can only be run
-by maintainers who have access. These tests will be kicked off by [reviewers](#reviews)
-for submitted PRs._
+You can also run tests with `go test`, for example to run tests individually:
+
+```shell
+export GCS_BUCKET="gs://<your bucket>"
+export IMAGE_REPO="gcr.io/somerepo"
+go test -v --bucket $GCS_BUCKET --repo $IMAGE_REPO -run TestLayers/test_layer_Dockerfile_test_copy_bucket
```
priyawadhwa: Since we have the flags here anyway, it might be easier to just pass in the names instead of having env variables as an extra step?

integration/gcs.go
```diff
@@ -0,0 +1,73 @@
...
+	log.Println("Creating tarball of integration test files to use as build context")
+	dir, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("Failed find path to integration dir: %s", err)
+	}
+	tempDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return "", fmt.Errorf("Failed to create temporary directoy to hold tarball: %s", err)
+	}
+	uuid := randomString(32)
```
priyawadhwa: It might be simpler to use a timestamp here, since the context will be deleted anyway!

bobcatfish: oh good call!

DEVELOPMENT.md
bobcatfish: kk sounds good!

DEVELOPMENT.md
bobcatfish: ooooo nice! and it's in, so I'll remove this from the doc

integration/integration_test.go
```diff
@@ -221,47 +185,42 @@ func TestRun(t *testing.T) {
...
+			if err := checkLayers(t, dockerImage, kanikoImage, offset[dockerfile]); err != nil {
 				t.Error(err)
 				t.Fail()
 			}
 		})
 	}
 }
 
-func checkLayers(image1, image2 string, offset int) error {
+func checkLayers(t *testing.T, image1, image2 string, offset int) error {
```
sharifelgamal: Why do we need to pass t in here?

bobcatfish: oh thanks for catching that @sharifelgamal I'll remove it - I had tried out calling `t.Helper()` in this function to change the line number reported for failures, but I ended up thinking it's better reporting the line number inside `checkLayers`

DEVELOPMENT.md
bobcatfish: Okay so I'm going to do something halfway to what you are suggesting, just removing the export lines so it looks like this:
```
go test -v --bucket $GCS_BUCKET --repo $IMAGE_REPO -run TestLayers/test_layer_Dockerfile_test_copy_bucket`
```

One alternative would be to hardcode some defaults, but i think this makes it hard for ppl to know what they need to change:

```
go test -v --bucket gs://kaniko-test-bucket --repo gcr.io/kaniko-test -run TestLayers/test_layer_Dockerfile_test_copy_bucket`
```

Or we could do something like this, which is what I used to do all the time, but the thing is you can't copy and paste it:
```
go test -v --bucket <your bucket> --repo <your repo> -run TestLayers/test_layer_Dockerfile_test_copy_bucket`
```

I like leaving the env vars in b/c then if ppl use env vars they can copy & paste the command from the doc

bobcatfish:
PTAL @priyawadhwa @sharifelgamal

DEVELOPMENT.md
priyawadhwa: For sure, SGTM!

bobcatfish:
image

:sweat_smile:

priyawadhwa:
LGTM, I think you just need to rebase & this should be ready to merge!

bobcatfish:
> LGTM, I think you just need to rebase & this should be ready to merge!

HERE WE GO that flake better not happen, that would be ironic

<.<

dlorenc:
```shell
integration/images.go:68:1:warning: singleSnapshotImages is unused (deadcode)
integration/images.go:29:1:warning: singleSnapshotFlag is unused (deadcode)
```

No flake, but a lint error!

bobcatfish:
image

:disappointed_relieved:

bobcatfish:
7th time's a charm!

PR #257, opened 2018-07-26
Fully qualify images in dockerfile tests

priyawadhwa:
@bobcatfish pointed out when running integration tests locally, if the local image isn't the actual latest version then container-diff will fail since kaniko and docker are building off different images.

Fully qualifying the base images in the test dockerfiles should fix this problem.


bobcatfish:
> @bobcatfish pointed out when running integration tests locally, if the local image isn't the actual latest version then container-diff will fail since kaniko and docker are building off different images.

Well you pointed it out to me first! ;)

PR #262, opened 2018-07-30
Set a kaniko user agent.

dlorenc:


pkg/executor/executor.go
```diff
@@ -181,7 +181,17 @@ func DoBuild(k KanikoBuildArgs) (v1.Image, error) {
 	return nil, err
 }
 
+type withUserAgent struct {
+	t http.RoundTripper
+}
+
+func (w *withUserAgent) RoundTrip(r *http.Request) (*http.Response, error) {
+	r.Header.Set("User-Agent", fmt.Sprintf("kaniko/version.Version()"))
```
priyawadhwa: I think this should be `fmt.Sprintf("kaniko/%s", version.Version())`

dlorenc: Yes, yes it should :)

PR #264, opened 2018-07-31
Add a bit more context to layer offset failures

bobcatfish:
In #251 we are investigating test flakes due to layer offsets not
matching, this change will give us a bit more context so we can be sure
which image has which number of layers.

Also updated reproducible Dockerfile to be built with reproducible flag,
which I think was the original intent (without this change, there is no
difference between how `kaniko-dockerfile_test_copy_reproducible` and
`kaniko-dockerfile_test_copy` are built.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -42,6 +42,16 @@ type gcpConfig struct {
 	onbuildBaseImage string
 }
 
+type imageDetails struct {
+	name      string
+	numLayers int
+	digest    string
+}
+
+func (i imageDetails) String() string {
+	return fmt.Sprintf("Image: [%s] Digest: [%s] Number of Layers: [%d]", i.name, i.digest, i.numLayers)
+}
+
 func initGCPConfig() *gcpConfig {
 	var c gcpConfig
 	flag.StringVar(&c.gcsBucket, "bucket", "", "The gcs bucket argument to uploaded the tar-ed contents of the `integration` dir to.")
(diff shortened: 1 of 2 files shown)


integration/integration_test.go
```diff
@@ -225,7 +225,7 @@ func checkLayers(image1, image2 string, offset int) error {
 	}
 	actualOffset := int(math.Abs(float64(lenImage1 - lenImage2)))
 	if actualOffset != offset {
-		return fmt.Errorf("incorrect offset between layers of %s and %s: expected %d but got %d", image1, image2, offset, actualOffset)
+		return fmt.Errorf("Difference in number of layers in each image is %d but should be %d. %s has %d layers and %s has %d layers", actualOffset, offset, image1, lenImage1, image2, lenImage2)
```
dlorenc: How tough would it be to dump the full image digest as part of this? That way we could debug later.

bobcatfish: IT IS DONE

integration/integration_test.go
```diff
@@ -206,42 +216,52 @@ func TestLayers(t *testing.T) {
...
 }
 
-func numLayers(image string) (int, error) {
+func getImageDetails(image string) (*imageDetails, error) {
 	ref, err := name.ParseReference(image, name.WeakValidation)
 	if err != nil {
-		return 0, fmt.Errorf("Couldn't parse referance to image %s: %s", image, err)
+		return nil, fmt.Errorf("Couldn't parse referance to image %s: %s", image, err)
+	}
+	imgRef, err := daemon.Image(ref)
```
dlorenc: Only one of these should be a daemon.Image, right? The other is remote?

bobcatfish: @dlorenc [the kaniko image is pulled](https://github.com/bobcatfish/kaniko/blob/f7b659d701b4788c8cf91d4a6e4221039857ff51/integration/integration_test.go#L217) right before this executes (maybe that should be made more clear somehow?)

bobcatfish:
Latest kokoro failure is a test timeout, very odd :thinking:

Issue #268, opened 2018-08-01
Running kaniko from inside a container doesn't seem to work


xinau:
HI.
I have the following problem:
I'm unable to build my Dockerfile with kaniko from inside the kaniko container.
```console
$ tree
.
├── Dockerfile
└── main.go
$ docker run --rm -it --entrypoint="/busybox/sh" -v $PWD:/workspace gcr.io/kaniko-project/executor:debug
/ # /kaniko/executor -d test.hub
```
Stops at with following line and raising 0 errors
```
...
INFO[0027] Taking snapshot of full filesystem...        
INFO[0029] Executing 0 build triggers                   
INFO[0029] cmd: Add [.]                                 
INFO[0029] dest: /src                                   
INFO[0029] cmd: copy [.]                                
INFO[0029] dest: /src                                   
INFO[0029] Creating directory /src                      
INFO[0029] Copying file /workspace/Dockerfile to /src/Dockerfile 
INFO[0029] Copying file /workspace/main.go to /src/main.go 
INFO[0029] cmd: /bin/sh                                 
INFO[0029] args: [-c cd /src && go build -o app]        
Killed
```
Doing the same without using `/busybox/sh` does work
```console
$ docker run --rm -it --entrypoint="/kaniko/executor" -v $PWD:/workspace gcr.io/kaniko-project/executor:debug -d test.hub
```

The contents of the `Dockerfile` and `app.go`.
```docker
FROM golang:1.10.3-alpine AS build
ADD . /src
RUN cd /src && go build -o app

FROM alpine
WORKDIR /app
COPY --from=build /src/app /app/
CMD [ "./app" ]
```
```
package main

import "fmt"

func main() {
    fmt.Println("Hello, World!")
}
```

Delphinator:
+1 on this. Everything after 6fe9ea42484eccbb2fc86e5ae74f87a18477df19 is completely broken. The only difference from the failing log below and the successful one is the commit kaniko was built from.

# 6fe9ea42484eccbb2fc86e5ae74f87a18477df19

```
$ kaniko-executor --dockerfile Dockerfile --context `pwd`/images/$CI_JOB_NAME --verbosity warn --single-snapshot --destination "[REDACTED]"
2018/08/01 12:40:59 No matching credentials were found, falling back on anonymous
+ apt-get update -qq
+ apt-get upgrade -yqq
+ apt-get install -y -qq --no-install-recommends rsync
[...]

PR #270, opened 2018-08-01
Only add whiteout files once

priyawadhwa:
I found a bug when building this [Dockerfile](https://github.com/GoogleContainerTools/kaniko/issues/268#issuecomment-409631156) -- whiteouts were being added in every layer after the command which removed the files. This was causing extra layers to be built.

I added functions to `LayeredMap` similar to the ones already there so that a whiteout file is only added once.


pkg/snapshot/snapshot.go
```diff
@@ -141,10 +142,16 @@ func (s *Snapshotter) snapShotFS(f io.Writer) (bool, error) {
 		// Only add the whiteout if the directory for the file still exists.
 		dir := filepath.Dir(path)
 		if _, ok := memFs[dir]; ok {
-			logrus.Infof("Adding whiteout for %s", path)
-			filesAdded = true
-			if err := util.Whiteout(path, w); err != nil {
-				return false, err
+			maybeAdd, err := s.l.MaybeAddWhiteout(path)
```
sharifelgamal: nit: this boolean should probably be called something like "addWhiteout"

priyawadhwa: done

PR #271, opened 2018-08-01
Fix process group handling.

### LINKED ISSUE #268: Running kaniko from inside a container doesn't seem to work

dlorenc:
Also add a makefile target to build the debug image.

I think this fixes #268, but I'm still not sure of the reasoning. I don't quite understand why this worked at all inside a container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,3 +51,4 @@ integration-test:
 .PHONY: images
 images:
 	docker build -t $(REGISTRY)/executor:latest -f deploy/Dockerfile .
+	docker build -t $(REGISTRY)/executor:debug -f deploy/Dockerfile_debug .
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -35,4 +35,4 @@ fi
 echo "Running integration tests..."
 make out/executor
 pushd integration
-go test -v --bucket "${GCS_BUCKET}" --repo "${IMAGE_REPO}"
+go test -v --bucket "${GCS_BUCKET}" --repo "${IMAGE_REPO}" --timeout 30m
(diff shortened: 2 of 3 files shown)


dlorenc:
I'll merge this for now, but continue to think on why this worked at all so we can try to come up with a better test case.

xinau:
@dlorenc Thanks. That was quick :smile_cat:

PR #273, opened 2018-08-02
Don't copy same files twice in copy integration tests

### LINKED ISSUE #251: Number of layers inconsistent with kaniko builds that copy/copy the same files multiple times

bobcatfish:
The default hashing algorithm used by kaniko to determine if two files
are the same uses the files' mtime (the inode's modification time). It
turns out that this time is not always up to date, meaning that a file
could be modified but when you stat the file, the modification time may
not yet have been updated.

The copy integration tests were adding the same directory twice, the
second instance being to test copying a directory with a wilcard '*'.
Since the mtime is sometimes not updated, this caused kaniko to
sometimes think the files were the same, and sometimes think they were
different, varying the number of layers it created.

Now we will update those tests to use a completely different set of
files instead of copying the same files again.

In a later commit (which will hopefully fix #251) we will add a fix for
this and a new test case that will intentionally exercise this
functionality. In the meantime we'll prevent noisy test failures for
submitters.

Also fixed a bug I introduced into cmd.go :sweat_smile:


bobcatfish:
```
    --- FAIL: TestLayers/test_layer_Dockerfile_test_copy_same_file (4.96s)
    	integration_test.go:220: Difference in number of layers in each image is 2 but should be 0. Image 1: Image: [gcr.io/kaniko-test/docker-dockerfile_test_copy_same_file] Digest: [ceedae45f627715ed0f4364ccbd2d9bb2636a4a3f4f163a212abe385b3157088] Number of Layers: [43], Image 2: Image: [gcr.io/kaniko-test/kaniko-dockerfile_test_copy_same_file] Digest: [75124d7c203dc626435e8d7e4eb949e8d5f93e857560040dcb62e4cd54ef5459] Number of Layers: [41]
```
**howling** NOOOOOO

pkg/util/util.go
```diff
@@ -45,6 +46,7 @@ func Hasher() func(string) (string, error) {
 		if err != nil {
 			return "", err
 		}
+		logrus.Infof("File %s Mode %s ModTime %s IsRegular %s\n", p, fi.Mode().String(), fi.ModTime().String(), fi.Mode().IsRegular())
```
priyawadhwa: we might want to remove this or make it logrus.Debugf since info on every file every time we snapshot could make it difficult to parse through logs

bobcatfish: Oh yeah, I don't want to commit this, I just added this so I could see what was happening with these latest failures :innocent:

PR #278, opened 2018-08-07
Provide failing test for an env variable with multiple '='

AndreaGiardini:
This pull request provides a failing integration test.

Whenever a Dockerfile contains an environment variable (`ENV`) with multiple `=` the final image contains an env variable with a stripped-down version of the content.

It would be nice if somebody could point me in the right direction to fix this problem :)

Thanks


AndreaGiardini:
I signed it!

AndreaGiardini:
A failing test case and the fix have been provided :)

PR #281, opened 2018-08-08
Set default home value

### LINKED ISSUE #280: Set default value of HOME to /root
Docker will set $HOME to `/root` by default -- this created a bug when trying to reproduce #272 with [this](https://github.com/GoogleContainerTools/kaniko/issues/272#issuecomment-410940532) Dockerfile

priyawadhwa:
Based on this [documentation](https://docs.docker.com/engine/reference/run/#env-environment-variables) it looks like the default value of home is set based on the user.

From some experimentation it looks like if the USER is root we set the default value to /root, and if USER is set then the default value of HOME is /home/${user}

Should fix #280

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -61,7 +61,7 @@ func (r *RunCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	replacementEnvs := buildArgs.ReplacementEnvs(config.Env)
-	cmd.Env = addDefaultEnvs(config.User, replacementEnvs)
+	cmd.Env = addDefaultHOME(config.User, replacementEnvs)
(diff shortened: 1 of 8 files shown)


pkg/commands/run.go
```diff
@@ -113,6 +115,22 @@ func (r *RunCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 	return nil
 }
 
+// addDefaultEnvs adds default values of variables (like $HOME) if they aren't already set
+func addDefaultEnvs(user string, envs []string) []string {
+	for _, env := range envs {
+		split := strings.SplitN(env, "=", 2)
+		if split[0] == constants.HOME {
+			return envs
```
dlorenc: This seems like it can only set the default HOME, as designed. Maybe rename the function, or refactor it to support more default values in the future?

dlorenc:
missed a spot:

```
# github.com/GoogleContainerTools/kaniko/pkg/commands
pkg/commands/run_test.go:68:14: undefined: addDefaultEnvs
```

Issue #282, opened 2018-08-09
Hardlinks are not (well) supported by image unpacker


vito:
Hi there! Very excited for Kaniko - I'm modeling Concourse's new [Registry Image resource](https://github.com/concourse/registry-image-resource) on Kaniko's image extraction process, and noticed one gotcha that affects Kaniko too.

The problem is rooted in the fact that Kaniko extracts image layers in reverse order.

Try having Kaniko fetch `vito/broken-base` and run `git` from the resulting rootfs.

Here's a `Dockerfile` for that:

```Dockerfile
FROM vito/broken-base
RUN ls -al /usr/libexec/git-core/git /usr/bin/git
RUN git --version
```

And here's the `Dockerfile` for `vito/broken-base`:

```Dockerfile
FROM alpine
RUN apk --no-cache add git
RUN rm /usr/bin/git && ln -s /usr/libexec/git-core/git /usr/bin/git
```

I see this output:

```
INFO[0000] Unpacking filesystem of vito/broken-base...  
2018/08/09 01:18:50 No matching credentials found for index.docker.io, falling back on anonymous
INFO[0000] Mounted directories: [/kaniko /var/run /sys /dev/shm /dev/pts /tmp/garden-init /proc /scratch /tmp/build/e55deab7 /tmp/build/e55deab7/source /tmp/build/e55deab7/image /etc/hosts /etc/resolv.conf] 
INFO[0000] Unpacking layer: 2                           
INFO[0000] Unpacking layer: 1                           
INFO[0000] Not adding /usr because it was added by a prior layer 
INFO[0000] Not adding /usr/bin because it was added by a prior layer 
INFO[0000] Not adding /usr/bin/git because it was added by a prior layer 
INFO[0001] Unpacking layer: 0                           
INFO[0001] Not adding /etc because it was added by a prior layer 
INFO[0001] Not adding /etc/apk because it was added by a prior layer 
INFO[0001] Not adding /etc/apk/protected_paths.d because it was added by a prior layer 
INFO[0001] Not adding /etc/apk/world because it was added by a prior layer 
INFO[0001] Not adding /etc/hosts because it is whitelisted 
INFO[0001] Not adding /etc/ssl because it was added by a prior layer 
INFO[0001] Not adding /etc/ssl/certs because it was added by a prior layer 
INFO[0001] Not adding /lib because it was added by a prior layer 
INFO[0001] Not adding /lib/apk because it was added by a prior layer 
[...]

PR #284, opened 2018-08-10
Fix bug in extracting hardlinks

priyawadhwa:
Should fix the issues brought up in #282


pkg/util/fs_util.go
```diff
@@ -247,6 +287,15 @@ func checkWhiteouts(path string, whiteouts map[string]struct{}) bool {
 	return false
 }
 
+func checkSymlinks(path string, symlinks map[string]struct{}) bool {
```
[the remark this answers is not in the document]
priyawadhwa: You're right, thanks for pointing that out! I'll remove it.

pkg/util/fs_util.go
```diff
@@ -101,13 +106,47 @@ func GetFSFromImage(root string, img v1.Image) error {
...
+				}
+				_, previouslyAdded := fs[linkname]
+				if previouslyAdded || checkWhiteouts(linkname, whiteouts) {
+					if h, ok := hardlinks[linkname]; ok {
+						h.links = append(h.links, hdr)
+						continue
+					}
+					hardlinks[linkname] = &hardlink{
+						links:  []*tar.Header{hdr},
+						reader: tr,
```
vito: Hmm, I'd be kinda surprised if saving off the tar reader here works later. Per the docs for `tr.Next`:

> Next advances to the next entry in the tar archive. The Header.Size determines how many bytes can be read for the next file. *Any remaining data in the current file is automatically discarded.*

(emphasis mine)

One problem is that it's basically streaming through the archvie sequentially as it's downloading. It never buffers to disk or RAM, so I don't think you can go "back" to earlier entries in the archive. Tricky.

pkg/util/fs_util.go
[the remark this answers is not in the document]
priyawadhwa: I think we might actually need the symlink check, since we only want to specially hardlink at the end of extraction if the linked file exists but can't be hardlinked (which I think only happens if it's a symlink)

If a file was normally extracted in a previous layer than we can extract the file and create the hardlink normally

pkg/util/fs_util.go
priyawadhwa: Thanks for pointing this out, fixed.

priyawadhwa:
Woot tests are finally passing, PTAL!

pkg/util/fs_util.go
```diff
@@ -55,6 +62,10 @@ func GetFSFromImage(root string, img v1.Image) error {
 
 	fs := map[string]struct{}{}
 	whiteouts := map[string]struct{}{}
+	hardlinks, err := retrieveHardlinks(layers)
```
dlorenc: Do you think it would make sense to process each layer separately instead of doing them all at once? I think we already disallow hardlinks to cross layer boundaries.

priyawadhwa: Yah I think that would be better, I changed it to retrieve hardlinks as we process layers

dlorenc:
I think this needs a rebase now.

priyawadhwa:
Done! :)

priyawadhwa:
Closing, fixed by #326

PR #289, opened 2018-08-14
Always snapshot files in COPY and RUN commands

### LINKED ISSUE #251: Number of layers inconsistent with kaniko builds that copy/copy the same files multiple times

bobcatfish:
Kaniko uses mtime (as well as file contents and other attributes) to
determine if files have changed. COPY and ADD commands should _always_
update the mtime, because they actually overwrite the files. However it
turns out that the mtime can lag, so kaniko would sometimes add a new
layer when using COPY or ADD on a file, and sometimes would not. This
leads to a non-deterministic number of layers.

To fix this, we have updated the kaniko commands to be more
authoritative in declaring when they have changed a file (e.g. WORKDIR
will now only create the directory when it doesn't exist) and we will
trust those files and _always_ add them, instead of only adding them if
they haven't changed.

It is possible for RUN commands to also change the filesystem, in which
case kaniko has no choice but to look at the filesystem to determine
what has changed. For this case we have added a call to `sync` however
we still cannot guarantee that sometimes the mtime will not lag, causing the
number of layers to be non-deterministic. However when I tried to cause
this behaviour with the RUN command, I couldn't.

This changes the snapshotting logic a bit; before this change, the last
command of the last stage in a Dockerfile would always scan the whole
file system and ignore the files returned by the kaniko command. Instead
we will now trust those files and assume that the snapshotting
performed by previous commands will be adequate.

Docker itself seems to rely on the storage driver to determine when
files have changed and so doesn't have to deal with these problems
directly.

An alternative implementation would use `inotify` to track which files
have changed. However that would mean watching every file in the
filesystem, and adding new watches as files are added. Not only is there
a limit on the number of files that can be watched, but according to the
man pages a) this can take a significant amount of time b) there is
complication around when events arrive (e.g. by the time they arrive,
the files may have changed) and lastly c) events can be lost, which
would mean we'd run into this non-deterministic behaviour again anyway.

Fixes #251


README.md
```diff
@@ -352,3 +354,19 @@ provides.
...
+[`--snapshotMode=time`](#--snapshotmode), only use) a file's
+[`mtime`](https://en.wikipedia.org/wiki/Inode#POSIX_inode_description) to determine
+if the file has changed. Unfortunately there is a delay between when changes to a
+file are made and when the `mtime` is updated. This means:
+
+* With the default snapshot mode (`--snapshotMode=full`), whether or not kaniko will
+  add a layer in the case where a `RUN` command modifies a file but the contents do
+  not change is non-deterministic.
+* With the time-only snapshot mode (`--snapshotMode=time`), kaniko may miss changes
+  introduced by `RUN` commands entirely.
```
priyawadhwa: Just a few thoughts:

1. Could we put "but the contents do not change" in bold? Just in case people only skim this section and think RUN is broken, when it only happens in the kinda strange case where nothing actually changes
2. We could mention that since this only happens when files don't change the final filesystem will still be correct and the final image should still function correctly, it's just the composition of the layers that will differ from "docker build"
3. After "kaniko may miss changes introduced by RUN completely" we should mention that this limitation is theoretical since it hasn't been reproduced (yet!)

integration/dockerfiles/Dockerfile_test_copy_same_file_many_times
```diff
@@ -0,0 +1,49 @@
...
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
+COPY context/foo /foo
```
priyawadhwa: SOLID

bobcatfish: X'D

README.md
bobcatfish: Okay I gave it a shot @priyawadhwa ! PTAL and let me know if we should make any more tweaks. Also another option is we could make another doc and link to it somewhere that's not the main README, esp. since as you pointed out we don't know if this will ever actually happen.

bobcatfish:
image

:thinking:

README.md
priyawadhwa: Awesome, I think this makes a lot of sense! I think we could:

1.  put the note "Note that these issues are currently theoretical only. If you see this issue occur, please open an issue." in the first bullet point since I think the second bullet point isn't theoretical. 

2. remove the word "theoretically" from the second bullet point (since I think this was the bug you found? please ignore this if I'm wrong!)

dlorenc:
I'm trying to setup prow on this repo, sorry for using your PR as a guinea pig :)

bobcatfish:
> I'm trying to setup prow on this repo, sorry for using your PR as a guinea pig :)

lol I totally saw this go by in my email and did not realize it was _I_ who was the guinea pig

pkg/snapshot/snapshot.go
```diff
@@ -81,16 +94,19 @@ func (s *Snapshotter) snapshotFiles(f io.Writer, files []string) (bool, error) {
 	}
 	logrus.Infof("Taking snapshot of files %v...", files)
 	snapshottedFiles := make(map[string]bool)
+	n := len(files)
 	for _, file := range files {
 		parentDirs := util.ParentDirectories(file)
+		// If we do end up snapshotting the dirs, we need to snapshot them before
+		// we snapshot their contents
 		files = append(parentDirs, files...)
```
bobcatfish: @priyawadhwa can you confirm if the comment i added is correct? I was trying to understand why we were actually appending to `parentDirs` instead of to `files`

pkg/snapshot/snapshot.go
```diff
@@ -108,12 +124,24 @@ func (s *Snapshotter) snapshotFiles(f io.Writer, files []string) (bool, error) {
 		if err != nil {
 			return false, err
 		}
-		// Only add to the tar if we add it to the layeredmap.
-		addFile, err := s.l.MaybeAdd(file)
+
+		var fileAdded bool
+		lastParentFileIndex := len(files) - n
+		isParentDir := i < lastParentFileIndex
```
bobcatfish: this is one of the grossest things ive written in a long time (*knock on wood*) @priyawadhwa so let me know if it's not clear and/or you'd just like it to be nicer

i think the next step to making this better would be to break some of this logic out into a separate function or two

bobcatfish:
Okay this is ready for another look @priyawadhwa ! Re. the failing test I had to:

1. Deal with that bit of logic trying to make sure the kaniko dir was in the whitelist (#307)
2. Add some logic to continue to use `MaybeAdd` when looking at a files parent dirs

bobcatfish:
what! we've got bark but no meow! what kind of world is this

dlorenc:
> what! we've got bark but no meow! what kind of world is this

muahhhaaaha

spiffxp:
P0 https://twitter.com/adenforshaw/status/1031485169448808448

bobcatfish:
@spiffxp I am relieved the situation is being taken seriously!! 😤

kimsterv:
🚀

pkg/snapshot/snapshot.go
bobcatfish: kk let's do it!

pkg/snapshot/snapshot.go
bobcatfish: ah kk! if we're going to have 2 separate lists it wont matter too much anyway

bobcatfish:
@priyawadhwa I added https://github.com/GoogleContainerTools/kaniko/pull/289/commits/530c0081410a4a79b4f273364afaf514d5689a55 to separate handling of parent dirs from the files we're adding - but I started seeing some (TOTALLY UNRELATED) tests failing locally so I'm waiting to see if that turns up in CI or not before investigating further 🤔

integration/integration_test.go
```diff
@@ -135,11 +135,10 @@ func TestMain(m *testing.M) {
 	defer DeleteFromBucket(fileInBucket)
 
 	fmt.Println("Building kaniko image")
-	buildKaniko := exec.Command("docker", "build", "-t", ExecutorImage, "-f", "../deploy/Dockerfile", "..")
-	err = buildKaniko.Run()
+	cmd := exec.Command("docker", "build", "-t", ExecutorImage, "-f", "../deploy/Dockerfile", "..")
+	_, err = RunCommandWithoutTest(cmd)
 	if err != nil {
```
priyawadhwa: nit: I think go convention is to write error messages like this as 

```golang
if _, err := RunCommandWithoutTest(cmd); err != nil {

}
```

pkg/snapshot/snapshot.go
```diff
@@ -49,17 +49,28 @@ func (s *Snapshotter) Init() error {
...
 // a tarball of the changed files. Return contents of the tarball, and whether or not any files were changed
 func (s *Snapshotter) TakeSnapshot(files []string) ([]byte, error) {
 	buf := bytes.NewBuffer([]byte{})
 	var filesAdded bool
-	var err error
-	if files == nil {
-		filesAdded, err = s.snapShotFS(buf)
-	} else {
-		filesAdded, err = s.snapshotFiles(buf, files)
+	filesAdded, err := s.snapshotFiles(buf, files)
```
priyawadhwa: nit: I don't think you need to declare `var filesAdded bool` on line 56 anymore

pkg/snapshot/snapshot.go
```diff
@@ -70,55 +81,87 @@ func (s *Snapshotter) TakeSnapshot(files []string) ([]byte, error) {
...
+
+		err = s.l.Add(file)
 		if err != nil {
-			return false, err
+			return false, fmt.Errorf("Unable to add file %s to layered map: %s", file, err)
 		}
-		// Only add to the tar if we add it to the layeredmap.
-		addFile, err := s.l.MaybeAdd(file)
+		err = t.AddFileToTar(file)
 		if err != nil {
```
priyawadhwa: nit: same error formatting as I commented above

```golang
if err := s.l.Add(file); err != nil {
}
```

pkg/snapshot/snapshot.go
```diff
@@ -70,55 +81,87 @@ func (s *Snapshotter) TakeSnapshot(files []string) ([]byte, error) {
...
+			return false, fmt.Errorf("Error checking if file %s can be snapshotted: %s", file, err)
+		}
+		if !shouldSnapshot {
 			continue
 		}
 		snapshottedFiles[file] = true
-		info, err := os.Lstat(file)
+
+		err = s.l.Add(file)
 		if err != nil {
```
priyawadhwa: nit: same error formatting as I commented above

pkg/util/tar_util.go
```diff
@@ -31,8 +32,32 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// AddToTar adds the file i to tar w at path p
-func AddToTar(p string, i os.FileInfo, hardlinks map[uint64]string, w *tar.Writer) error {
+// Tar knows how to write files to a tar file.
+type Tar struct {
```
priyawadhwa: Nice!!

pkg/snapshot/snapshot.go
bobcatfish: ah good call

pkg/snapshot/snapshot.go
bobcatfish: whops thanks!

PR #292, opened 2018-08-17
Add Flag to Disable Push to Container Registry

rawkode:
The flag, `--no-push`, is added to allow building a container image
without pushing to a container registry. It can be common, especially
with multi-stage builds and `--target`, to build enough to run the tests,
and then perform a push in a separate CI step. This will facilitate these
workflows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,10 @@ Set this flag as `--tarPath=<path>` to save the image as a tarball at path inste
 
 Set this flag to indicate which build stage is the target build stage.
 
+#### --no-push
+
+Set this flag if you only want to build the image, without pushing to a registry.
+
 ### Debug Image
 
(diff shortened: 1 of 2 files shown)


priyawadhwa:
I'm also curious how you plan to use this flag in CI? If the image isn't saved anywhere, it can't be tested or used in later steps.

Are you just trying to verify the build works?

rawkode:
Hi @priyawadhwa, yeah; I'll update the README.

As for your question - yeah. We want to `--target tests` and ensure the code passes tests. We also use `--target lint` for linting checks, so we don't want to push these images anywhere.

priyawadhwa:
Cool, thanks for clarifying!

rawkode:
@priyawadhwa `README.md` updated :+1:

Issue #290, opened 2018-08-16
401 when pushing image to ECR - docker config location


brendanhatton:
When I follow the instructions on the Kaniko page Pushing to Amazon ECR (using the gcr.io/kaniko-project/executor:debug image), I get an error

```
ERRO[0022] Failed to push to destination ....amazonaws.com/... 
ERRO[0022] unsupported status code 401; body: Not Authorized
```

The instructions currently say to mount the docker configmap to `/root/.docker`. I noticed that one of the environment variables in the image is `DOCKER_CONFIG=/kaniko/.docker/` so when I mount the configmap to `/kaniko/.docker`, the credentials are detected and the push succeeds.

Is this just a documentation update?

priyawadhwa:
Hey @brendanhatton , thanks for pointing this out! I think you're right and this is a documentation issue.

priyawadhwa:
Made the documentation change in #306 , so I'll go ahead and close this issue. @brendanhatton please comment if you see this error again!

PR #306, opened 2018-08-23
Refactor command line arguments and the executor

priyawadhwa:
In this refactor I:

1. Created KanikoOptions to make it easier to pass around arguments
passed in through the command line
2. Reorganized executor.go by putting the logic for pushing the image in
a new file push.go
3. Made some error messages clearer
4. Fixed a mistake in the README for pushing to AWS
5. Marked the --bucket flag as hidden since we want people to use
--context instead, and marked an aws flag as hidden which is set in a
vendored directory (not totally sure why but it shows up in the help text for the command)

Should fix documentation issue in #290

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ To configure credentials, you will need to do the following:
         - name: aws-secret
           mountPath: /root/.aws/
         - name: docker-config
-          mountPath: /root/.docker/
+          mountPath: /kaniko/.docker/
     restartPolicy: Never
     volumes:
       - name: aws-secret
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -39,7 +39,8 @@ var (
 )
 
 func init() {
-	addSetupFlags(RootCmd)
+	RootCmd.PersistentFlags().StringVarP(&logLevel, "verbosity", "v", constants.DefaultLogLevel, "Log level (debug, info, warn, error, fatal, panic")
+	RootCmd.PersistentFlags().BoolVarP(&force, "force", "", false, "Force building outside of a container")
 	addKanikoOptionsFlags(RootCmd)
 	addHiddenFlags(RootCmd)
 }
@@ -74,11 +75,6 @@ var RootCmd = &cobra.Command{
 		}
 		return executor.DoPush(image, opts)
 	},
-}
-
-func addSetupFlags(cmd *cobra.Command) {
-	RootCmd.PersistentFlags().StringVarP(&logLevel, "verbosity", "v", constants.DefaultLogLevel, "Log level (debug, info, warn, error, fatal, panic")
-	RootCmd.PersistentFlags().BoolVarP(&force, "force", "", false, "Force building outside of a container")
 }
 
 // addKanikoOptionsFlags configures opts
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -240,7 +240,7 @@ To configure credentials, you will need to do the following:
         - name: aws-secret
           mountPath: /root/.aws/
         - name: docker-config
-          mountPath: /root/.docker/
+          mountPath: /kaniko/.docker/
```
tstromberg: Is this intentional? It's not mentioned in the PR description.

cmd/executor/cmd/root.go
```diff
@@ -17,52 +17,31 @@ limitations under the License.
...
-	RootCmd.PersistentFlags().BoolVarP(&dockerInsecureSkipTLSVerify, "insecure-skip-tls-verify", "", false, "Push to insecure registry ignoring TLS verify")
-	RootCmd.PersistentFlags().StringVarP(&logLevel, "verbosity", "v", constants.DefaultLogLevel, "Log level (debug, info, warn, error, fatal, panic")
-	RootCmd.PersistentFlags().BoolVarP(&force, "force", "", false, "Force building outside of a container")
-	RootCmd.PersistentFlags().StringVarP(&tarPath, "tarPath", "", "", "Path to save the image in as a tarball instead of pushing")
-	RootCmd.PersistentFlags().BoolVarP(&singleSnapshot, "single-snapshot", "", false, "Set this flag to take a single snapshot at the end of the build.")
-	RootCmd.PersistentFlags().BoolVarP(&reproducible, "reproducible", "", false, "Strip timestamps out of the image to make it reproducible")
-	RootCmd.PersistentFlags().StringVarP(&target, "target", "", "", " Set the target build stage to build")
-	RootCmd.PersistentFlags().BoolVarP(&noPush, "no-push", "", false, "Do not push the image to the registry")
+	addSetupFlags(RootCmd)
```
tstromberg: Adding a level of indirection for flag setup is unusual, but clearly you have something in mind. Do you mind clarifying the intent?

pkg/executor/push.go
```diff
@@ -0,0 +1,95 @@
...
+		tr := http.DefaultTransport
+		if opts.DockerInsecureSkipTLSVerify {
+			tr.(*http.Transport).TLSClientConfig = &tls.Config{
+				InsecureSkipVerify: true,
+			}
+		}
+		rt := &withUserAgent{t: tr}
+
+		if err := remote.Write(destRef, image, pushAuth, rt, remote.WriteOptions{}); err != nil {
+			logrus.Error(fmt.Errorf("Failed to push to destination %s", destination))
```
tstromberg: Logging and error and propagating it upwards usually results in duplicate errors being logged. I generally suggest one or the other unless there is a good reason.

pkg/executor/push.go
```diff
@@ -0,0 +1,95 @@
...
+	if opts.NoPush {
+		logrus.Info("Skipping push to container registry due to --no-push flag")
+		return nil
+	}
+	// continue pushing unless an error occurs
+	for _, destination := range opts.Destinations {
+		// Push the image
+		destRef, err := name.NewTag(destination, name.WeakValidation)
+		if err != nil {
+			return err
```
tstromberg: DoPush() has many code paths that can return an error. Do you mind adding some context to the errors returned by it to assist future debuggers? For instance:

"nametag failed: %v", err

I noticed other parts of the kaniko codebase use errors.Wrap for this.

PR #308, opened 2018-08-24
support to pull base images  from insecure registries.

everpeace:
## Motivation
We operate an insecure docker registry and kubernetes cluster in on-premise environment.  kaniko can push container images to the insecure registry.  But, can't pull from the insecure registry.

## How
make `--insecure-skip-tls-verify` be able to affect when pulling images.

Because user should be able to use multi stage build, with this option, it tries to connect to a registry in secure way first.  Only when it failed, it will switch to connect insecure mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,10 @@ Set this flag to indicate which build stage is the target build stage.
 
 Set this flag if you only want to build the image, without pushing to a registry.
 
+### --insecure-skip-tls-verify
+
+Pull from or Push to insecure registry ignoring TLS verify
+
 ### Debug Image
 
 The kaniko executor image is based off of scratch and doesn't contain a shell.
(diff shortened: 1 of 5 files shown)


pkg/options/options.go
```diff
@@ -18,16 +18,17 @@ package options
...
-	Reproducible                bool
-	Target                      string
-	NoPush                      bool
+	DockerfilePath                    string
+	Destinations                      multiArg
+	SrcContext                        string
+	SnapshotMode                      string
+	Bucket                            string
+	DockerInsecureSkipTLSVerify       bool
+	DockerInsecureSkipTLSVerifyAtPull bool
```
dlorenc: I'd be fine with just using the same DockerInsecureSkipTLSVerify flag for both of these, WDYT?

everpeace: I'd be fine too.  I'll fix it 👍

everpeace:
@dlorenc I changed to re-use `--insecure-skip-tls-verify` for pulling from insecure registries.

and I squashed my commits.  Would you mind taking a look again?

dlorenc:
Looked like an IO timeout, I'll retry the test run.

priyawadhwa:
Hey @everpeace, with #409 kaniko supports pulling and pushing from insecure registries. I'll go ahead and close this PR.

everpeace:
Thnaks, actually dtaniwaki is my colleague :-)

PR #311, opened 2018-08-24
Separate --insecure-skip-tls-verify flag into two separate flags

### LINKED ISSUE #310: --insecure-skip-tls-verify forces plain HTTP making the skip-tls-verify part confusing
Hi everybody,

setting the `--insecure-skip-tls-verify` option seems to both disable TLS verification in the HTTP client doing the upload, but allso sets the Registry.insecure flag internally, which according to my understanding of its source code forces the scheme to plain http, rendering the former option superfluous.

Worse, I originally set the flag above when testing against an internal HTTPS-only registry with an internal CA, i.e. HTTPS is required, but the registry's certificate is unknown, but the flag resulted in the executor forcing plain HTTP.

I therefore find that the flag in its current form is confusingly named and eithe name or function should be altered.

One solution, which I've already implemented, would be splitting the flag into two distinct flags. One for allowing plain HTTP registries and one for skipping certificate verification. As this would change existing command line flags there is probably some additional pondering to do on this one.

Another way would be to try HTTPS with certificate verification disabled and falling back to HTTP if the connection attempt fails, but this seems more complicated and error-prone to me.

DerDackel:
Possibly fixes #310


pkg/options/options.go
```diff
@@ -18,16 +18,17 @@ package options
...
-	SingleSnapshot              bool
-	Reproducible                bool
-	Target                      string
-	NoPush                      bool
+	DockerfilePath string
+	Destinations   multiArg
+	SrcContext     string
+	SnapshotMode   string
+	Bucket         string
+	DockerInsecure bool
```
priyawadhwa: Could we rename `DockerInsecure` to something which doesn't include Docker in the name, maybe `InsecurePush`?

DerDackel:
Done and done. I took the liberty of adding a warning about their intended use to both flags.

priyawadhwa:
Thanks! I think you just have to rebase against master and this should be good.

Issue #317, opened 2018-08-27
Add integration tests to diff individual layers


priyawadhwa:
Add integration tests to diff individual layers of our test images via `container-diff diff image1 image2 --type file --type layer`

bobcatfish:
Would be pretty cool to check the layer contents! ❤️

PR #319, opened 2018-08-27
Snapshot only specific files for COPY

### LINKED ISSUE #314: Image sizes increased dramatically since merging #289
Since merging #289, Kaniko seems to create layers containing not only the files copied by `COPY` instruction, but also files from the previous layers.

priyawadhwa:
Before #289 was merged, when copying over directories for COPY kaniko
would get a list of all files at the destination specified and add them
to the list of files to be snapshotted. If the destination was root it
would add all files. This worked because the snapshotter made sure the
file had been changed before adding it to the layer.

After #289, we changed the logic to add all files snapshotted to a layer
without checking if the files had been changed. This created the bug in which kaniko
got all the files at root and added them to the layer without checking
if they had been changed.

This change should fix this bug. Now, the CopyDir function returns a
list of files it copied over and only those files are added to the list
of files to be snapshotted.

I also added this change to the `UnpackLocalTarArchive` function in ADD, so that we only add specific files extracted from local tar archives to the list of snapshotted files.

Should fix #314

Once #317 is implemented, this change will be tested by integration tests. Until then, I locally tested the Dockerfile in the bug and it seems to be working now.

### DIFF
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -459,16 +459,18 @@ func DownloadFileToDest(rawurl, dest string) error {
 }
 
 // CopyDir copies the file or directory at src to dest
-func CopyDir(src, dest string) error {
+// It returns a list of files it copied over
+func CopyDir(src, dest string) ([]string, error) {
 	files, err := RelativeFiles("", src)
 	if err != nil {
-		return err
+		return nil, err
 	}
+	var copiedFiles []string
 	for _, file := range files {
 		fullPath := filepath.Join(src, file)
 		fi, err := os.Lstat(fullPath)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		destPath := filepath.Join(dest, file)
 		if fi.IsDir() {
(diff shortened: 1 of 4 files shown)


pkg/util/fs_util.go
```diff
@@ -155,21 +155,24 @@ func ChildDirInWhitelist(path, directory string) bool {
 	return false
 }
 
-func unTar(r io.Reader, dest string) error {
+// unTar returns a list of files extracted from the tar archive
```
bobcatfish: would be good to including in this docstring what the params `r` and `dest` are for (https://blog.golang.org/godoc-documenting-go-code) e.g. `unTar returns a list of files that have been extracted from the tar archive at r to the path at dest`

bobcatfish:
I'm sorry that that dog looks so suspicious

pkg/util/fs_util.go
priyawadhwa: for sure, just added it!

PR #326, opened 2018-08-29
Extract filesystem in order rather than in reverse

priyawadhwa:
Extracting the layers of the filesystem in order will make it easier to
extract cached layers and deal with hardlinks.

This PR implements extracting in order and adds an integration tests to
make sure hardlinks are extracted properly.

It also fixes two bugs I found when extracting symlinks:

1. We'd get a "file exists" error when trying to symlink to an existing
file with a whiteout later in the layer tarball
2. We'd get a "file exists" error when trying to create a symlink from a
file that was created in a prior layer (perhaps as a regular file or as
a symlink pointing to someting else)

To fix both of these, we resolve all symlinks in a layer at the end. I
also added logic to delete any existing paths before creating the
symlink.

### DIFF
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -61,7 +61,6 @@ func GetFSFromImage(root string, img v1.Image) error {
 			return err
 		}
 		tr := tar.NewReader(r)
-		symlinks := []*tar.Header{}
 		for {
 			hdr, err := tr.Next()
 			if err == io.EOF {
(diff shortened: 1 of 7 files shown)


pkg/util/fs_util.go
```diff
@@ -107,13 +98,18 @@ func GetFSFromImage(root string, img v1.Image) error {
 					logrus.Debugf("skipping symlink from %s to %s because %s is whitelisted", hdr.Linkname, path, hdr.Linkname)
 					continue
 				}
+				symlinks = append(symlinks, hdr)
```
dlorenc: Why do we need to special case symlinks here?

dlorenc: It seems like we could just create them during the pass.

priyawadhwa: You're right! I misunderstood how a layer would be formed if a file was removed and then symlinked in the same layer -- fixed it now.

Issue #313, opened 2018-08-25
Kaniko creates more layers than Docker


lubo:
When using `VOLUME` Dockerfile instruction, Kaniko creates a new layer, which contains only the directories specified in `VOLUME` instruction. Docker does not create those directories in the image at all.

peter-evans:
This issue was interesting to me so I had a quick look at Kaniko's code and it looks like a snapshot of the file system is taken with each layer it builds. If the snapshot contains data it adds a layer. The `VOLUME` command is creating directories and so the layer has "content."

If Docker doesn't add a layer with the `VOLUME` command, where does that info get stored in the final container image? Presumably it's written the image somewhere and the directories are created when the container starts.

PR #334, opened 2018-09-04
Fix handling of the volume directive

peter-evans:
This is an attempt to fix https://github.com/GoogleContainerTools/kaniko/issues/313 by not adding the created directories to the `VOLUME` command snapshot. Adding to the snapshot creates a layer in the container that doesn't exist when building with Docker.

### DIFF
--- a/pkg/commands/volume.go
+++ b/pkg/commands/volume.go
@@ -48,7 +48,7 @@ func (v *VolumeCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.
 	for _, volume := range resolvedVolumes {
 		var x struct{}
 		existingVolumes[volume] = x
-		err := util.AddPathToVolumeWhitelist(volume)
+		err := util.AddVolumePathToWhitelist(volume)
 		if err != nil {
 			return err
 		}
@@ -56,7 +56,7 @@ func (v *VolumeCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.
 		// Only create and snapshot the dir if it didn't exist already
 		if _, err := os.Stat(volume); os.IsNotExist(err) {
 			logrus.Infof("Creating directory %s", volume)
-			v.snapshotFiles = []string{volume}
+			v.snapshotFiles = append(v.snapshotFiles, volume)
 			if err := os.MkdirAll(volume, 0755); err != nil {
 				return fmt.Errorf("Could not create directory for volume %s: %s", volume, err)
 			}
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -138,6 +138,7 @@ func (s *stageBuilder) build(opts *config.KanikoOptions) error {
 	if err := s.snapshotter.Init(); err != nil {
 		return err
 	}
+	var volumes []string
 	args := dockerfile.NewBuildArgs(opts.BuildArgs)
 	for index, cmd := range s.stage.Commands {
 		finalCmd := index == len(s.stage.Commands)-1
@@ -167,6 +168,10 @@ func (s *stageBuilder) build(opts *config.KanikoOptions) error {
 			return err
 		}
 		files := command.FilesToSnapshot()
+		if cmd.Name() == constants.VolumeCmdName {
+			volumes = append(volumes, files...)
+			continue
+		}
 		var contents []byte
 
 		// If this is an intermediate stage, we only snapshot for the last command and we
@@ -188,9 +193,14 @@ func (s *stageBuilder) build(opts *config.KanikoOptions) error {
 				// the files that were changed, we'll snapshot those explicitly, otherwise we'll
 				// check if anything in the filesystem changed.
 				if files != nil {
+					if len(files) > 0 {
+						files = append(files, volumes...)
+						volumes = []string{}
+					}
 					contents, err = s.snapshotter.TakeSnapshot(files)
 				} else {
 					contents, err = s.snapshotter.TakeSnapshotFS()
+					volumes = []string{}
 				}
 			}
 		}
@@ -198,7 +208,6 @@ func (s *stageBuilder) build(opts *config.KanikoOptions) error {
 			return fmt.Errorf("Error taking snapshot of files for command %s: %s", command, err)
 		}
 
-		util.MoveVolumeWhitelistToWhitelist()
 		if contents == nil {
 			logrus.Info("No files were changed, appending empty layer to config. No layer added to image.")
 			continue
--- a/pkg/snapshot/snapshot.go
+++ b/pkg/snapshot/snapshot.go
@@ -25,7 +25,6 @@ import (
 	"path/filepath"
 	"syscall"
 
-	"github.com/GoogleContainerTools/kaniko/pkg/constants"
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
 	"github.com/sirupsen/logrus"
 )
@@ -84,21 +83,6 @@ func (s *Snapshotter) TakeSnapshotFS() ([]byte, error) {
 	return contents, err
 }
 
-func shouldSnapshot(file string, snapshottedFiles map[string]bool) (bool, error) {
-	if val, ok := snapshottedFiles[file]; ok && val {
-		return false, nil
-	}
-	whitelisted, err := util.CheckWhitelist(file)
-	if err != nil {
-		return false, fmt.Errorf("Error checking for %s in whitelist: %s", file, err)
-	}
-	if whitelisted && !isBuildFile(file) {
-		logrus.Infof("Not adding %s to layer, as it's whitelisted", file)
-		return false, nil
-	}
-	return true, nil
-}
-
 // snapshotFiles creates a snapshot (tar) and adds the specified files.
 // It will not add files which are whitelisted.
 func (s *Snapshotter) snapshotFiles(f io.Writer, files []string) (bool, error) {
@@ -123,11 +107,7 @@ func (s *Snapshotter) snapshotFiles(f io.Writer, files []string) (bool, error) {
 	}
 	for _, file := range parentDirs {
 		file = filepath.Clean(file)
-		shouldSnapshot, err := shouldSnapshot(file, snapshottedFiles)
-		if err != nil {
-			return false, fmt.Errorf("Error checking if parent dir %s can be snapshotted: %s", file, err)
-		}
-		if !shouldSnapshot {
+		if val, ok := snapshottedFiles[file]; ok && val {
 			continue
 		}
 		snapshottedFiles[file] = true
@@ -148,33 +128,20 @@ func (s *Snapshotter) snapshotFiles(f io.Writer, files []string) (bool, error) {
 	// Next add the files themselves to the tar
 	for _, file := range files {
 		file = filepath.Clean(file)
-		shouldSnapshot, err := shouldSnapshot(file, snapshottedFiles)
-		if err != nil {
-			return false, fmt.Errorf("Error checking if file %s can be snapshotted: %s", file, err)
-		}
-		if !shouldSnapshot {
+		if val, ok := snapshottedFiles[file]; ok && val {
 			continue
 		}
 		snapshottedFiles[file] = true
 
-		if err = s.l.Add(file); err != nil {
+		if err := s.l.Add(file); err != nil {
 			return false, fmt.Errorf("Unable to add file %s to layered map: %s", file, err)
 		}
-		if err = t.AddFileToTar(file); err != nil {
+		if err := t.AddFileToTar(file); err != nil {
 			return false, fmt.Errorf("Error adding file %s to tar: %s", file, err)
 		}
 		filesAdded = true
 	}
 	return filesAdded, nil
-}
-
-func isBuildFile(file string) bool {
-	for _, buildFile := range constants.KanikoBuildFiles {
-		if file == buildFile {
-			return true
-		}
-	}
-	return false
 }
 
 // shapShotFS creates a snapshot (tar) of all files in the system which are not
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -34,17 +34,30 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var whitelist = []string{
-	"/kaniko",
-	// /var/run is a special case. It's common to mount in /var/run/docker.sock or something similar
-	// which leads to a special mount on the /var/run/docker.sock file itself, but the directory to exist
-	// in the image with no way to tell if it came from the base image or not.
-	"/var/run",
-	// similarly, we whitelist /etc/mtab, since there is no way to know if the file was mounted or came
-	// from the base image
-	"/etc/mtab",
+type WhitelistEntry struct {
+	Path            string
+	PrefixMatchOnly bool
 }
-var volumeWhitelist = []string{}
+
+var whitelist = []WhitelistEntry{
+	{
+		Path:            "/kaniko",
+		PrefixMatchOnly: false,
+	},
+	{
+		// /var/run is a special case. It's common to mount in /var/run/docker.sock or something similar
+		// which leads to a special mount on the /var/run/docker.sock file itself, but the directory to exist
+		// in the image with no way to tell if it came from the base image or not.
+		Path:            "/var/run",
+		PrefixMatchOnly: false,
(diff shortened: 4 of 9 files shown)


peter-evans:
The integration test for [Dockerfile_test_volume](https://github.com/GoogleContainerTools/kaniko/blob/master/integration/dockerfiles/Dockerfile_test_volume) might fail because of [this offset](https://github.com/GoogleContainerTools/kaniko/blob/master/integration/integration_test.go#L226-L228). I'll try and run the integration tests locally to fix if necessary.

priyawadhwa:
Yup, as expected the layers test failed:

```
FAIL: TestLayers/test_layer_Dockerfile_test_volume (13.38s)
    	integration_test.go:244: Difference in number of layers in each image is 0 but should be 1. Image 1: Image: [gcr.io/kaniko-test/docker-dockerfile_test_volume] Digest: [032330a4ebbb08b66fcb1bc8dd5670309fcf99a8138b0a201f4e4a3644e23abb] Number of Layers: [4], Image 2: Image: [gcr.io/kaniko-test/kaniko-dockerfile_test_volume] Digest: [983740719a6d9a9a7e1afa618f390e7db2170a1188f730067d0ed77f55f61055] Number of Layers: [4]
```

But the filesystem test also failed:

```
FAIL: TestRun/test_Dockerfile_test_volume (27.64s)
    	integration_test.go:196: diff = [
    		  {
    		    "Image1": "gcr.io/kaniko-test/docker-dockerfile_test_volume",
    		    "Image2": "gcr.io/kaniko-test/kaniko-dockerfile_test_volume",
    		    "DiffType": "File",
    		    "Diff": {
    		      "Adds": null,
    		      "Dels": [
    		        {
    		          "Name": "/baz/bat",
    		          "Size": 0
    		        },
    		        {
    		          "Name": "/foo/bar",
    		          "Size": 0
    		        }
    		      ],
    		      "Mods": null
    		    }
    		  },
    		  {
    		    "Image1": "gcr.io/kaniko-test/docker-dockerfile_test_volume",
    		    "Image2": "gcr.io/kaniko-test/kaniko-dockerfile_test_volume",
    		    "DiffType": "Metadata",
    		    "Diff": {
    		      "Adds": [],
    		      "Dels": []
    		    }
    		  }
    		]
```
Meaning that `/baz/bat` and `/foo/bar` were in the docker built version of the image but not the kaniko version (since we removed snapshotting).

I think this will require further exploration into how the `VOLUME` command works, because it seems like the directories are created and snapshotted. However, this doesn't happen in a unique layer, which would explain why kaniko has an extra layers. 

One theory could be that Docker snapshotting `VOLUME` happens in conjunction with another command. Would you be interested in exploring this more?

peter-evans:
Sure. I'll do some further investigation.

peter-evans:
I can't find any official documentation to back this up but this is my theory of how Docker treats the `VOLUME` directive based on some testing.

Dockerfile directives seem to be split into two categories:
1. Directives that trigger a layer to be written to the image __if the file-system has changes__.
   e.g. `FROM`, `COPY`, `ADD`, `RUN`, etc.
2. Directives that do not trigger a layer to be written to the image __regardless of whether the file-system has changes__.
   e.g. `ARG`, `ENV`, `ENTRYPOINT`, `EXPOSE`, etc.

The `VOLUME` directive is in the second category. It never triggers a layer to be written, even if it causes file-system changes. If directives in the second category make file-system changes since the last layer written to the image they are carried over until a directive in the first category is met. In actuality, `VOLUME` seems to be the only second category directive that _can_ make file-system changes. So in that regard it could be considered an exception. If the end of the Dockerfile is reached and there are still file-system changes carried over from `VOLUME` directives they are discarded—no final layer is written.

Below is an example to demonstrate. The the `VOLUME` directive changes are combined with file-system changes at the next `RUN` directive. Notice how that even though `RUN echo "foo2"` makes no file-system changes it still triggers a layer. This is because the file-system changes of the previous `VOLUME` directive have been carried over. If we removed `VOLUME /volume1` and rebuilt the image `RUN echo "foo2"` would not trigger a new layer to be written. Also notice that the final `VOLUME /volume5` file-system changes are discarded and no layer is added.

__Dockerfile__:
```dockerfile
VOLUME /volume1
RUN mkdir /foo1
VOLUME /volume2
VOLUME /volume3
RUN echo "foo2"
VOLUME /volume4
RUN mkdir /foo3
VOLUME /volume5
```
```
$ container-diff analyze daemon://kanikotest --type layer
-----FileLayer-----

Analysis for kanikotest Layer 0:
<Truncated for brevity>

Analysis for kanikotest Layer 1:
FILE            SIZE
/foo1           0
/volume1        0

Analysis for kanikotest Layer 2:
FILE            SIZE
/volume2        0
/volume3        0

Analysis for kanikotest Layer 3:
FILE            SIZE
/foo3           0
/volume4        0
```

priyawadhwa:
Hey @peter-evans, thanks for investigating this! Your findings match what I've found testing locally as well, so I think we can go ahead with these assumptions for now. 

Right now, kaniko snapshots in two ways:
1. The entire filesystem (after RUN commands)
2. Specific files (ADD/COPY/WORKDIR/VOLUME)

The code for this is [here](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/build.go#L111-L115)


In the case where VOLUME is followed by one of the commands in (2), you'd need some way of remembering which files were created by VOLUME and appending them to the list of files to be snapshotted [here](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/build.go#L112). 

Right now, we maintain a [whitelist](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/build.go#L122) of directories created by volumes, which are added [here](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/commands/volume.go#L51). We could repurpose this to store volumes that haven't been snapshotted, add these files to `snapshotFiles`, and then clear out the list after snapshotting.

This, along with a few more integration tests (you'd just have to add a couple test Dockerfiles in the [dockerfiles](https://github.com/GoogleContainerTools/kaniko/tree/master/integration/dockerfiles) directory, should be good for now. 

And again, thanks so much for looking into this!

priyawadhwa:
Another (probably better) option is to remove the volumeWhitelist from the `util` package entirely, and instead store dirs created by volume in `build.go`:

```golang
// This should live outside the command execution loop
var volumes []string
...
// This check should happen after a command has been executed
if dockerCommand.Name() == "volume" {
   volumes = append(volumes, snapshotFiles...)
   continue
}
...
if snapshotFiles != nil {
        if len(snapshotFiles) > 0  {
            snapshotFiles = append(snapshotFiles, volumes...)
            volumes = []string{}
       }
      contents, err = snapshotter.TakeSnapshot(snapshotFiles)
} else {
        contents, err = snapshotter.TakeSnapshotFS()
        volumes = []string{}
}
```
Then, before snapshotting specific files you can append the contents of `volumes`, and then clear out `volumes` after snapshotting has happened for a different command.

peter-evans:
@priyawadhwa Thanks for the pointers. I'll see what I can do.

peter-evans:
Missed some tests. Not ready for review yet.

pkg/executor/build.go
```diff
@@ -167,6 +168,10 @@ func (s *stageBuilder) build(opts *config.KanikoOptions) error {
 			return err
 		}
 		files := command.FilesToSnapshot()
+		if command.Name() == "volume" {
```
priyawadhwa: I think `cmd.Name()` should work here, that way you won't need to add the function to every command.

Let's also add `volume` to `constants.go`

pkg/util/fs_util.go
```diff
@@ -385,22 +401,15 @@ func CreateFile(path string, reader io.Reader, perm os.FileMode, uid uint32, gid
...
-// in this step to be whitelisted for all subsequent docker commands.
-func MoveVolumeWhitelistToWhitelist() error {
-	if len(volumeWhitelist) > 0 {
-		whitelist = append(whitelist, volumeWhitelist...)
-		volumeWhitelist = []string{}
-	}
+// AddVolumePathToWhitelist adds the given path to the whitelist with
+// PrefixMatchOnly set to true. Snapshotting will ignore paths prefixed
+// with the volume, but the volume itself will not be ignored.
+func AddVolumePathToWhitelist(path string) error {
```
priyawadhwa: In what case would we need to whitelist volumes? I'm not totally sure we need to whitelist them at all.

peter-evans: Yes, volumes do need to be whitelisted. This is because any file-system changes inside a volume should not be added to the snapshot. Consider this example:

```
VOLUME /foo
RUN echo "hello world" > /foo/hello
```
After the volume `/foo` is created any file-system changes within that volume should not be in the snapshot. The file `hello` will not be in the layer created at the `RUN` directive. However, the directory `foo` itself *should* be included in the snapshot. Due to this complication I felt that the best way to handle it was by being able to flag whitelist entries as `PrefixMatchOnly=true`. This only allows matches where the whitelisted directory is a prefix of the path being checked. If it's an exact match it will not be whitelisted. This allows the volume directories to be snapshot, but anything inside them to be whitelisted.

pkg/commands/volume.go
```diff
@@ -48,15 +48,15 @@ func (v *VolumeCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.
...
+		err := util.AddVolumePathToWhitelist(volume)
 		if err != nil {
 			return err
 		}
 
 		// Only create and snapshot the dir if it didn't exist already
 		if _, err := os.Stat(volume); os.IsNotExist(err) {
 			logrus.Infof("Creating directory %s", volume)
-			v.snapshotFiles = []string{volume}
+			v.snapshotFiles = append(v.snapshotFiles, volume)
```
peter-evans: Just a comment about this change. This is a bug I found where it is not respecting multiple volumes created by the same `VOLUME` directive.

The `Dockerfile_test_volume` test did not catch it because `tmp` is already a directory and so `v.snapshotFiles` did not get overridden. I added an extra volume for the directive to create to provide test coverage.
```
VOLUME /foo/bar /tmp /qux/quux
```

pkg/util/fs_util.go
peter-evans: After I wrote the above comment I went back to double check. Turns out it's more complicated still. See the following example built with Docker and the container-diff analysis.

```
VOLUME /foo
RUN echo “hello world” > /foo/hello
COPY run.sh /foo/run.sh
VOLUME /bar
COPY run.sh /bar/run.sh
VOLUME /baz
ADD run.sh /baz/run.sh
WORKDIR /baz/bat
RUN echo “hello world” > /bar/hello
RUN echo “hello world” > /baz/hello
RUN echo “hello world” > /tmp/hello
```

```
Analysis for kaniko-test Layer 1:
FILE        SIZE
/foo        0

Analysis for kaniko-test Layer 2:
FILE               SIZE
/foo               24B
/foo/run.sh        24B

Analysis for kaniko-test Layer 3:
FILE               SIZE
/bar               24B
/bar/run.sh        24B

Analysis for kaniko-test Layer 4:
FILE               SIZE
/baz               24B
/baz/run.sh        24B

Analysis for kaniko-test Layer 5:
FILE            SIZE
/baz            0
/baz/bat        0

Analysis for kaniko-test Layer 6:
FILE              SIZE
/tmp              12B
/tmp/hello        12B
```

File-system changes within volumes *are* whitelisted during system-wide snapshots. None of the `hello` files created within volumes appear in the layers. The only `hello` file that does appear is contained in `/tmp`, which isn't a volume. However, directives that specify files (ADD/COPY/WORKDIR) override the whitelist.

At the moment, `TakeSnapshot(files)` is checking the whitelist. I will try removing it and improve the volume integration test to cover these cases.

peter-evans: I removed checking the whitelist for `TakeSnapshot(files)` and it now works as expected. I added to `Dockerfile_test_volume_2` to provide coverage of these specific file directive cases.

peter-evans:
@priyawadhwa I can see that the kokoro run failed. Any way I can see the detail for that?
I've run all the integration tests on a clean VM and they all passed so I'm not sure what could be wrong.

priyawadhwa:
@peter-evans looks like the number of layers was incorrect:

```
    --- FAIL: TestLayers/test_layer_Dockerfile_test_volume (13.26s)
    	integration_test.go:233: Difference in number of layers in each image is 1 but should be 0. Image 1: Image: [gcr.io/kaniko-test/docker-dockerfile_test_volume] Digest: [e76e94ae181e524b80f45d8b43f6cf7eee04bb50b1bfd9bddd07c63449a73145] Number of Layers: [4], Image 2: Image: [gcr.io/kaniko-test/kaniko-dockerfile_test_volume] Digest: [0482d5821245a91e6cb5aca60a50bd5658e7f6644365583bd93031513138f3d9] Number of Layers: [5]
```

peter-evans:
@priyawadhwa I just spun up a new VM in GCP and ran the integration tests again and they all pass.
Please see the log here: [kaniko-integration-test.log](https://github.com/GoogleContainerTools/kaniko/files/2426381/kaniko-integration-test.log)

Ubuntu 16.04.5 LTS
Docker 18.06.1-ce, build e68fc7a
go version go1.11 linux/amd64

It would be great if you could let me know some detail about kokoro's build environment, such as distro, Docker version, etc.

peter-evans:
I also checked the layers in `kaniko-dockerfile_test_volume` and it correctly returns 4, not 5. Is kokoro definitely building the correct branch? I wonder if it got confused because I squashed the commits.

```
$ container-diff analyze asia.gcr.io/cloud-shell-215704/kaniko-dockerfile_test_volume --type=layer

-----FileLayer-----

Analysis for asia.gcr.io/cloud-shell-215704/kaniko-dockerfile_test_volume Layer 0:
<Truncated for brevity>

Analysis for asia.gcr.io/cloud-shell-215704/kaniko-dockerfile_test_volume Layer 1:
FILE        SIZE
/foo        0

Analysis for asia.gcr.io/cloud-shell-215704/kaniko-dockerfile_test_volume Layer 2:
FILE            SIZE
/foo            6B
/foo/hey        6B

Analysis for asia.gcr.io/cloud-shell-215704/kaniko-dockerfile_test_volume Layer 3:
FILE             SIZE
/baz             0
/baz/bat         0
/foo             0
/foo/bar         0
/qux             0
/qux/quux        0
/tmp             0
```

peter-evans:
I rebased onto master and ran the integration tests again. All pass.

pkg/snapshot/snapshot.go
```diff
@@ -123,11 +107,7 @@ func (s *Snapshotter) snapshotFiles(f io.Writer, files []string) (bool, error) {
 	}
 	for _, file := range parentDirs {
 		file = filepath.Clean(file)
-		shouldSnapshot, err := shouldSnapshot(file, snapshottedFiles)
-		if err != nil {
-			return false, fmt.Errorf("Error checking if parent dir %s can be snapshotted: %s", file, err)
-		}
-		if !shouldSnapshot {
+		if val, ok := snapshottedFiles[file]; ok && val {
```
priyawadhwa: Why did you remove shouldSnapshot() and isBuildFile()? I think those functions are necessary so that kaniko is able to build itself.

peter-evans: For the specific files cases (ADD, COPY, WORKDIR) that call `snapshotFiles()` it is correct that they are added to the snapshot even if one of their parent directories is a volume. That is the main reason I removed checking the whitelist, which is what `shouldSnapshot()` was doing. Now that we aren't checking the whitelist there is no need for `isBuildFile()` either. That function's job was to make sure that build files that were whitelisted were added to the snapshot anyway.

The key thing to consider about this change is, is it ok to not check the whitelist at all for specific files cases, directives ADD, COPY and WORKDIR? (VOLUME is included too, but only if it's snapshot with another specific file case and not a system-wide snapshot.) Up to now, Kaniko has prevented adding whitelisted files and directories with those directives. It doesn't seem to be necessary to me. The main purpose of the whitelist seems to be to ignore files and directories when there is no way to know if the file was mounted or came from the base image. If a Dockerfile is explicitly trying to add files and directories with ADD, COPY and WORKDIR, I don't think there is a reason to prevent that. Let me know if you think I've missed something!

priyawadhwa: @peter-evans that makes sense, thanks for explaining!

PR #336, opened 2018-09-04
Add CacheCommand to DockerCommand interface

priyawadhwa:
CacheCommand returns true if the command should be cached. Currently,
it's only true for RUN but can be added to ADD/COPY later on (these are
different since the contents of files for ADD/COPY need to be included
in the cache key generation).

I also changed CreatedBy to String so that we can log each command
before cache extraction or regular execution takes place.

### DIFF
--- a/pkg/commands/add.go
+++ b/pkg/commands/add.go
@@ -18,7 +18,6 @@ package commands
 
 import (
 	"path/filepath"
-	"strings"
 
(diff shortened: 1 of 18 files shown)


priyawadhwa:
For sure -- once some other open PRs are merged (#320)  I'll start to piece all of these together and implement caching for kaniko. I'll use CacheCommand to determine if a command should be cached or not (part 1 will only cache RUN commands, later on we could add ADD/COPY).

PR #337, opened 2018-09-04
Add Key() to LayeredMap and Snapshotter

priyawadhwa:
This will return a string representaiton of the current filesystem to be
used with caching.

Whenever a file is explictly added (via ADD or COPY), it will be stored
in "added" in the LayeredMap. The file will map to a hash created by
CacheHasher (which doesn't take into account mtime, since that will be
different with every build, making the cache useless)

Key() will returns a sha of the added files which will be used in
determining the overall cache key for a command.

### DIFF
--- a/pkg/snapshot/layered_map.go
+++ b/pkg/snapshot/layered_map.go
@@ -27,10 +27,11 @@ import (
 )
 
 type LayeredMap struct {
-	layers      []map[string]string
-	whiteouts   []map[string]string
-	added       []map[string]string
-	hasher      func(string) (string, error)
+	layers    []map[string]string
+	whiteouts []map[string]string
+	added     []map[string]string
+	hasher    func(string) (string, error)
+	// cacheHasher doesn't include mtime in it's hash so that filesystem cache keys are stable
 	cacheHasher func(string) (string, error)
 }
 
--- a/pkg/snapshot/layered_map.go
+++ b/pkg/snapshot/layered_map.go
@@ -17,20 +17,27 @@ limitations under the License.
 package snapshot
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"strings"
+
+	"github.com/GoogleContainerTools/kaniko/pkg/util"
 )
 
 type LayeredMap struct {
-	layers    []map[string]string
-	whiteouts []map[string]string
-	hasher    func(string) (string, error)
+	layers      []map[string]string
+	whiteouts   []map[string]string
+	added       []map[string]string
+	hasher      func(string) (string, error)
+	cacheHasher func(string) (string, error)
 }
 
-func NewLayeredMap(h func(string) (string, error)) *LayeredMap {
+func NewLayeredMap(h func(string) (string, error), c func(string) (string, error)) *LayeredMap {
 	l := LayeredMap{
-		hasher: h,
+		hasher:      h,
+		cacheHasher: c,
 	}
 	l.layers = []map[string]string{}
 	return &l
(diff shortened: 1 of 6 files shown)


pkg/snapshot/layered_map.go
```diff
@@ -39,8 +46,18 @@ func NewLayeredMap(h func(string) (string, error)) *LayeredMap {
 func (l *LayeredMap) Snapshot() {
 	l.whiteouts = append(l.whiteouts, map[string]string{})
 	l.layers = append(l.layers, map[string]string{})
+	l.added = append(l.added, map[string]string{})
+}
+
+// Key returns a hash for added files
+func (l *LayeredMap) Key() (string, error) {
+	c := bytes.NewBuffer([]byte{})
+	enc := json.NewEncoder(c)
```
dlorenc: It's not clear to me why we're json encoding anything. Is the idea to get a stable hash for a map[string]string?

I'm not sure if the sorting of a map when encoding it will be stable...

pkg/snapshot/layered_map.go
```diff
@@ -85,11 +102,18 @@ func (l *LayeredMap) MaybeAddWhiteout(s string) (bool, error) {
 
 // Add will add the specified file s to the layered map.
 func (l *LayeredMap) Add(s string) error {
```
tejal29: is the name of the file or contents?

priyawadhwa: `s` is the name of the file

priyawadhwa: sorry, is that what you were asking?

pkg/snapshot/layered_map_test.go
```diff
@@ -0,0 +1,78 @@
...
+			lm2 := LayeredMap{added: []map[string]string{test.map2}}
+			k1, err := lm1.Key()
+			if err != nil {
+				t.Fatalf("error getting key for map 1: %v", err)
+			}
+			k2, err := lm2.Key()
+			if err != nil {
+				t.Fatalf("error getting key for map 2: %v", err)
+			}
+			if test.equal && k1 != k2 {
```
tejal29: You could avoid t.Fatalf by adding this 1 check instead of checking them separately.
```
if test.equal  != (k1 == k2) {
t.Errof("keys expected to be same: %t. Got %t", test.equal, k1==k2)
}
```
That way, the other test cases still keep running

pkg/util/util.go
```diff
@@ -86,3 +117,10 @@ func MtimeHasher() func(string) (string, error) {
 	}
 	return hasher
 }
+
+// SHA256 returns the shasum of the contents of r
+func SHA256(r io.Reader) (string, error) {
+	hasher := sha256.New()
+	_, err := io.Copy(hasher, r)
```
tejal29: return error if its not nil.
the next line might throw an exception.

pkg/snapshot/layered_map.go
```diff
@@ -17,20 +17,27 @@ limitations under the License.
...
 
 type LayeredMap struct {
-	layers    []map[string]string
-	whiteouts []map[string]string
-	hasher    func(string) (string, error)
+	layers      []map[string]string
+	whiteouts   []map[string]string
+	added       []map[string]string
+	hasher      func(string) (string, error)
+	cacheHasher func(string) (string, error)
```
tejal29: What is the difference between hasher and cacheHasher?

pkg/snapshot/layered_map.go
tejal29: yes. i was just wondering if we shd be hashing the file contents as well.

priyawadhwa: we call the hasher function on line 106:
```
newV, err := l.hasher(s)
```
and that function will read the file contents, modtime, and some other things as well

tejal29: hmm.. After reading the PR description again, i think the cacheHasher does it.

priyawadhwa: yah, it does!

pkg/snapshot/layered_map.go
tejal29: nvm, read the description again. Can you add comments here describing hasher and cacheHasher?

priyawadhwa: cacheHasher doesn't include file mtime in the hash, since that would be different for every build and we need a stable key for caching

pkg/util/util.go
priyawadhwa: done

pkg/snapshot/layered_map_test.go
priyawadhwa: thanks, just changed it!

I think t.Fatalf will only kill the current test in this case, since it's running within t.Run

pkg/snapshot/layered_map.go
priyawadhwa: added the comment, there's also more detailed descriptions in util.go where the hash functions live

pkg/util/util.go
priyawadhwa: done

PR #341, opened 2018-09-05
set default HOME env properly

### LINKED ISSUE #333: USER $HOME uses uid instead of username
It seems that $HOME is [set to the value of the USER instruction](https://github.com/GoogleContainerTools/kaniko/blob/2e10d2761cebd268e6f80e31a401e6039dd0d257/pkg/commands/run.go#L131).  

This causes problems with [OpenShift compatible images](https://docs.openshift.com/enterprise/3.0/creating_images/guidelines.html#openshift-specific-guidelines)
> the final USER declaration in the Dockerfile should specify the user ID (numeric value) and not the user name

Using the uid causes unexpected behaviour:
```
USER 1000

RUN touch ~/invalid
```
output
```
INFO[0051] cmd: /bin/sh                                 
INFO[0051] args: [-c touch ~/invalid]                   
touch: cannot touch '/home/1000/invalid': No such file or directory
Error: error building image: waiting for process to exit: exit status 1
```

Workaround is to set $HOME explicitly with `ARG` or `ENV`, but it would be great if $HOME was resolved to the users $HOME.

Thanks for the great work!

sharifelgamal:
Fixes #333


pkg/commands/run.go
```diff
@@ -116,19 +117,28 @@ func (r *RunCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
...
 	// If user is set, set value of HOME to /home/${user}
-	return append(envs, fmt.Sprintf("%s=/home/%s", constants.HOME, user))
+	// If uid is saved instead of username, lookup the user and supply the username
+	userObj, err := user.LookupId(u)
+	// The user is guaranteed to exist, so if err is not nil, then the username is set in the config
+	if err == nil {
+		u = userObj.Username
+	}
+
+	return append(envs, fmt.Sprintf("%s=/home/%s", constants.HOME, u))
```
priyawadhwa: I think that if the UID is set instead of username, the value of $HOME defaults to root `/`

I tried it out with this Dockerfile:
```
FROM ubuntu
USER 1000
CMD echo $HOME
```

and ran it:
```
$ docker run test
/
```
that's what seems like is happening. We still ignore Dockerfile_test_user_run in the integration tests, which is they didn't catch it.

I couldn't find any documentation to verify this unfortunately, it just says that HOME is determined by USER [here](https://docs.docker.com/engine/reference/run/#env-environment-variables) :disappointed:

sharifelgamal: Yeah, I can verify that that is docker's behavior, silly as it is. Should we default to that? It seems like bad behavior since specifying username sets $HOME properly.

dlorenc: Any updates on this? I think copying Docker is probably the right thing to do by default.

Issue #346, opened 2018-09-10
setting entrypoint should clear cmd from child image


mathroc:
**Actual behavior**

when an entrypoint is set, the cmd from the child is not cleard

**Expected behavior**

the cmd should be cleared unless it was set in the same Dockerfile

see https://github.com/docker/docker.github.io/issues/6142

bobcatfish:
I think this was resolved in #348, re-open if I'm crazy!

PR #348, opened 2018-09-11
Review config for cmd/entrypoint after building a stage

priyawadhwa:
As mentioned in #346, if only ENTRYPOINT is set in a stage then any
CMD inherited from a parent should be cleared.

If both entrypoint and cmd are set then nothing should change.

I added a function and unit test to review the config file after building a stage
which clears out config.Cmd if ENTRYPOINT was declared but CMD wasn't.

I also added an integration test to make sure this works, which should
be tested by the preexisting container-diff --metadata test.

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -145,6 +145,9 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 				return nil, err
 			}
 		}
+		if err := reviewConfig(stage, &imageConfig.Config); err != nil {
+			return nil, err
+		}
 		sourceImage, err = mutate.Config(sourceImage, imageConfig.Config)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


pkg/executor/build.go
```diff
@@ -228,3 +231,24 @@ func resolveOnBuild(stage *config.KanikoStage, config *v1.Config) error {
...
+
+// reviewConfig makes sure the value of CMD is correct after building the stage
+// If ENTRYPOINT was set in this stage but CMD wasn't, then CMD should be cleared out
+// See Issue #346 for more info
+func reviewConfig(stage config.KanikoStage, config *v1.Config) error {
+	entrypoint := false
+	cmd := false
+
+	for _, c := range stage.Commands {
+		if c.Name() == "cmd" {
```
tejal29: use constants.

pkg/executor/build.go
```diff
@@ -228,3 +231,24 @@ func resolveOnBuild(stage *config.KanikoStage, config *v1.Config) error {
 	config.OnBuild = nil
 	return nil
 }
+
+// reviewConfig makes sure the value of CMD is correct after building the stage
+// If ENTRYPOINT was set in this stage but CMD wasn't, then CMD should be cleared out
+// See Issue #346 for more info
+func reviewConfig(stage config.KanikoStage, config *v1.Config) error {
```
tejal29: This function does not encounter an error flow and always returns nil.
There is no need to return an error.

pkg/executor/build_test.go
```diff
@@ -0,0 +1,76 @@
...
+		originalEntrypoint []string
+		expectedCmd        []string
+	}{
+		{
+			name: "entrypoint and cmd declared",
+			dockerfile: `
+			FROM scratch
+			CMD ["mycmd"]
+			ENTRYPOINT ["myentrypoint"]`,
+			originalEntrypoint: []string{"myentrypoint"},
```
tejal29: by originalCmd, do you mean command from previous stage?

tejal29: use different values for dockerfile.CMD and dockerfile.ENTRYPOINT.

priyawadhwa: by originalCmd, I mean the original value of Cmd in the config file (it should match the value specified in the Dockerfile)

The function reviews the Dockerfile and then clears the Cmd field if it isn't declared in the Dockerfile.

pkg/executor/build.go
priyawadhwa: done

pkg/executor/build.go
priyawadhwa: done

PR #349, opened 2018-09-11
Replace gometalinter with GolangCI-Lint

priyawadhwa:
gometalinter is broken @ HEAD, and I looked into why that was. During
that process, I remembered that we took the linting scripts from
skaffold, and found that in skaffold gometalinter was replaced with
GolangCI-Lint:

https://github.com/GoogleContainerTools/skaffold/pull/619

The change made linting in skaffold faster, so I figured instead of
fixing gometalinter it made more sense to remove it and replace it with
GolangCI-Lint for kaniko as well.


integration/integration_test.go
```diff
@@ -240,30 +239,26 @@ func TestLayers(t *testing.T) {
...
-				t.Fail()
-			}
+			checkLayers(t, dockerImage, kanikoImage, offset[dockerfile])
 		})
 	}
 }
 
-func checkLayers(t *testing.T, image1, image2 string, offset int) error {
+func checkLayers(t *testing.T, image1, image2 string, offset int) {
 	img1, err := getImageDetails(image1)
```
tstromberg: I'm not a big fan of custom assert methods, but this should at least call t.Helper().

Issue #366, opened 2018-09-26
FR: Add a flag to print timing data at the end of a build


dlorenc:
Some interesting info to breakdown:

- Base image pulling time
- Extraction time
- Layer pushing time
- RUN command execution time
- Snapshot time
- Other stuff?

It would be nice to have a flag to dump all this info at the end of a build, to help us diagnose and debug slow builids.

PR #367, opened 2018-09-26
Add a benchmark package to store and monitor timings.

dlorenc:
Ref #366


dlorenc:
This will be used to implement the work outlined in #366

PR #369, opened 2018-09-27
Whitelist /busybox in the debug image

vbehar:
In the debug image, declare /busybox as a volume so that it is automatically whitelisted, because we don't want to delete it when building multi-stages images.

FYI this is required when using Jenkins, because we need to use the debug kaniko image to be able to start the container "doing nothing" (with /busybox/cat) before building (by executing /kaniko/executor directly inside the container)
See https://issues.jenkins-ci.org/browse/JENKINS-52576


priyawadhwa:
Hey @vbehar, instead of hardcoding `/busybox` into the whitelist, which may negatively impact people trying to use that directory with the standard (non-debug) image, let's instead add

`VOLUME /busybox`
to the debug Dockerfile [here](https://github.com/GoogleContainerTools/kaniko/blob/master/deploy/Dockerfile_debug#L41), which should automatically add it to the whitelist for the debug image only.

vbehar:
@priyawadhwa ok thanks, yes that's way better (and easier) that way!

vbehar:
Thanks for your review and your good work on Kaniko !

kumprj:
I'm still experiencing this issue in Jenkins vbehar. Jenkins on Kubernetes, starting the kaniko container with /busybox/cat. In my pipeline I run most of my job on a linux agent, but want to swap to kaniko container for docker images. Receiving the nohup error. Any advice for how you got it working?

Trying to run even a sh 'ls' returns '/busybox/sh: nohup: not found'

PR #370, opened 2018-09-27
Add a new flag to cleanup the filesystem at the end

vbehar:
Currently, kaniko can only build a single image per container run, because the filesystem is full of the content of the first image.
When running kaniko in Jenkins, where we need to start the container "doing nothing" first (using the debug kaniko container), and then exec /kaniko/executor, this is a limitation because it means that if we want to build multiple images, we need to start multiple containers - see https://groups.google.com/forum/#!topic/kaniko-users/_7LivHdMdy0 for more details

A solution to fix this issue is to add a new flag to cleanup the filesystem at the end - the same way it is done between stages when building a multi-stages image. This way, the same (debug) container can be used to build multiple images.

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -97,6 +97,7 @@ func addKanikoOptionsFlags(cmd *cobra.Command) {
 	RootCmd.PersistentFlags().BoolVarP(&opts.NoPush, "no-push", "", false, "Do not push the image to the registry")
 	RootCmd.PersistentFlags().StringVarP(&opts.CacheRepo, "cache-repo", "", "", "Specify a repository to use as a cache, otherwise one will be inferred from the destination provided")
 	RootCmd.PersistentFlags().BoolVarP(&opts.Cache, "cache", "", false, "Use cache when building image")
+	RootCmd.PersistentFlags().BoolVarP(&opts.Cleanup, "cleanup", "", false, "Clean the filesystem at the end")
 }
 
 // addHiddenFlags marks certain flags as hidden from the executor help text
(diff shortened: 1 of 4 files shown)


svalo:
This would be awesome! I'm using GitLab CI, I need to build 2 images, the second depending on the first. To be able to use kaniko I had  to create 2 stages because GitLab's jobs can't be sequential. If this gets merged everything could be achieved in a single job

cmd/executor/cmd/root.go
```diff
@@ -37,11 +37,13 @@ var (
 	opts     = &config.KanikoOptions{}
 	logLevel string
 	force    bool
+	cleanup  bool
```
priyawadhwa: Instead of creating a new variable here, could we add this to [KanikoOptions](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/config/options.go#L20)? This should help with executing the comment below.

cmd/executor/cmd/root.go
```diff
@@ -77,6 +79,11 @@ var RootCmd = &cobra.Command{
 		if err := executor.DoPush(image, opts); err != nil {
 			exit(errors.Wrap(err, "error pushing image"))
 		}
+		if cleanup {
```
priyawadhwa: Instead of having this in `root.go`, I think it would be better placed in [this](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/build.go#L256) if statement in `build.go`

If opts.Cleanup is `true`, then delete the filesystem before returning the image.

vbehar:
@priyawadhwa ok, thanks for the review! I changed the implementation and squashed.

PR #371, opened 2018-09-27
Fixes a whitelist issue when untarring files in ADD commands.

dlorenc:
I think this fixes an issue with extracting tar files during ADD commands where those tar files contain whitelisted entries.


priyawadhwa:
Looks like kokoro caught something

```
FAIL: TestRun/test_Dockerfile_test_add (49.00s)
    	integration_test.go:203: diff = [
    		  {
    		    "Image1": "gcr.io/kaniko-test/docker-dockerfile_test_add",
    		    "Image2": "gcr.io/kaniko-test/kaniko-dockerfile_test_add",
    		    "DiffType": "File",
    		    "Diff": {
    		      "Adds": null,
    		      "Dels": [
    		        {
    		          "Name": "/sys/fs",
    		          "Size": 0
    		        },
    		        {
    		          "Name": "/sys/fs/foo",
    		          "Size": 0
    		        }
    		      ],
    		      "Mods": null
    		    }
    		  },
    		  {
    		    "Image1": "gcr.io/kaniko-test/docker-dockerfile_test_add",
    		    "Image2": "gcr.io/kaniko-test/kaniko-dockerfile_test_add",
    		    "DiffType": "Metadata",
    		    "Diff": {
    		      "Adds": [],
    		      "Dels": []
    		    }
    		  }
    		]
```

integration-test.sh
```diff
@@ -35,4 +35,4 @@ fi
 echo "Running integration tests..."
 make out/executor
 pushd integration
-go test -v --bucket "${GCS_BUCKET}" --repo "${IMAGE_REPO}" --timeout 30m
+go test -v --bucket "${GCS_BUCKET}" --repo "${IMAGE_REPO}" --timeout 30m -run ^TestRun/Dockerfile_test_add$
```
priyawadhwa: I think this may be left over from testing?

dlorenc: D'oh. Thanks!

priyawadhwa:
Cool so kokoro found this:
```
--- FAIL: TestLayers (276.84s)
    --- FAIL: TestLayers/test_layer_Dockerfile_test_add (14.16s)
    	integration_test.go:237: Difference in number of layers in each image is 11 but should be 10. Image 1: Image: [gcr.io/kaniko-test/docker-dockerfile_test_add] Digest: [a583f57d4c7e97863a73954ab33e2f16dd5065a55ed4c308fc7c99939562f8f8] Number of Layers: [13], Image 2: Image: [gcr.io/kaniko-test/kaniko-dockerfile_test_add] Digest: [9b9f5909da8e19a34c3c93589cea3d45e3acca74e2fcb8dcc5a6ceaf2128a30e] Number of Layers: [2]
```

This is one of the images we test the `--single-snapshot` flag on, so we just need to update the expected [offset](https://github.com/GoogleContainerTools/kaniko/blob/master/integration/integration_test.go#L214) to 11 since a layer was added to the Dockerfile.

PR #375, opened 2018-09-29
Rework cache key generation a bit.

dlorenc:
Cache keys are now based on the previous commands, rather than the previous state
of the filesystem.

This will allow us to add some further optimizations to caching.

Still needs some tests.

### DIFF
--- a/pkg/commands/add.go
+++ b/pkg/commands/add.go
@@ -29,6 +29,7 @@ import (
 )
 
 type AddCommand struct {
+	BaseCommand
 	cmd           *instructions.AddCommand
 	buildcontext  string
 	snapshotFiles []string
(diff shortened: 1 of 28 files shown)


priyawadhwa:
I really like the CompositeCache idea! So I was using the current state of the filesystem for Dockerfiles like this:

```
RUN apt-get update && apt-get install -y make
COPY foo /foo
```
With this switch, if foo is changed then the entire Dockerfile will be rebuilt, even though the first RUN command may have been cached on a previous run and could still be used. 

Instead of the entire source context, WDYT of adding the [snapshotted files](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/build.go#L169) as they're added by each command? (I think that should work but not 100% sure).

dlorenc:
Whoo, finally passing!

PR #381, opened 2018-10-03
Don't cut everything after an equals sign

blacksails:
I tried to provide a build build arg like this: `--build-arg ARG_WITH_EQUALS="This has an = char"` which results in `$ARG_WITH_EQUALS` having a value of `This has an `


blacksails:
I signed it!

blacksails:
Just fyi i pushed an image with the fix to lixtech/kaniko, and i seems to work as intended.

PR #383, opened 2018-10-03
Persistent volume caching for base images

sharifelgamal:


cmd/warmer/cmd/root.go
```diff
@@ -0,0 +1,74 @@
...
+func addKanikoOptionsFlags(cmd *cobra.Command) {
+	RootCmd.PersistentFlags().VarP(&opts.Images, "image", "i", "Image to cache. Set it repeatedly for multiple images.")
+	RootCmd.PersistentFlags().StringVarP(&opts.CacheDir, "cache-dir", "c", "/cache", "Directory of the cache.")
+}
+
+// addHiddenFlags marks certain flags as hidden from the executor help text
+func addHiddenFlags(cmd *cobra.Command) {
+}
+
+func exit(err error) {
```
dlorenc: nit: we shouldn't need this if we return an error

pkg/cache/cache.go
```diff
@@ -18,19 +18,25 @@ package cache
...
 	"github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
 // RetrieveLayer checks the specified cache for a layer with the tag :cacheKey
 func RetrieveLayer(opts *config.KanikoOptions, cacheKey string) (v1.Image, error) {
+	if local, _ := LocalDestination(opts, cacheKey); local != nil {
```
dlorenc: Maybe "source" instead of "destination"?

pkg/cache/warm.go
```diff
@@ -0,0 +1,72 @@
...
+	fmt.Printf("WARMER: cachedir = %s\n", cacheDir)
+	fmt.Printf("WARMER: images = %v\n", images)
+
+	for _, image := range images {
+		cacheRef, err := name.NewTag(image, name.WeakValidation)
+		if err != nil {
+			fmt.Print(err)
+			errors.Wrap(err, fmt.Sprintf("Failed to verify image name: %s", image))
+		}
+		k8sc, err := k8schain.NewNoClient()
```
dlorenc: hmm, should we use a normal keychain if we're not on k8s?

deploy/Dockerfile_warmer
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Builds the static Go image to execute in a Kubernetes job
+
+FROM golang:1.10
+WORKDIR /go/src/github.com/GoogleContainerTools/kaniko
+COPY . .
+RUN make
+# Get GCR credential helper
```
dlorenc: Do we need the helpers if we're relying on open auth?

sharifelgamal: nope. removed.

pkg/util/image_util.go
```diff
@@ -106,7 +105,22 @@ func cachedImage(opts *config.KanikoOptions, image string) (v1.Image, error) {
 	if err != nil {
 		return nil, err
 	}
-	cacheKey := ref.Name()
-	fmt.Printf("CACHEKEY=%s", cacheKey)
-	return cache.LocalDestination(opts, cacheKey)
+
+	k8sc, err := k8schain.NewNoClient()
```
dlorenc: I think there's an easier way to do this with no auth: https://github.com/google/go-containerregistry/blob/d54baf9aa28edb9b985a6b35b57e26e3410c2443/pkg/authn/anon.go

PR #385, opened 2018-10-08
Refactor the build loop.

dlorenc:
This change refactors the build loop a bit to make cache optimization easier in the future. Some notable changes:

- The special casing around volume snapshots is removed. Every volume is added to the snapshotFiles list for every command that will snapshot anyway.
-  Snapshot saving was extracted to a sub-function


dlorenc:
cc @priyawadhwa

dlorenc:
OK, tests finally passing!

PR #387, opened 2018-10-09
Use only the necessary files in the cache keys.

dlorenc:


dlorenc:
> Nice! LGTM, just one question :)

I don't think your question made it through :(

priyawadhwa:
Whoops I think I forgot it anyway, no worries :sweat_smile:

PR #388, opened 2018-10-10
Update go-containerregistry dependency

### LINKED ISSUE #245: Long builds fail with "UNAUTHORIZED: \"authentication required\""
## steps to reproduce

```bash
mkdir -p ./kaniko-issue

cat > ./kaniko-issue/Dockerfile <<EOF
FROM debian:stable-slim
RUN sleep 1
EOF
# This will work as expected
docker run -v `pwd`/kaniko-issue/:/workspace \
    gcr.io/kaniko-project/executor:latest \
    -c /workspace -f Dockerfile \
    -d registry.example.com/does/not:matter \
    --tarPath /workspace/tarball.tar

cat > ./kaniko-issue/Dockerfile <<EOF
FROM debian:stable-slim
RUN sleep 360
EOF
# This will fail with "UNAUTHORIZED: \"authentication required\""
docker run -v `pwd`/kaniko-issue/:/workspace \
    gcr.io/kaniko-project/executor:latest \
    -c /workspace -f Dockerfile \
    -d registry.example.com/does/not:matter \
    --tarPath /workspace/tarball.tar

# rm -r ./kaniko-issue
```

## additional obervations

I also ran `tcpdump` on the network interface of the container. I saw a quite a bit of traffic at the start (I assume pulling the image) and a single, short TLS connection to index.docker.io after `sleep` was done.

The issue seems to be gone (or at least takes substantially longer to arise) if I substitute `debian:stable-slim` with any image from my [harbor (private docker registry)](https://github.com/vmware/harbor/) instance.

## working theory

My working theory of the underlying cause based on those two observations is, that kaniko tries to fetch the image config of the base image using an expired bearer token. This config would normally be extended and included in the tarball or pushed to the registry.

ianberinger:
fixes #245

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -431,7 +431,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:edf64d541c12aaf4f279642ea9939f035dcc9fc2edf649aba295e9cbca2c28d4"
+  digest = "1:764d0a9bb2c987d9333c5b0a256bc94791db50c0b5be2fa10f2247b1dcbd7a04"
   name = "github.com/google/go-containerregistry"
   packages = [
     "pkg/authn",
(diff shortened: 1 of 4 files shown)


yurrriq:
Thanks, @ianberinger!

Issue #380, opened 2018-10-02
Issue with symlink during layer extraction


devderek:
**Actual behavior**
I'm experiencing an issue when downloading and extracting layers during a build. I see the following error:
> error building image: error building stage: error removing opt/eap/standalone/deployments to make way for new symlink: remove /opt/eap/standalone/deployments: directory not empty

This occurs when I'm downloading this image:
> registry.access.redhat.com/jboss-eap-7/eap71-openshift:1.3-10

This behavior is not experienced when performing _docker pull_ then _docker run_.

**Expected behavior**
The image downloads and extracts without symlink errors.

**To Reproduce**
In a fresh folder run the following commands:

> echo "FROM registry.access.redhat.com/jboss-eap-7/eap71-openshift:1.3-10" > Dockerfile

> docker run -it -v "$(pwd):/workspace" --entrypoint "/kaniko/executor" gcr.io/kaniko-project/executor@sha256:f22531406dfd38b6f84d9310cd713d26c623ae152ced8e4ac490734c3fdb66b5 --no-push

You should see output similar to the following:

> Unable to find image 'gcr.io/kaniko-project/executor@sha256:f22531406dfd38b6f84d9310cd713d26c623ae152ced8e4ac490734c3fdb66b5' locally
sha256:f22531406dfd38b6f84d9310cd713d26c623ae152ced8e4ac490734c3fdb66b5: Pulling from kaniko-project/executor
3f7027d6fbd3: Pull complete
b8ca5337ef48: Pull complete
7db8e39884c2: Pull complete
a4562b062142: Pull complete
717aa8849da8: Pull complete
1e5beb4c8e0d: Pull complete
afe73cead066: Pull complete
Digest: sha256:f22531406dfd38b6f84d9310cd713d26c623ae152ced8e4ac490734c3fdb66b5
Status: Downloaded newer image for gcr.io/kaniko-project/executor@sha256:f22531406dfd38b6f84d9310cd713d26c623ae152ced8e4ac490734c3fdb66b5
INFO[0000] Downloading base image registry.access.redhat.com/jboss-eap-7/eap71-openshift:1.3-10
2018/10/02 21:14:07 No matching credentials were found, falling back on anonymous
INFO[0001] Executing 0 build triggers
INFO[0001] Extracting layer 0
INFO[0012] Extracting layer 1
INFO[0012] Extracting layer 2
INFO[0013] Extracting layer 3
INFO[0021] Extracting layer 4
[...]

PR #360, opened 2018-09-18
Enable overwriting of links (solves #351)

xanonid:

PR #389, opened 2018-10-11
Fix symlink extraction

peter-evans:
Fixes https://github.com/GoogleContainerTools/kaniko/issues/380

The issue seems to be that `os.Remove` can only remove a file or empty directory. Changing it to `os.RemoveAll` removes everything at the path making way for the extracted symlink.

I've tested this with the example provided by @devderek.

```bash
$ echo “FROM registry.access.redhat.com/jboss-eap-7/eap71-openshift:1.3-10” > Dockerfile
$ docker run -it -v “$(pwd):/workspace” --entrypoint “/kaniko/executor” executor:fix --no-push
INFO[0000] Downloading base image registry.access.redhat.com/jboss-eap-7/eap71-openshift:1.3-10
ERROR: logging before flag.Parse: E1011 00:34:04.660699       1 metadata.go:159] while reading ‘google-dockercfg-url’ metadata: http status code: 404 while fetching url http://metadata.google.internal./computeMetadata/v1/instance/attributes/google-dockercfg-url
ERROR: logging before flag.Parse: E1011 00:34:04.664735       1 metadata.go:142] while reading ‘google-dockercfg’ metadata: http status code: 404 while fetching url http://metadata.google.internal./computeMetadata/v1/instance/attributes/google-dockercfg
2018/10/11 00:34:04 No matching credentials were found, falling back on anonymous
INFO[0001] Executing 0 build triggers
INFO[0001] Extracting layer 0
INFO[0008] Extracting layer 1
INFO[0008] Extracting layer 2
INFO[0009] Extracting layer 3
INFO[0014] Extracting layer 4
INFO[0025] Extracting layer 5
INFO[0042] Taking snapshot of full filesystem...
INFO[0052] Skipping push to container registry due to --no-push flag
```

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_extraction
+++ b/integration/dockerfiles/Dockerfile_test_extraction
@@ -1,2 +1,2 @@
-# Tests extraction of a symlink to a path that is a non-empty directory
-FROM registry.access.redhat.com/jboss-eap-7/eap71-openshift:1.3-10
+# Tests extraction of symlink, hardlink and regular files to a path that is a non-empty directory
+FROM gcr.io/kaniko-test/extraction-base-image:latest
--- a/integration/dockerfiles/Dockerfile_test_extraction
+++ b/integration/dockerfiles/Dockerfile_test_extraction
@@ -1,0 +1,2 @@
+# Tests extraction of a symlink to a path that is a non-empty directory
+FROM registry.access.redhat.com/jboss-eap-7/eap71-openshift:1.3-10
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -193,7 +193,7 @@ func extractFile(dest string, hdr *tar.Header, tr io.Reader) error {
 		// Check if something already exists at path (symlinks etc.)
 		// If so, delete it
 		if FilepathExists(path) {
-			if err := os.Remove(path); err != nil {
+			if err := os.RemoveAll(path); err != nil {
 				return errors.Wrapf(err, "error removing %s to make way for new file.", path)
 			}
 		}
@@ -242,7 +242,7 @@ func extractFile(dest string, hdr *tar.Header, tr io.Reader) error {
 		// Check if something already exists at path
 		// If so, delete it
 		if FilepathExists(path) {
-			if err := os.Remove(path); err != nil {
+			if err := os.RemoveAll(path); err != nil {
 				return errors.Wrapf(err, "error removing %s to make way for new link", hdr.Name)
 			}
 		}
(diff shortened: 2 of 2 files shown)


priyawadhwa:
Hey @peter-evans thanks for adding this fix! Could you add a `Dockerfile_test_extraction` to this [directory](https://github.com/GoogleContainerTools/kaniko/tree/master/integration/dockerfiles) which just pulls in this image so our integration tests can make sure this works from now on? Thanks!

peter-evans:
I just noticed that this code was added recently:
https://github.com/GoogleContainerTools/kaniko/pull/360/files#diff-d36eb675aa49a7b471e3a2be77005b18

A similar code block also has been added to remove existing files for regular file type extraction (`tar.TypeReg`) as well as symlinks. Should I update the regular file case block to match?

peter-evans:
I've confirmed that the following image when pulled and extracted also fails because of the same issue in the `tar.TypeReg` case block.

```
RUN mkdir /foo && touch /foo/bar
RUN rm -fr /foo && touch /foo
```

When the above image is built and tagged as `asia.gcr.io/cloud-shell-215704/extractregtest`:
```
$ echo "FROM asia.gcr.io/cloud-shell-215704/extractregtest" > Dockerfile
$ docker run -it -v "$(pwd):/workspace" --entrypoint "/kaniko/executor" executor:latest --no-push
INFO[0000] Downloading base image asia.gcr.io/cloud-shell-215704/extractregtest 
INFO[0000] Extracting layer 0                           
INFO[0003] Extracting layer 1                           
INFO[0003] Extracting layer 2                           
error building image: error building stage: error removing /foo to make way for new file.: remove /foo: directory not empty
```

So I'll update the `tar.TypeReg` block as well to fix this. Are you able to push the test image to a registry so that I can pull it in an integration test Dockerfile? It's probably best if that is hosted somewhere official rather than my registry.

peter-evans:
Same issue with `tar.TypeLink` so I've updated that too.

I've used the example image that @devderek provided in the integration test for the symlink case. However, if you want to have an integration test for each of the three cases you could push the following images to a registry and I'll create integration tests that use them.

symlink:
```
RUN mkdir /foo && touch /foo/bar
RUN rm -fr /foo && touch /baz && ln -s /baz /foo
```

hardlink:
```
RUN mkdir /foo && touch /foo/bar
RUN rm -fr /foo && touch /baz && ln /baz /foo
```

regular files:
```
RUN mkdir /foo && touch /foo/bar
RUN rm -fr /foo && touch /foo
```

priyawadhwa:
Cool, I pushed the image as `gcr.io/kaniko-test/extraction-base-image:latest`

peter-evans:
@priyawadhwa Thank you! I updated `Dockerfile_test_extraction` so this should be ready now.

josephndenton:
Can we get this merged in?

PR #394, opened 2018-10-12
Add support for .dockerignore file

### LINKED ISSUE #384: Ignore files in .dockerignore
I am using [Kaniko in Gitlab](https://docs.gitlab.com/ee/ci/docker/using_kaniko.html) and Kaniko does not seem to use my `.dockerignore.` file. This means it's adding my Git repository to the image. Is it possible to make Kaniko ignore certain files?

priyawadhwa:
kaniko should support the .dockerignore file. If it finds one in the build context, it will parse the file and delete all matching paths from the build context. It will then delete the .dockerignore before starting the build.
Should fix #384

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,10 @@ _This flag must be used in conjunction with the `--cache=true` flag._
 
 Set this flag to cleanup the filesystem at the end, leaving a clean kaniko container (if you want to build multiple images in the same container, using the debug kaniko image)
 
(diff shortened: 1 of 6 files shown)


dlorenc:
What's the overall plan here? Is kaniko going to add support for parsing .dockerignore files, or are we doing to expect users to set this flag?

PR #397, opened 2018-10-15
More cache cleanups:

dlorenc:
- move the layer cache to an interface
- refactor the DockerCommand implementations to support Cached and non-cached implementations.


dlorenc:
This should be ready for review now! cc @priyawadhwa @sharifelgamal

PR #398, opened 2018-10-15
Allow using --tarPath without --destination

YorikSar:
Currently user has to specify `--destination` argument even if one wants to generate tarball with resulting image:

```
% docker run -v $PWD:/context -v $PWD/out:/out gcr.io/kaniko-project/executor:v0.4.0 --context=/context --tarPath=/out/output.tar
Error: You must provide --destination, or use --no-push
Usage:
...
```

This can be worked around by specifying `--destination=nothing`:

```
% docker run -v $PWD:/context -v $PWD/out:/out gcr.io/kaniko-project/executor:v0.4.0 --context=/context --tarPath=/out/output.tar --destination=nothing
INFO[0000] Downloading base image ...
```

This change removes need for such workaround since `--destination` is ignored when `--tarPath` is present anyway.

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -52,8 +52,8 @@ var RootCmd = &cobra.Command{
 		if err := util.ConfigureLogging(logLevel); err != nil {
 			return err
 		}
-		if !opts.NoPush && len(opts.Destinations) == 0 {
-			return errors.New("You must provide --destination, or use --no-push")
+		if !opts.NoPush && opts.TarPath == "" && len(opts.Destinations) == 0 {
+			return errors.New("You must provide --destination or --tarPath, or use --no-push")
 		}
 		if err := cacheFlagsValid(); err != nil {
 			return errors.Wrap(err, "cache flags invalid")


dlorenc:
> Currently user has to specify `--destination` argument even if one wants to generate tarball with resulting image:

Can you explain the use-case here? You're correct that the value "nothing" works, but we *do* need a value. The name of the image specified here is actually saved in the tarball, and used if you pass the tarball to "docker load".

We could default it to something to avoid making people pass a flag in here, but it is relevant depending on what you're going to do with the produced tar.

YorikSar:
Thanks for swift reply, @dlorenc!

> The name of the image specified here is actually saved in the tarball, and used if you pass the tarball to "docker load".

I'm sorry, I didn't dig deep enough to realize that. I guess it's just a documentation issue then: `--tarPath` claims to work *instead of pushing the image*, and `--destination` is where *image should be pushed to*, so they don't seem to work together at all. I'll try to change wording there to make it clearer for other newcomers.

sharifelgamal:
Since this is actually a documentation issue, I'm going to go ahead and close this PR.

PR #401, opened 2018-10-18
Support insecure pull

dtaniwaki:
I run a system which pushes a Docker image by user's request and creates another Docker images based on the built image. As I use insecure Docker registry in my local environment for this system, I also need to pull Docker images in an insecure manner.

Could you consider merging it? If the option `--insecure` should be separated for push and pull, please tell me, then I will make it like so.


avorima:
@dtaniwaki I would really appreciate if the `--insecure` flag could be separated for push and pull.

I'm using a registry that doesn't allow HTTP for pulling images and pushing the resulting images after a build to an insecure registry. With this generalized option I don't see a way for this to be done without some intermediate steps.

dtaniwaki:
@avorima That makes sense. How do the separated options look like? Maybe, `--insecure-pull` and `--insecure-push`?

avorima:
@dtaniwaki That would be the most obvious choice, yes. Maybe even keep the `--insecure` flag and make it set both pull and push to true, to not break behaviour.

dtaniwaki:
I'll make a pull request now.

PR #404, opened 2018-10-18
parse arg commands at the top of dockerfiles

### LINKED ISSUE #395: ARG defaults
**Actual behavior**
Kaniko doesn't understand how to handle ARG values declared before any stage. Kaniko output:
```
INFO[0000] Downloading base image golang:-alpine        
2018/10/09 20:10:09 No matching credentials were found, falling back on anonymous
error building image: getting stage builder for stage 0: unsupported status code 404; body: 404 page not found
```

**Expected behavior**
There are at least 2 use cases for this:

1. Declare ARGs to be used in a FROM directive. Kaniko should perform variable expansion in the FROM directive, based on ARGs at the beginning of the Dockerfile.
1. Pass ARGs from the Dockerfile prologue into a stage. Kaniko should pass ARG values from the beginning of the Dockerfile into Dockerfile stages if there's a valueless ARG declaration in the stage.

**To Reproduce**
1. Run Kaniko with a Dockerfile like the example in this issue.

**Additional Information**
 - Dockerfile
    ```Dockerfile
    ARG GO_VERSION=1.10
    ARG ALPINE_VERSION=3.8
    ARG PACKAGE=example.com/myapp
    ARG APP=myapp
    
    FROM golang:${GO_VERSION}-alpine${ALPINE_VERSION} as builder
    
    ARG PACKAGE
    RUN mkdir -p /go/src/${PACKAGE}
    WORKDIR /go/src/${PACKAGE}
    
    COPY . ./
    ARG APP
    RUN go build -o /${APP}
    
    FROM alpine:${ALPINE_VERSION} as target
    
    ARG APP
    COPY --from=builder /${APP} /
    CMD [ "/${APP}" ]
    ```
 - Build Context: See https://github.com/cjyar/skaffold-bug for a nearly identical bug I reported to skaffold.
 - Kaniko Image: gcr.io/kaniko-project/executor:debug sha256:6939de37b5b68c24f148b611ea38841534228625bde0c37d83c20a5b951f8b5e

sharifelgamal:
Fixes #395

### DIFF
--- a/pkg/snapshot/snapshot.go
+++ b/pkg/snapshot/snapshot.go
@@ -164,7 +164,6 @@ func (s *Snapshotter) TakeSnapshotFS() (string, error) {
 		}
 	}
 
-	addedPaths := []string{}
 	// Now create the tar.
 	for path := range memFs {
 		whitelisted, err := util.CheckWhitelist(path)
@@ -183,21 +182,9 @@ func (s *Snapshotter) TakeSnapshotFS() (string, error) {
 		}
 		if maybeAdd {
 			logrus.Debugf("Adding %s to layer, because it was changed.", path)
-			addedPaths = append(addedPaths, path)
 			if err := t.AddFileToTar(path); err != nil {
 				return "", err
 			}
-		}
-	}
-
-	if len(addedPaths) == 1 {
-		d, err := os.Stat(addedPaths[0])
-		if err != nil {
-			return "", nil
-		}
-		if d.IsDir() {
-			logrus.Infof("Only added one directory, not taking snapshot.")
-			return "", nil
 		}
 	}
 


pkg/dockerfile/buildargs.go
```diff
@@ -53,3 +54,15 @@ func (b *BuildArgs) ReplacementEnvs(envs []string) []string {
 	filtered := b.FilterAllowed(envs)
 	return append(envs, filtered...)
 }
+
+func (b *BuildArgs) AddMetaArgs(metaArgs map[string]string) *BuildArgs {
+	for key, value := range metaArgs {
+		logrus.Infof("META ARG: adding %s = %s", key, value)
+		b.AddMetaArg(key, &value)
```
priyawadhwa: I think the bug is here -- since there is a single `value` variable, the address it's pointing to is the same each time, so when that's passed in to AddMetaArg, all of the meta args end up pointing to the same value.

I think if you change it to 
```golang
for key, value := range metaArgs {
  v := value
  b.AddMetaArg(key, &v)
}

``` 
this should work?

sharifelgamal: That was it. Thank you!

sharifelgamal:
This should be ready for review.

Makefile
```diff
@@ -14,7 +14,7 @@
 
 # Bump these on release
 VERSION_MAJOR ?= 0
-VERSION_MINOR ?= 3
+VERSION_MINOR ?= 5
```
priyawadhwa: Thanks for catching this!! Must have missed it...many times... :sweat_smile:

pkg/executor/build.go
```diff
@@ -250,12 +253,21 @@ func (s *stageBuilder) saveSnapshot(createdBy string, ck string, contents []byte
 // DoBuild executes building the Dockerfile
 func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 	// Parse dockerfile and unpack base image to root
-	stages, err := dockerfile.Stages(opts)
+	stages, metaArgs, err := dockerfile.Stages(opts)
```
priyawadhwa: WDYT of storing the metaArgs for each stage in the KanikoStage type?

sharifelgamal: yeah, that simplified things considerably. done.

pkg/snapshot/snapshot.go
```diff
@@ -208,12 +209,18 @@ func (s *Snapshotter) snapShotFS(f io.Writer) (bool, error) {
...
+			addedPaths = append(addedPaths, path)
 			filesAdded = true
 			if err := t.AddFileToTar(path); err != nil {
 				return false, err
 			}
 		}
 	}
 
+	if filesAdded && len(addedPaths) == 1 && addedPaths[0] == "/" {
+		logrus.Infof("Only added /, not taking snapshot.")
```
dlorenc: Why is this change needed?

sharifelgamal: so for a command like `RUN touch /${FOO}`, where $FOO doesn't resolve to anything, there should just be an empty layer appended to the image, that's what docker does. For some reason, we see that / has changed and add that to the snapshot, resulting in an extra non-empty layer compared to docker.

dlorenc: Hrmm, what if it's like "/foo/${FOO}"? Maybe the check is if the only added path is a directory, rather than root?

sharifelgamal: yep, done.

PR #409, opened 2018-10-24
Separate Insecure Pull Options

dtaniwaki:
As discussed in https://github.com/GoogleContainerTools/kaniko/pull/401#issuecomment-432704443, it's possible to have either a source or destination registry is insecure. To get fine tuning of the security, we should separate the option.


dtaniwaki:
Do you have any plan to release a new version including this fix?

priyawadhwa:
Hey @dtaniwaki we just released v0.6.0!

dtaniwaki:
Thank you!

PR #423, opened 2018-10-31
Preserve options when doing a cache push

carlossg:
Otherwise options like `insecure` are lost

### DIFF
--- a/pkg/executor/push.go
+++ b/pkg/executor/push.go
@@ -126,7 +126,7 @@ func pushLayerToCache(opts *config.KanikoOptions, cacheKey string, layer v1.Laye
 	if err != nil {
 		return errors.Wrap(err, "appending layer onto empty image")
 	}
-	return DoPush(empty, &config.KanikoOptions{
-		Destinations: []string{cache},
-	})
+	cacheOpts := opts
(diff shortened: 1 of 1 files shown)


carlossg:
For the record this moves forward allowing insecure locally deployed docker registries but doesn't make it work yet, the builds still can't find the cached image even though it is in the repo

pkg/executor/push.go
```diff
@@ -126,7 +126,7 @@ func pushLayerToCache(opts *config.KanikoOptions, cacheKey string, layer v1.Laye
 	if err != nil {
 		return errors.Wrap(err, "appending layer onto empty image")
 	}
-	return DoPush(empty, &config.KanikoOptions{
-		Destinations: []string{cache},
-	})
+	cacheOpts := opts
+	cacheOpts.Destinations = []string{cache}
```
dlorenc: This mutates opts, could we do it by creating a new object?

carlossg: sorry, I think I've fixed it

priyawadhwa: I think this should work from this [test](https://play.golang.org/p/Nld2igwDRAx)

Issue #358, opened 2018-09-16
Ability to use a TempFile instead of an in-memory buffer for tarball snapshots to build large Docker containers?


nelsonjchen:
https://github.com/GoogleContainerTools/kaniko/blob/13accbaf3243a5cf477c9e3d5116dc09dd80ae37/pkg/snapshot/snapshot.go#L57-L70

I'm interested in building multi-GB Docker images with Kaniko. Unfortunately, doing this immediately exhausts the RAM. I think this is because the snapshotting steps take a tarball snapshot of the whole filesystem in-memory. 

It would be nice to have an option to tell kaniko to use a Tempfile for the snapshot backing.

priyawadhwa:
@nelsonjchen this should be fixed with #428. I'll go ahead and close this issue, please comment or open another if you run into any more problems!

nelsonjchen:
Woah! I didn't expect this to be resolved. "Have a bit of faith." - Some game I'm currently playing. 

I look forward to trying this.

lovetheguitar:
Hey,

sorry for the stupid question, but how do I enable this?
I'd like to build a 10+GB image from within Gitlab CI.
Thanks for any hints.

nelsonjchen:
@lovetheguitar I would look at the code. Always look at the code. It's authoritative. Check out the PR. Apparently, you don't have to do anything except provide disk space.

lovetheguitar:
> @lovetheguitar I would look at the code. Always look at the code. It's authoritative. Check out the PR that was merged and mentioned above. Apparently, you don't have to do anything except provide disk space.

Thanks for your answer. Yeah, I tried, but I don't know go at all. 
Either way it appears to be a different bug #909.

PR #428, opened 2018-11-02
Add buffering for large layers.

dlorenc:
This writes layer tarballs to disk rather than keeping them all in memory.

#358


pkg/snapshot/snapshot.go
```diff
@@ -55,32 +54,16 @@ func (s *Snapshotter) Key() (string, error) {
 
 // TakeSnapshot takes a snapshot of the specified files, avoiding directories in the whitelist, and creates
 // a tarball of the changed files. Return contents of the tarball, and whether or not any files were changed
-func (s *Snapshotter) TakeSnapshot(files []string) ([]byte, error) {
-	buf := bytes.NewBuffer([]byte{})
```
bobcatfish: it looks like you decided to move this buffering up a layer - personally i think it's nicer for the caller not to have to worry about it but either way, maybe the commit message could go into some detail about why that choice was made?

pkg/snapshot/snapshot_test.go
```diff
@@ -165,23 +168,6 @@ func TestSnapshotFiles(t *testing.T) {
 	testutil.CheckErrorAndDeepEqual(t, false, nil, expectedFiles, actualFiles)
 }
 
-func TestEmptySnapshotFS(t *testing.T) {
```
bobcatfish: y no longer a test case to test what happens when there is nothing to snapshot?

pkg/executor/build.go
```diff
@@ -160,30 +161,37 @@ func (s *stageBuilder) build() error {
 			return err
 		}
 		files = command.FilesToSnapshot()
-		var contents []byte
 
 		if !s.shouldTakeSnapshot(index, files) {
 			continue
 		}
 
+		f, err := ioutil.TempFile("/kaniko", "")
```
bobcatfish: is there any case where we would want this to be in-memory instead? (i.e. configurable)

pkg/snapshot/snapshot_test.go
dlorenc: The special casing was removed so now a tar is always created, it's just empty. I added the test back with that assertion.

pkg/snapshot/snapshot.go
```diff
@@ -29,6 +27,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var snapshotPathPrefix = "/kaniko"
```
priyawadhwa: nit: we could use [KanikoDir](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/constants/constants.go#L30) instead wherever this is used

Issue #410, opened 2018-10-25
Trouble getting caching to work


stepchowfun:
**Actual behavior**

Hello! We're running into a problem where Kaniko can push to our AWS ECR repo, but it reports an authorization error when attempting to use it as a cache. The following transcript demonstrates the issue:

```sh
$ mkdir context
$ echo "FROM redacted.amazonaws.com/jorb:trusty-v11" > context/Dockerfile
$ echo "{\"credHelpers\": { \"redacted.amazonaws.com\": \"ecr-login\"}}" > config.json

$ docker run \
>   -v "${PWD}/config.json":/kaniko/.docker/config.json:ro \
>   -v "${PWD}/context":/workspace:ro \
>   gcr.io/kaniko-project/executor:latest \
>   --context /workspace \
>   --dockerfile Dockerfile \
>   --destination redacted.amazonaws.com/jorb/cache:trusty-v11 \
>   --cache true \
>   --cache-repo redacted.amazonaws.com/jorb/cache
WARN[0000] Error while retrieving image from cache: unsupported status code 401; body: Not Authorized

INFO[0000] Downloading base image redacted.amazonaws.com/jorb:trusty-v11
INFO[0000] Extracting layer 0
INFO[0003] Extracting layer 1
INFO[0004] Extracting layer 2
INFO[0004] Extracting layer 3
INFO[0004] Extracting layer 4
INFO[0004] Extracting layer 5
INFO[0005] Extracting layer 6
INFO[0005] Extracting layer 7
INFO[0005] Extracting layer 8
INFO[0006] Extracting layer 9
INFO[0007] Extracting layer 10
INFO[0007] Extracting layer 11
INFO[0007] Extracting layer 12
INFO[0007] Taking snapshot of full filesystem...
2018/10/25 18:46:56 pushed blob sha256:5d7664d63832d5cb10465ec41d8f179020768a2df761ca70fb6773bf2cff0a63
2018/10/25 18:46:56 pushed blob sha256:8869e3292613da11e85476de60d76f79dd73b78f43a48c5044a44dc77476d95b
2018/10/25 18:46:56 pushed blob sha256:3bb895a3487dbb281df67c697c33233d183d74260eb3b95694b96c52532fb493
2018/10/25 18:46:56 pushed blob sha256:7ebc4f0b1cc63fd679d0b3c8c5d360411aff0c72b386a61de2fab0a621e7d149
[...]

PR #441, opened 2018-11-08
Fix caching for multi-step builds.

dlorenc:
only have a single command.

This change fixes that by properly "replaying" the Dockerfile and mutating the config when
calculating cache keys. Previously we were looking at the wrong cache key for each command
when there was more than one.

I think this should fix https://github.com/GoogleContainerTools/kaniko/issues/410


priyawadhwa:
Looks like the cache test failed in kokoro :disappointed:

xoen:
Hello @dlorenc - first of all, thanks so much for the quick fix ❤️ 

I've checked out this branch and built the executor docker image locally and tried again with our template `Dockerfile` (as commented on [Issue #410](https://github.com/GoogleContainerTools/kaniko/issues/410#issuecomment-436729409)).

I tried both with the original version and with the version tweaked to remove the `WORKDIR` and the good news is that in both cases it now seems to use the cache as I see a bunch of `Found cached layer, extracting to filesystem`, so that's great 🎊 .

I still see that it pushes more stuff to the `/cache` repo on subsequent builds despite nothing changing. These are very small images (AWS ECR says "< 0.01 MiB") but over time it can hit the default AWS ECR limit of 1,000 images per repository, especially if using one single cache repository per class of applications - unless of kaniko has some kind of built-in cache cleanup mechanism which I'm not aware of.

integration/dockerfiles/Dockerfile_test_cache_install
```diff
@@ -18,3 +18,5 @@
 
 FROM gcr.io/google-appengine/debian9@sha256:1d6a9a6d106bd795098f60f4abb7083626354fa6735e81743c7f8cfca11259f0
 RUN apt-get update && apt-get install -y make
+COPY context/bar /context
```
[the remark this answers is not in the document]
dlorenc: Thanks! Good idea.

dlorenc:
> Looks like the cache test failed in kokoro 😞

It looks like that was a bug with container-diff. I changed the Dockerfile to not triggger it.

integration-test.sh
```diff
@@ -36,4 +36,4 @@ echo "Running integration tests..."
 make out/executor
 make out/warmer
 pushd integration
-go test -v --bucket "${GCS_BUCKET}" --repo "${IMAGE_REPO}" --timeout 30m
+go test -v --bucket "${GCS_BUCKET}" --repo "${IMAGE_REPO}" --timeout 30m -run=TestCache
```
priyawadhwa: -run leftover from debugging!

dlorenc: Ah I always do that.

PR #448, opened 2018-11-13
Add benchmarking code

sharifelgamal:


integration/images.go
```diff
@@ -182,6 +182,11 @@ func (d *DockerFileBuilder) BuildImage(imageRepo, gcsBucket, dockerfilesPath, do
 		}
 	}
 
+	benchmarkFile := ""
```
dlorenc: What do you think of enabling this as an environment variable rather than a flag?

sharifelgamal: just for the integration tests? or in general?

dlorenc: In general.

PR #451, opened 2018-11-14
Make snapshotting faster by using filepath.SkipDir.

dlorenc:
filepath.Walk has a special error you can return from your walkFn
indicating it should skip directories. This change makes use of that
to skip whitelisted directories.


dlorenc:
In my non-scientific experiments, this saves about 2s (or 50%) per snapshot of debian (4s without, 2s with).

Issue #437, opened 2018-11-07
multi-stage copy of symlinks 


mtsyganov:
**Actual behavior**
In the multistage build process when a directory with symlinks is copied, the symlinks are badly created in the target image. Too many levels of symbolic links are generated.

**Expected behavior**
Symlinks are copied from source image into the target image as they are created in the source image

**To Reproduce**
1. create an image with symbolic link. For example `/usr/local/psql/bin/postmaster -> postgres` (you can check `registry.gitlab.com/gitlab-org/build/cng/postgresql:9.6.8`)
2. build a multistage image where directory with symlinks is copied to target image `COPY --from=pg_image /usr/local/psql/bin/ /usr/bin/`
3. run created image and check the copied symlinks. In the source image they are defined as `/usr/local/psql/bin/postmaster -> postgres`, but in the target image as `/usr/bin/postmaster -> /usr/bin/postmaster/postgres`. Correct symlink should be in the target image `/usr/bin/postmaster -> postgres`

**Additional Information**
 - Dockerfile 
```
FROM registry.gitlab.com/gitlab-org/build/cng/postgresql:9.6.8 as pg_image
FROM ubuntu
COPY --from=pg_image /usr/local/psql/bin/ /usr/bin/
```

mtsyganov:
@priyawadhwa retested with latest debug image. problem is still reproducable

andrewrynhard:
@priyawadhwa Where is this on the list of priorities? This is a major blocker for my particular use case. If low on your list I'd consider picking this up.

miguelitoq76:
Dear @priyawadhwa we had the same error during the use of kaniko in a custom image,
this a for us a problem, until we started to create a (Dummy build via executor --no_push), to force a filesystem cleanup. See my post in https://github.com/GoogleContainerTools/kaniko/issues/434

I assume, it would be a good idea, to start the build with a "delete FS"

priyawadhwa:
@miguelitoq76 sorry, I'm a bit confused -- do you want kaniko to delete the filesystem before starting the image build? Wouldn't that essentially be the same as directly using the kaniko image (which is built from scratch)?

tequilarista:
Triage - collapsing 'waiting for response' category into 'blocked'

cvgw:
I am unable to repro this issue; leading me to think that it has since been fixed. If this issue can still be repro'd please share the appropriate steps to do so.

Kami-no:
[...]

PR #459, opened 2018-11-18
Overwrite existing dest when copying symlink and preserve link target

### LINKED ISSUE #442: Copying symlink from previous stage fails instead of overwriting
**Actual behavior**
`COPY --from=0 X Y` where X is a symlink from stage 0 and Y is an pre-existing symlink in the current stage fails with "file exists".

**Expected behavior**
`COPY --from=0 X Y` where X is a symlink from stage 0 and Y is a pre-existing symlink in the current stage should overwrite Y with X as it would if neither were symlinks.

**To Reproduce**

```bash
docker run --rm -v $PWD:/workspace:ro gcr.io/kaniko-project/executor:v0.6.0 --no-push
```
with the following Dockerfile:
```dockerfile
FROM python:slim

FROM python:slim
COPY --from=0 /usr/local/lib/libpython3.7m.so.1.0 /usr/local/lib/libpython3.7m.so.1.0
```

Although this is a bit contrived, we had a Dockerfile which was building into /usr/local/lib and then copying that into the final stage which seems like a legitimate use-case.

aduong:
Fixes #442, #437

### DIFF
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -501,8 +501,12 @@ func CopySymlink(src, dest string) error {
 	if err != nil {
 		return err
 	}
-	linkDst := filepath.Join(dest, link)
-	return os.Symlink(linkDst, dest)
+	if FilepathExists(dest) {
+		if err := os.RemoveAll(dest); err != nil {
+			return err
+		}
+	}
+	return os.Symlink(link, dest)
 }
 
 // CopyFile copies the file at src to dest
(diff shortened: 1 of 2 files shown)


aduong:
I see that the integration test run was unsuccessful. It succeeds locally but I don't have access to the kokoro data. How should I proceed here?

priyawadhwa:
@aduong looks like a flaky test, I'll rerun kokoro.

aduong:
Looks like that was it. All checks pass now.

Issue #457, opened 2018-11-17
Files from /etc added in RUN do not get copied into snapshot


dzoeteman:
**Actual behavior**
Files added in RUN command to the `/etc` directory do not get added to the snapshot.

**Expected behavior**
Files in `/etc` directory added by RUN command should be added to snapshot.

**To Reproduce**
1. Use Dockerfile below for a build
2. Notice in build log that the two `ls` commands both show files in directory
3. Open shell in the built image
4. Notice that `/var/test/testfile` exists, but `/etc/test/testfile` doesn't.

**Additional Information**
 - Dockerfile 

```
FROM debian

RUN mkdir -p /var/test \
    && mkdir -p /etc/test \
    && touch /var/test/testfile \
    && touch /etc/test/testfile \
    && ls -lah /var/test \
    && ls -lah /etc/test;
```

 - Build Context

No additional files needed.

 - Kaniko Image (fully qualified with digest)

`gcr.io/kaniko-project/executor:latest` (`gcr.io/kaniko-project/executor@sha256:434bbb1d998ba1bd8ebc04c90d93afa859fd5c7ff93326bca9f6e7da0d6277ff`)

This is a pretty big issue. Any installs where config is added in `/etc` will not work properly. How I found this out for example, was that I tried to install `nginx`, but the `/etc/nginx` directory was missing from the image.
The issue is not in release `v0.6.0`, which leads me to believe as #451 might be the problem? That's a pure guess though.

slavoutich:
I am experiencing kinda the same issue, but with a different flavour. I have a base image, where I `ADD` a file to `/etc/init.d`, where it is added successfully, and I have an image constructed `FROM` base image, where it magically disappears. I think that this should be the same bug, actually, but can try to open separate bug with more detailed info upon request.
```
Pulling docker image gcr.io/kaniko-project/executor:debug ...
Using docker image sha256:6e43810f8da1b757c553ca5d8f70956f127a6eb70f1400fca537db32ef89e7ca for gcr.io/kaniko-project/executor:debug ...
```

dzoeteman:
Can confirm this is fixed with #461, so this can be closed.

Issue #453, opened 2018-11-15
COPY to root path doesn't work


gimler:
**Actual behavior**
no files in the image under `/etc/php7`

**Expected behavior**
files in the container under `/etc/php7`

**To Reproduce**
1. with docker it works
```
docker build -t copy .
docker run --rm -it copy ls -al /etc/php7/php.ini
> -rw-r--r--    1 root     root         74844 Nov 14 11:53 /etc/php7/php.ini
```
2. with kaniko not
```
/kaniko/executor -f `pwd`/${version}/Dockerfile -c `pwd`/${version}...
docker run --rm -it copy ls -al /etc/php7/php.ini
> ls: /etc/php7/php.ini: No such file or directory
```

**Additional Information**
 - it works with
```
COPY docker-entrypoint.d/* /docker-entrypoint.d/
```
 - Dockerfile 
```
FROM alpine:3.8
COPY rootfs/ /
```

 - Build Context
rootfs/etc/php7/php.ini

 - Kaniko Image (fully qualified with digest)
gcr.io/kaniko-project/executor:debug digest sha256:1b96ff24d5257e8ace559d4625c80a01f17671d8aa087ee175a6e2aabe02d042

Overbryd:
I have the same problem:

* Excerpt from the Dockerfile

```
FROM debian:stretch-slim

COPY ./docker/live/supervisor/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
```

* Kaniko build arguments:

```
/kaniko/executor \
  --cache=true \
  --dockerfile=/tmp/build/git-repo/docker/live/Dockerfile \
  --destination=gcr.io/<redacted>/kaniko-build \
  --context=/tmp/build/git-repo \
  --single-snapshot \
  --reproducible
```

* The file does not exist in the image:

```
$ docker run -ti kaniko-build /bin/bash
root@3c638df61f30:/data# ls -lah /etc/supervisor/conf.d/supervisord.conf
ls: cannot access '/etc/supervisor/conf.d/supervisord.conf': No such file or directory
root@3c638df61f30:/data# ls -lah /etc/supervisor/conf.d
ls: cannot access '/etc/supervisor/conf.d': No such file or directory
root@3c638df61f30:/data# ls -lah /etc/supervisor
ls: cannot access '/etc/supervisor': No such file or directory
```

Overbryd:
Ping @dlorenc @priyawadhwa how can we help debug this issue?

[...]

PR #461, opened 2018-11-19
Only return filepath.SkipDir for directories

priyawadhwa:
From the docs on filepath.SkipDir:

> If the function returns SkipDir when invoked on a non-directory file, Walk skips the remaining files in the containing directory

This was causing the bug in #457. Since the file `/etc/hosts` was in the whitelist, when filepath.SkipDir was called the entire etc directory was skipped.

This change only returns filepath.SkipDir on directories.


gimler:
When will be this change available in the images?

it fix eventually #453

priyawadhwa:
@gimler it's already available in the `:latest` and `:commit_sha` version of the image, and it will be part of the next release as well.

PR #468, opened 2018-11-25
Allow to create empty layer

### LINKED ISSUE #315: Missing records in image history
Dockerfile instructions that do not create new layers are not recorded in image history.

emmanuelCarre:
Hi,

This PR fix #315 by allow to create empty layer to keep in track command like `ENV` into image history.

I take a look about add an integration test; but they are wired to GCloud. Why choose to do it?

There are still issues in the history as the date of the layers but I thought it is better to submit a first PR.

Regards,

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -236,7 +237,7 @@ func (s *stageBuilder) build() error {
 				return pushLayerToCache(s.opts, ck, tarPath, command.String())
 			})
 		}
-		if err := s.saveSnapshotToImage(command.String(), tarPath); err != nil {
+		if err := s.saveSnapshotToImage(&s.cf.Config, command.RequiresUnpackedFS(), command.String(), tarPath); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


priyawadhwa:
Hi @emmanuelCarre, thanks for taking this issue on! From my understanding, metadata commands don't add empty layers to the image, they just add to the image history in the config file.

To update the history in the image config file, you'll first have to add a function to the [go-containerregistry](https://github.com/google/go-containerregistry) library, which we use to modify images. It would look really similar to this [CreatedAt](https://github.com/google/go-containerregistry/blob/c0dc13359f0157195b20e7b2af90ca7cad2db828/pkg/v1/mutate/mutate.go#L110) function, which modifies the image creation time, except the function would append a new history object to the config file  [history](https://github.com/GoogleContainerTools/kaniko/blob/master/vendor/github.com/google/go-containerregistry/pkg/v1/config.go#L31).

Then, kaniko can use that function to append an empty layer to the config history whenever it encounters an empty layer (that would happen [here](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/build.go#L220)). kaniko would append a history for a layer with [EmptyLayer](https://github.com/GoogleContainerTools/kaniko/blob/master/vendor/github.com/google/go-containerregistry/pkg/v1/config.go#L45) set to true.

I'm not sure if I explained that super well, so let me know if you have any questions!

emmanuelCarre:
Hi, 

Thank you very much for explanatiosn and details. I will work on it this week and update pull request this weekend.

tejal29:
closing due to inactivity.

Issue #466, opened 2018-11-21
Kaniko doesn't support exclusions in `.dockerignore`


nartamonov:
**Actual behavior**
Kaniko doesn't support exclusions in `.dockerignore`.

**Expected behavior**
Kaniko should respect exclusion patterns in `.dockerignore`.

**To Reproduce**
1. Create `.dockerignore` with the following content:

```
*
!target/scala-2.12/backend.jar
```

2. Kaniko (latest version) will delete all files in current context, but it should not delete `target/scala-2.12/backend.jar`.

etiennelb:
Hi folks.

I had the same issue right now.
I have a dockerignore file with the same content : 
```
*
!target/xxx/xxx.war
```

I ran kaniko with the following command, mounting my local workspace in the container : 
```
docker container run --rm \
           -v $HOME/.docker/config.json:/root/.docker/config.json \
           -v $PWD:/workspace \
           $KANIKO_IMAGE"
```

Kaniko deleted all my project files.

I lost everything. Even my local git repository... :disappointed:

priyawadhwa:
@etiennelb I'm SO sorry to hear that. I'll see if I can change the functionality so that dockerignore files are stored in memory and ignored during build time, instead of removing them at the beginning.

bobcatfish:
@priyawadhwa looks like maybe we can close this one and create a follow-up issue to implement `.dockerignore` ?

priyawadhwa:
Hey @nartamonov @etiennelb, I just merged #486  which should add dockerignore support back into kaniko. Please comment on this issue or open another if you run into an more issues, and again sorry for the trouble this bug has caused.

PR #470, opened 2018-11-26
Remove support for .dockerignore

priyawadhwa:
Remove support for .dockerignore file until I can change it to ignore
files in memory instead of deleting them

Refer to #466


sharifelgamal:
You can just ignore the dockerignore test instead of deleting it. Otherwise LGTM.

PR #474, opened 2018-12-01
Fix intermediate layer caching

### LINKED ISSUE #458: Bug: Caching for intermediate stages is broken
**Actual behavior**
There are no snapshots being made for individual 'layers' in intermediate stages, resulting in severely degraded performance for docker-images which use intermediate stages as a way to cache multiple components.

**Expected behavior**
When I specify `--cache=true`, it should push every intermediate layer to the cache registry.

**To Reproduce**
```Dockerfile
## ./Dockerfile

## First stage, this does not get sliced into cached layers
FROM debian:stretch AS stageone

RUN touch cacheable_operation
RUN touch stageone_file

## Second stage, this does not get sliced into cached layers
FROM debian:stretch AS stagetwo

RUN touch cacheable_operation
RUN touch stagetwo_file

## Third stage, this does not get sliced into cached layers
FROM debian:stretch AS stagethree

RUN touch cacheable_operation
RUN touch stagethree_file

## Final stage, this one does get sliced into cached layers
FROM debian:stretch

RUN mkdir files
COPY --from=stageone stageone_file files/
COPY --from=stagetwo stagetwo_file files/
COPY --from=stagethree stagethree_file files/
```
```sh
/kaniko/executor --cache=true --context . --dockerfile ./Dockerfile
```

**Additional Information**
This specifically happens [here](https://github.com/GoogleContainerTools/kaniko/blob/0c294138b891f0dc05f495d60b274889894e10fc/pkg/executor/build.go#L220). Kaniko iterates over the steps in the build process and `continue`s (and therefore skips the later cache upload step) if [`stage.shouldTakeSnapshot(index, ...)`](https://github.com/GoogleContainerTools/kaniko/blob/0c294138b891f0dc05f495d60b274889894e10fc/pkg/executor/build.go#L261) is false, which it always is for intermediate commands in a non-final stage.

andrewrynhard:
Fixes #458.

@bobcatfish 
@MMeent

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -226,7 +226,7 @@ func (s *stageBuilder) build() error {
 		}
 		files = command.FilesToSnapshot()
 
-		if !s.shouldTakeSnapshot(index, files) {
+		if !s.shouldTakeSnapshot(index, files) && !s.opts.Cache {
 			continue
 		}
 
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -226,7 +226,7 @@ func (s *stageBuilder) build() error {
 		}
 		files = command.FilesToSnapshot()
 
-		if !s.shouldTakeSnapshot(index, files) && !s.opts.Cache {
+		if !s.shouldTakeSnapshot(index, files) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


andrewrynhard:
I signed it!

andrewrynhard:
I'm not convinced the fix is this simple :D. Jumping into the code for the first time so there might be some concepts I'm missing.

bobcatfish:
Thanks for tackling this @andrewrynhard !

At first glance (like you said!) it's a bit hard to understand how this change would address the problem - could you maybe update the commit mesage to explain a bit more about it? (I'm adding some commit message standards in #476 :D)

This one might benefit from a comment as well - I think it would be hard to understand in the code that that specific line was affecting caching of intermediate layers.

dlorenc:
I refactored this a bit and added a unit test, hope you don't mind @andrewrynhard !

PR #486, opened 2018-12-10
Add support for .dockerignore file

### LINKED ISSUE #466: Kaniko doesn't support exclusions in `.dockerignore`

priyawadhwa:
This PR adds support for the dockerignore file. Previously when kaniko
had support for the dockerignore file, kaniko first went through the
build context and deleted files that were meant to be ignored. This
resulted in a really bad bug where files in user mounted volumes would
be deleted (my bad).

This time around, instead of modifying the build context at all, kaniko
will check if a file should be excluded when executing ADD/COPY
commands. If a file should be excluded (based on the .dockerignore) it
won't be copied over from the buildcontext and shouldn't end up in the
final image.

I also added a .dockerignore file and Dockerfile as an integration test,
which should fail if the dockerignore is not being processed correctly or if files aren't being excluded correctly.
Also, I removed all the integration testing from the previous version of the
dockerignore support.
Should fix #466


pkg/util/fs_util.go
```diff
@@ -504,37 +511,78 @@ func CopyDir(src, dest string) ([]string, error) {
...
 	gid := fi.Sys().(*syscall.Stat_t).Gid
-	return CreateFile(dest, srcFile, fi.Mode(), uid, gid)
+	return false, CreateFile(dest, srcFile, fi.Mode(), uid, gid)
+}
+
+// excludeFile returns true if the .dockerignore specified this file should be ignored
+func excludeFile(path, buildcontext string) bool {
+	excluded, err := parseDockerignore(buildcontext)
+	if err != nil {
+		return false
```
sharifelgamal: Maybe log a warning here so it's less confusing for debugging?

sharifelgamal:
One comment, otherwise LGTM.

PR #487, opened 2018-12-11
Add more benchmarks

dlorenc:


pkg/executor/build.go
```diff
@@ -409,6 +409,8 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 }
 
 func fetchExtraStages(stages []config.KanikoStage, opts *config.KanikoOptions) error {
+	t := timing.Start("Fetching Extra Stages")
+	timing.DefaultRun.Stop(t)
```
sharifelgamal: This should be deferred.

dlorenc: doh

PR #488, opened 2018-12-12
Adding TTL to caching

### LINKED ISSUE #455: Add support for cache TTLs
Add a flag to only use cached layers that were built within the specified TTL.

sharifelgamal:
Fixes #455


integration/images.go
```diff
@@ -247,11 +247,17 @@ func (d *DockerFileBuilder) buildCachedImages(imageRepo, cacheRepo, dockerfilesP
 	cacheFlag := "--cache=true"
 
 	for dockerfile := range d.TestCacheDockerfiles {
+		benchmarkEnv := "BENCHMARK_FILE=false"
+		if b, err := strconv.ParseBool(os.Getenv("BENCHMARK")); err != nil && b {
```
priyawadhwa: I'm not sure if the condition here `err != nil && b` could actually happen, from the code [here](https://golang.org/src/strconv/atob.go?s=351:391#L1), unless I'm missing something?

sharifelgamal: oops, it should be `err == nil && b`, nice catch

integration/integration_test.go
```diff
@@ -284,6 +294,15 @@ func TestCache(t *testing.T) {
 			checkContainerDiffOutput(t, diff, expected)
 		})
 	}
+	if b, err := strconv.ParseBool(os.Getenv("BENCHMARK")); err == nil && b {
```
priyawadhwa: I haven't looked super closely, but would it be possible to move this entire block into a separate function and pass in `benchmark_cache` and `benchmark_layers` as args?

PR #491, opened 2018-12-13
Adds COPY --from=previous stage name/number validation

MMeent:
This fixes an issue in which COPY --from=previous-stage-name would try to download docker image previous-stage-name instead of checking that previous-stage-name could be a named stage.


dlorenc:
Hey! There was a flake with our integration tests. If you rebase it should pick up the fix and pass.

dlorenc:
I think I fixed the travis issue. Once that passes I can merge.

PR #500, opened 2018-12-28
Adding capability to get Dockerfile from URL.

linuxshokunin:
Added to specify an URL to --dockerfile parameter.

In our CI pipelines, we just specify dockerfile in github to build dockerfile. 
Before, I'm using initContainer to git clone first into the volume and then kaniko to build docker images.
In this way, kaniko directly gets dockerfile from the URL.

### DIFF
--- a/pkg/dockerfile/dockerfile.go
+++ b/pkg/dockerfile/dockerfile.go
@@ -48,7 +48,7 @@ func Stages(opts *config.KanikoOptions) ([]config.KanikoStage, error) {
 	}
 
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, fmt.Sprintf("reading dockerfile at path %s", opts.DockerfilePath))
 	}
 
 	stages, metaArgs, err := Parse(d)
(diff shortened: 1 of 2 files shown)


creation-shin-chan:
Does anyone know how to fix the error in the Travis CI build? 
I don't know why it's saying ineffectual assignment.

pkg/dockerfile/dockerfile.go
```diff
@@ -32,10 +34,23 @@ import (
 
 // Stages parses a Dockerfile and returns an array of KanikoStage
 func Stages(opts *config.KanikoOptions) ([]config.KanikoStage, error) {
-	d, err := ioutil.ReadFile(opts.DockerfilePath)
+	var err error
+	var d []uint8
+	match, _ := regexp.MatchString("^https?://", opts.DockerfilePath)
+	if match {
+		response, err := http.Get(opts.DockerfilePath)
```
dlorenc: I think this is the issue - you have err defined on line 37, then this redefines it in the new scope. You set err again below on 45 and 47, but don't check until 50. The checked err is outside this scope.

One fix would be to check the errors on 45 and 47 right away, or you could define response separately:

```golang
var response *http.Response
response, err = http.Get(opts.Dockerfile)
```

dlorenc:
> Does anyone know how to fix the error in the Travis CI build?
> I don't know why it's saying ineffectual assignment.

Left a comment on why I think you're getting this.

PR #513, opened 2019-01-09
Fill in missing docs for --insecure

dcecile:
Although the name of the option doesn't state it, I'm pretty sure this is for **push** only (based on the [help text](https://github.com/GoogleContainerTools/kaniko/blob/master/cmd/executor/cmd/root.go#L101) and the [usage](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/executor/push.go#L75)).

Simple change to fill in the missing link from the TOC:
```md
  - [Additional Flags](#additional-flags)
    - [--insecure](#--insecure)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,6 +359,10 @@ _This flag must be used in conjunction with the `--cache=true` flag._
 
 Set this flag to clean the filesystem at the end of the build.
 
+#### --insecure
+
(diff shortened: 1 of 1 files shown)


dcecile:
I signed it!

Issue #521, opened 2019-01-10
Stages with WORKDIR/only mkdir break caching


dinvlad:
**Actual behavior**
In a multi-stage build, if there's a step that only creates a folder (using `mkdir` or `WORKDIR`), then the downstream stages won't be cached.

**Expected behavior**
All stages cached if Dockerfile and context are unchanged.

**To Reproduce**
1. Create a Dockerfile with mkdir (example below)
2. Build once  with `--cache=true`
3. Build again with `--cache=true`
4. Observe the logs for both builds.
    Example from the 1st build:
    ```
    INFO[0003] RUN mkdir /cached_folder 
    INFO[0003] Found cached layer, extracting to filesystem 
    INFO[0004] Taking snapshot of files... 
    INFO[0004] Storing source image from stage 1 at path /kaniko/stages/1 
    INFO[0004] Deleting filesystem... 
    INFO[0004] Base image from previous stage 1 found, using saved tar at path /kaniko/stages/1 
    INFO[0004] Checking for cached layer gcr.io/project/kaniko-stage-caching-bug/cache:0116a345bc87ada9dc8811779ae7fe241a4c9e4fe45b55576dadd56adf2fa807
    INFO[0004] No cached layer found for cmd RUN echo bug_stage 
    ```
    Example from the 2nd build:
    ```
    INFO[0002] RUN mkdir /cached_folder 
    INFO[0002] Found cached layer, extracting to filesystem 
    INFO[0002] Taking snapshot of files... 
    INFO[0002] Storing source image from stage 1 at path /kaniko/stages/1 
    INFO[0003] Deleting filesystem... 
    INFO[0003] Base image from previous stage 1 found, using saved tar at path /kaniko/stages/1 
    INFO[0003] Checking for cached layer gcr.io/project/kaniko-stage-caching-bug/cache:7cdf15b929af18bcb142fdb9647f6d3b07a3d2cad3a056950792b0afbcb2b562
    INFO[0003] No cached layer found for cmd RUN echo bug_stage
    ```
5. Observe that hashes for the `RUN mkdir /cached_folder` layer *differ* in both cases, hence breaking caching of subsequent layers.
6. The same issue occurs if one uses `WORKDIR /cached_folder` instruction instead of `RUN mkdir /cached_folder`.

**Additional Information**
 - Dockerfile
    ```Dockerfile
    FROM alpine as base_stage

    RUN echo base_stage


    FROM base_stage as cached_stage

[...]

PR #525, opened 2019-01-11
Change cache key calculation to be more reproducible.

dlorenc:
Before we were using the full image digest, but that contains a timestamp. Now
we only use the layers themselves and the image config (env vars, etc.).

Also fix a bug in unpacking the layers themselves. mtimes can change during unpacking,
so set them all once at the end.

I think this will fix https://github.com/GoogleContainerTools/kaniko/issues/521

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_cache
+++ b/integration/dockerfiles/Dockerfile_test_cache
@@ -16,7 +16,15 @@
 # If the image is built twice, /date should be the same in both images
 # if the cache is implemented correctly
 
-FROM gcr.io/google-appengine/debian9@sha256:1d6a9a6d106bd795098f60f4abb7083626354fa6735e81743c7f8cfca11259f0
+FROM alpine as base_stage
+
+RUN mkdir foo && echo base_stage > foo/base
+
+FROM base_stage as cached_stage
+
+RUN echo cached_stage > foo/cache
+
+FROM cached_stage as bug_stage
+
+RUN echo bug_stage > foo/bug
 RUN date > /date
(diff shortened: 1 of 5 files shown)


dinvlad:
Thanks! I've tested this image with a real-world example (which I previously reduced to the example from #521), and for some reason caching still doesn't work. I've shared the full example at https://github.com/broadinstitute/kaniko-stage-caching-bug/blob/master/Dockerfile.full

Could you take a look? I'm also still trying to understand how to enable more debugging.

integration/dockerfiles/Dockerfile_test_cache
```diff
@@ -16,7 +16,14 @@
 # If the image is built twice, /date should be the same in both images
 # if the cache is implemented correctly
 
-FROM gcr.io/google-appengine/debian9@sha256:1d6a9a6d106bd795098f60f4abb7083626354fa6735e81743c7f8cfca11259f0
-RUN date > /date
-COPY context/foo /foo
-RUN echo hey
+FROM alpine as base_stage
```
priyawadhwa: Would this Dockerfile fail integration tests if caching didn't work? As in, if kaniko built it a second time without the cache during the tests, I'm not sure if container-diff would catch the difference (that's why I had `RUN date > /date` in the old version, since container-diff would fail if the `date` layer wasn't pulled from the cache).

dlorenc: Great point! I changed this to reproduce the caching issue I was hitting, but dropped the date thing. I'll double check.

PR #526, opened 2019-01-11
Make benchmark runs push to a separate namespace.

dlorenc:


benchmark.sh
```diff
@@ -17,4 +17,5 @@ set -ex
 script_name=$0
 script_full_path=$(dirname "$0")
 export BENCHMARK=true
+export IMAGE_REPO="gcr.io/kaniko-test-project/benchmarks"
```
priyawadhwa: I think we use the test project `gcr.io/kaniko-test`, unless you want this to be a different one!

dlorenc: whoops, fix coming...

PR #551, opened 2019-02-04
Generate the config.json from gcloud rather than relying on a hardcod…

dlorenc:
…ed one.

marketplace.gcr.io now requires auth, but our credential helper doesn't know that.


deploy/Dockerfile
```diff
@@ -26,12 +26,15 @@ RUN make -C /go/src/github.com/awslabs/amazon-ecr-credential-helper linux-amd64
 COPY . .
 RUN make
 
+FROM gcr.io/cloud-builders/gcloud
```
ImJasonH: Does this make the kaniko executor image `FROM gcr.io/cloud-builders/gcloud`? That image is really stupidly huge, so we should avoid that. We could make it `FROM gcr.io/cloud-builders/gcloud-slim` (which is half as large, and _still_ stupidly huge), but I'd prefer to remove a dependency on gcloud in general if we can.

dlorenc: No, it just uses it as an intermediate step to generate the config. The image size should be the same.

ImJasonH: Ohh I see, I didn't catch that the next section was a separate stage.

So the real benefit here is that Kaniko doesn't have to know and keep up-to-date the full set of gcr subdomains, and instead will just outsource that to gcloud which should be up-to-date. :+1:

PR #553, opened 2019-02-06
let's return errors when warming a cache

rawlingsj:


rawlingsj:
I'll take a look at the test failure tomorrow - I just had a quick glance and it wasn't immediately obvious why returning an error failed the test but will have a proper look.

dlorenc:
Ah looks like it was a flake from order being undefined when iterating over a map. A rerun should make the test pass, I'll fix the flake.

dlorenc:
There was a little lint error. I hope you don't mind that I fixed it and pushed the extra commit.

rawlingsj:
Just been catching up on my list of things to finish and saw you sorted this - not sure if you see github notifications as I miss them all the time but thank you @dlorenc!

PR #565, opened 2019-02-16
Add missing tests for bucket util

shuheiktgw:
I just added some missing tests for bucket_util.go! Thanks for your review! 👍

### DIFF
--- a/pkg/util/bucket_util.go
+++ b/pkg/util/bucket_util.go
@@ -24,7 +24,7 @@ import (
 
 func GetBucketAndItem(context string) (string, string) {
 	split := strings.SplitN(context, "/", 2)
-	if len(split) == 2 {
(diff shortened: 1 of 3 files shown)


pkg/util/bucket_util.go
```diff
@@ -24,7 +24,7 @@ import (
 
 func GetBucketAndItem(context string) (string, string) {
 	split := strings.SplitN(context, "/", 2)
-	if len(split) == 2 {
+	if len(split) == 2 && split[1] != "" {
```
shuheiktgw: For example, if `context` is `test/`, itemName will be `""` and I think that is not how it is supposed to work, so I fixed it!

dlorenc: Thanks!

PR #580, opened 2019-02-22
Environment variables should be replaced in URLs in ADD commands.

dlorenc:
We were previously explicitly skipping this for some reason, but Docker
seems to expand these in URLs so we should too.

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_add
+++ b/integration/dockerfiles/Dockerfile_test_add
@@ -24,4 +24,7 @@ COPY $file /arg
 
 # Finally, test adding a remote URL, concurrently with a normal file
 ADD https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v1.4.3/docker-credential-gcr_linux_386-1.4.3.tar.gz context/foo /test/all/
-ADD https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v1.4.3-static/docker-credential-gcr_linux_amd64-1.4.3.tar.gz /destination
+
+# Test environment replacement in the URL
+ENV VERSION=v1.4.3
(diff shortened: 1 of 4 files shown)


dlorenc:
There's one more separate fix here (this should trigger it in integration test runs) that I'll add after they fail.

integration/dockerfiles/Dockerfile_test_add
```diff
@@ -24,4 +24,7 @@ COPY $file /arg
 
 # Finally, test adding a remote URL, concurrently with a normal file
 ADD https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v1.4.3/docker-credential-gcr_linux_386-1.4.3.tar.gz context/foo /test/all/
-ADD https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v1.4.3-static/docker-credential-gcr_linux_amd64-1.4.3.tar.gz /destination
+
+# Test environment replacement in the URL
+ENV VERSION=v1.4.3
+ADD https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/${VERSION}-static/docker-credential-gcr_linux_amd64-1.4.3.tar.gz /destination
```
ImJasonH: Should this also do something to ensure the file was fetched and named correctly?

`RUN ls /destination/docker-credential-gcr_linux_amd64-1.4.3.tar.gz`

Or even better, `tar xcvf` it to make sure you actually got a `.tar.gz` and not a 404 response.

dlorenc: I could add that here, but we actually get that for free. The way these tests work is that we build each of these with Docker and Kaniko, then compare the two images.

PR #586, opened 2019-02-25
README.me: update Buildah description

vrothberg:
Use the latest description of Buildah from upstream.

Signed-off-by: Valentin Rothberg <rothberg@redhat.com>


vrothberg:
We received some additional feedback after the latest update that it wasn't clear if Buildah can work with Dockerfiles or not. The new description aims at addressing the feedback. Please excuse the noise.

dlorenc:
Thanks!

PR #599, opened 2019-03-05
Update go-containerregistry

### LINKED ISSUE #535: Building a stage using image with both tag and SHA digest fails
**Actual behavior**
With a Dockerfile containing a stage like this:
```
FROM ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68
```

Kaniko fails with the following message:
```
Downloading base image ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68 
error building image: getting stage builder for stage 0: could not parse reference
```

**Expected behavior**
Since Docker itself can build the image from the Dockerfile, I expect kaniko to be able to do the same.

**To Reproduce**
1. `echo "FROM ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68" >Dockerfile`
2. `docker run -v "$(pwd):/workspace" gcr.io/kaniko-project/executor:debug@sha256:901f53b350d0681d1fcde2654bdab048610428e1150beda51acb76fedd62ad85 --dockerfile Dockerfile --destination test`

priyawadhwa:
Update go-containerregistry since it can now handle image names of the
format repo:tag@digest.

Should fix #535.

Thanks @ViceIce for the fix!


ViceIce:
@priyawadhwa Is the failing check a problem?

priyawadhwa:
Hmm it seems like go-containerregistry is complaining about pushing an image with a bunch of the same layer ([Dockerfile__test_copy_same_file_many_times](https://github.com/GoogleContainerTools/kaniko/blob/master/integration/dockerfiles/Dockerfile_test_copy_same_file_many_times))

```
failed to push to destination gcr.io/kaniko-test/kaniko-dockerfile_test_copy_same_file_many_times:latest: 
UNKNOWN: Unable to write blob
 sha256:2a72f5f901f1612717dc31200476234795bdef01972d9de90edc0e7eab561402
```

@dlorenc have you seen this before?

jonjohnsonjr:
That's a real error from GCR, probably this: https://cloud.google.com/storage/docs/key-terms#immutability

cc @ImJasonH maybe the streaming layer stuff introduced this? We could handle errors better with retries but ideally we wouldn't be trying to PUT the same blobs. I guess in real-life images this would never happen.

@priyawadhwa what is this testing? Looking at some of the output artifacts, it's really bizarre to me that not all the layers end up being the same.

ImJasonH:
AFAIK Kaniko doesn't use `stream.Layer` at all (yet?).

Maybe there's some problem with uploading all the layers in concurrent goroutines, and when there are >10 identical blobs uploading at once they're more likely to conflict during a write? Just a guess, I don't actually know how GCR works.

jonjohnsonjr:
`remote.Write` used to use `BlobSet` to dedupe the layers, which was removed in this PR: https://github.com/google/go-containerregistry/pull/301/files

We could something similar to this but with `BlobSet`:
https://github.com/google/go-containerregistry/blob/4aac97bd085d96caedbc8213ddb6f489b7eb6706/pkg/v1/remote/write.go#L68

jonjohnsonjr:
From GCS docs:

> a single particular object can only be updated or overwritten up to once per second. For example, if you have an object bar in bucket foo, then you should only upload a new copy of foo/bar about once per second. Updating the same object faster than once per second may result in 429 Too Many Requests errors.

jonjohnsonjr:
Try again!

dlorenc:
Whoohoo!

priyawadhwa:
Thanks @jonjohnsonjr & @ImJasonH! 

This Dockerfile is meant to be testing snapshotting (ref #289)

PR #602, opened 2019-03-06
Write data about pushed images if env var is set

ImJasonH:
If the env var `BUILDER_OUTPUT` is set, it points to a path specified by GCB to write image output data to. GCB will use this to report the images pushed by Kaniko to end-users, and to report to Container Analysis that the images were built by that specific GCB build ID.


tejal29:
@ImJasonH Can you please rebase?

PR #618, opened 2019-03-14
Fix file mode bug

dtaniwaki:
I had an issue that I cannot `sudo` on a Docker image built with kaniko. I got the following error.

```
sudo: /usr/bin/sudo must be owned by uid 0 and have the setuid bit set
```

After a long investigation, I found it happened when I built an image over kaniko cache and the `setuid` bit can be removed by `chown`. You can check this behavior by the following steps.

```
$ touch foo
$ ls -l foo
-rw-rw-r-- 1 dtaniwaki dtaniwaki 0 Mar 14 22:56 foo
$ chmod u+s foo
$ ls -l foo
-rwSrw-r-- 1 dtaniwaki dtaniwaki 0 Mar 14 22:56 foo
$ chown dtaniwaki foo
$ ls -l foo
-rw-rw-r-- 1 dtaniwaki dtaniwaki 0 Mar 14 22:56 foo
```

It seems the syscall behavior, but at least the fix of this PR works.

You can see this behavior in kaniko by a Dockerfile like below.

```
FROM ubuntu:16.04

RUN mkdir aaa && chmod +t aaa
RUN touch bbb && chmod u+s bbb
# Change the echo string below to make a layer over the cache above.
RUN echo 1
RUN ls -la
```

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_cache_perm
+++ b/integration/dockerfiles/Dockerfile_test_cache_perm
@@ -1,0 +1,8 @@
+# Test to make sure the cache works with special file permissions properly.
+# If the image is built twice, directory foo should have the sticky bit,
+# and file bar should have the setuid and setgid bits.
+
+FROM busybox
+
+RUN mkdir foo && chmod +t foo
+RUN touch bar && chmod u+s,g+s bar
--- a/integration/images.go
+++ b/integration/images.go
@@ -134,6 +134,7 @@ func NewDockerFileBuilder(dockerfiles []string) *DockerFileBuilder {
 	d.TestCacheDockerfiles = map[string]struct{}{
 		"Dockerfile_test_cache":         {},
 		"Dockerfile_test_cache_install": {},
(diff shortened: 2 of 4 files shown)


dtaniwaki:
I updated the code to fix file mode for sticky bit on directories too.

dlorenc:
Would you mind adding a Dockerfile that reproduces this to the integration/dockerfiles directory as well? That way we can check that it keeps working in a real-world example.

dtaniwaki:
@dlorenc I added the Dockerfile 👍

dtaniwaki:
Oops, the Dockerfile I added has one odd line... Let me remove it and update the PR.

dtaniwaki:
I now realized `chown` doesn’t reset sticky bit while it resets `setuid` and `setgid` bits although I thought it did when I tested a few days ago, but it might’ve been my mistake. At any rate, I think we can keep this change as is because using the calling order of `chown` and `chmod` is easier to read.

PR #619, opened 2019-03-16
Fix parent directory permissions

dtaniwaki:
I found `chown` a directory after adding a file in it causes wrong directory owner and permission.

My `Dockerfile` was something like below.

```
FROM ubuntu:16.04

RUN adduser --disabled-password --gecos "" --uid 1000 user
RUN mkdir -p /home/user/foo
RUN chown -R user /home/user
ADD bar /home/user/foo/bar
RUN chown -R user /home/user
```

I tried the Docker image built in Kaniko and checked the owner and permissions by `docker run -it IMAGE ls -l /home/user`, then I got the following result.

```
total 4
drwxr-xr-x    1 root     root          4096 Mar 16 16:34 foo
```

I expect it to be:

```
total 4
drwx------    1 user     user          4096 Mar 16 16:34 foo
```

I compared the built image with the one built by `docker build`, then I found the image layers of `docker build` has parent directories if layers have newly added files.

I think some Docker volume driver cannot update a file system if intermediate layers lost info of parent directories of a file.

Here's a similar issue in Docker (originally an issue of aufs).
https://github.com/moby/moby/issues/1295

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_parent_dir_perms
+++ b/integration/dockerfiles/Dockerfile_test_parent_dir_perms
@@ -4,5 +4,5 @@ RUN adduser --disabled-password --gecos "" --uid 1000 user
 RUN mkdir -p /home/user/foo
 RUN chown -R user /home/user
 RUN chmod 700 /home/user/foo
-ADD Makefile /home/user/foo/Makefile
+ADD https://raw.githubusercontent.com/GoogleContainerTools/kaniko/master/README.md /home/user/foo/README.md
 RUN chown -R user /home/user
--- a/integration/dockerfiles/Dockerfile_test_parent_dir_perms
+++ b/integration/dockerfiles/Dockerfile_test_parent_dir_perms
@@ -1,0 +1,8 @@
+FROM busybox
+
+RUN adduser --disabled-password --gecos "" --uid 1000 user
+RUN mkdir -p /home/user/foo
+RUN chown -R user /home/user
+RUN chmod 700 /home/user/foo
+ADD Makefile /home/user/foo/Makefile
+RUN chown -R user /home/user
--- a/pkg/snapshot/snapshot_test.go
+++ b/pkg/snapshot/snapshot_test.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-	"strings"
 	"testing"
 
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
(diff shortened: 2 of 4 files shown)


pkg/snapshot/snapshot_test.go
```diff
@@ -147,7 +158,10 @@ func TestSnapshotFiles(t *testing.T) {
 	}
 	defer os.Remove(tarPath)
 
-	expectedFiles := []string{"/", "/tmp", filepath.Join(testDir, "foo")}
```
dtaniwaki: I actually don't know why `/` and `/tmp` was included...

pkg/snapshot/snapshot_test.go
```diff
@@ -230,3 +246,17 @@ func setUpTestDir() (string, *Snapshotter, func(), error) {
 
 	return testDir, snapshotter, cleanup, nil
 }
+
+func getParentDirectories(file string) []string {
```
dlorenc: I think we have a helper for this already in `util.ParentDirectories`

dtaniwaki:
@dlorenc I updated the PR!

integration/dockerfiles/Dockerfile_test_parent_dir_perms
```diff
@@ -0,0 +1,8 @@
+FROM busybox
+
+RUN adduser --disabled-password --gecos "" --uid 1000 user
+RUN mkdir -p /home/user/foo
+RUN chown -R user /home/user
+RUN chmod 700 /home/user/foo
+ADD Makefile /home/user/foo/Makefile
```
dlorenc: Could you remove this line? We don't have a Makefile in the context which is causing the integration test build to fail.

dtaniwaki: This line cannot be deleted because having ADD between chown cause this issue. Is there any file available in the environment?

dtaniwaki: Ah, maybe we can use some URL to download a file because it’s ADD. let me update the PR.

dtaniwaki:
@dlorenc I updated the Dockerfile, please check it.

dtaniwaki:
Thank you for merging the PR! Do you have any plan to release new version?

dlorenc:
Id like to get https://github.com/GoogleContainerTools/kaniko/pull/621 in then cut a release

Issue #615, opened 2019-03-13
Multistage builds not copying files


nfvs:
**Actual behavior**
Files that are generated in a stage do not get copied over in the next stages.

**Expected behavior**
Files should get copied over to the next stages

**To Reproduce**
1. Build a simple two-stage image
```
FROM python:3.5 AS builder
RUN pip install virtualenv
RUN virtualenv /venv -p python3

FROM python:3.5-slim
COPY --from=builder /venv /venv
```
2. Inspecting the resulting image, we notice there is no `/venv`. Also this output:
```
INFO[0117] Taking snapshot of full filesystem...        
INFO[0118] COPY --from=builder /venv /venv              
INFO[0119] Taking snapshot of full filesystem...        
INFO[0119] No files were changed, appending empty layer to config. No layer added to image. 
```

**Additional Information**
 - Runing this in GitLab using `gcr.io/kaniko-project/executor:debug` (tested `sha256:642ef6209052ed2d96bfc20d2edc2803300ce3d858773629683b6d0903624777` and `sha256:3c820e1112b24421d103582dbf27e7fca05b4eb335adaadca3d616d3d48689b9`)
 - Works fine with Docker:
```shell
❯ cat Dockerfile2
FROM python:3.5 AS builder
RUN pip install virtualenv
RUN virtualenv /venv -p python3

FROM python:3.5-slim
COPY --from=builder /venv /venv

❯ docker build . -t test -f Dockerfile2
Sending build context to Docker daemon  547.8kB
Step 1/5 : FROM python:3.5 AS builder
...
Successfully built 05bc82648689
Successfully tagged test:latest

❯ docker run -it test:latest ls -als /venv
total 20
4 drwxr-xr-x 5 root root 4096 Mar 13 16:02 .
4 drwxr-xr-x 1 root root 4096 Mar 13 16:03 ..
4 drwxr-xr-x 2 root root 4096 Mar 13 15:29 bin
4 drwxr-xr-x 2 root root 4096 Mar 13 15:29 include
4 drwxr-xr-x 3 root root 4096 Mar 13 15:29 lib
```

nfvs:
Seems like it's an issue with `.dockerignore`:
```
DEBU[0060] /kaniko/0/venv found in .dockerignore, ignoring
```

[...]

PR #621, opened 2019-03-18
Fix arg handling for multi-stage images in COPY instructions.

dlorenc:
I think this should fix https://github.com/GoogleContainerTools/kaniko/issues/615


dtaniwaki:
@dlorenc Could you consider making a new release for the recent bug fixes?

PR #622, opened 2019-03-18
Check push permissions before building images

ImJasonH:
This calls `remote.CheckPushPermission` for each unique repository in `--destination` flags, and fails if the current keychain doesn't have push permissions to any repository.

Tested locally:

```
$ ./run_in_docker.sh ./integration/dockerfiles/Dockerfile_test_target $PWD gcr.io/mattmoor-public/foo
error checking push permissions: checking push permission: DENIED: Token exchange failed for project 'mattmoor-public'. Caller does not have permission 'storage.buckets.get'. To configure permissions, follow instructions at: https://cloud.google.com/container-registry/docs/access-control
```

```
$ ./run_in_docker.sh ./integration/dockerfiles/Dockerfile_test_target $PWD gcr.io/jasonhall-kube/foo
INFO[0000] Resolved base name gcr.io/distroless/base:latest to gcr.io/distroless/base:latest 
INFO[0000] Resolved base name scratch to scratch        
INFO[0000] Resolved base name base to base              
...
```

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -79,7 +79,7 @@ var RootCmd = &cobra.Command{
 			logrus.Warn("kaniko is being run outside of a container. This can have dangerous effects on your system")
 		}
 		if err := executor.CheckPushPermissions(opts); err != nil {
-			exit(errors.Wrap(err, "error checking push permissions"))
+			exit(errors.Wrap(err, "error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again"))
 		}
 		if err := os.Chdir("/"); err != nil {
 			exit(errors.Wrap(err, "error changing to root dir"))
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -24,12 +24,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/GoogleContainerTools/kaniko/pkg/timing"
-
 	"github.com/GoogleContainerTools/kaniko/pkg/buildcontext"
 	"github.com/GoogleContainerTools/kaniko/pkg/config"
 	"github.com/GoogleContainerTools/kaniko/pkg/constants"
 	"github.com/GoogleContainerTools/kaniko/pkg/executor"
+	"github.com/GoogleContainerTools/kaniko/pkg/timing"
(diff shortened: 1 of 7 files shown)


cmd/executor/cmd/root.go
```diff
@@ -79,6 +78,9 @@ var RootCmd = &cobra.Command{
 			}
 			logrus.Warn("kaniko is being run outside of a container. This can have dangerous effects on your system")
 		}
+		if err := executor.CheckPushPermissions(opts); err != nil {
+			exit(errors.Wrap(err, "error checking push permissions"))
```
dlorenc: The main use case here is to fail early because of a typo right? Maybe a more user-friendly error message would be useful here in this case.

ImJasonH: It's possible it's a typo, but it could also be the wrong account logged in at the time. I'm not sure I could detect the error messages for one or the other well enough to give a better message.

Unless you mean something like `error checking permissions -- check for typos and try again`?

dlorenc: Yeah, just adding something like: "make sure you entered the right repository %s and that you are authenticated correctly and try again"

ImJasonH: Done:

```
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "gcr.io/g0000000000gle/foo:latest": UNKNOWN: Project 'projects/g0000000000gle' not found or deleted.
```

iciclespider:
This breaks pushing to remote registries which are http based.

georgecookeIW:
@iciclespider Confirmed, I am getting the following error when pushing to a HTTP based registry:
`$ kubectl logs pod/registry-test-kaniko-push-np7wg
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "kubernetes.docker.internal:5000/registrytest/kanikopush:latest": Get https://kubernetes.docker.internal:5000/v2/: http: server gave HTTP response to HTTPS client`
Using **gcr.io/kaniko-project/executor:debug-v0.10.0**
When I switch back to v0.9.0 this goes away.
I have tried using the --insecure, --skip-tls-verify flags and none of them seem to have any effect on the push permission check.
Executor command:
`/kaniko/executor --destination=kubernetes.docker.internal:5000/registrytest/kanikopush:latest --insecure *.internal`

PR #624, opened 2019-03-25
Reduce log level for adding file message

mcfedr:
This makes for massive logs and with it at debug level will be simpler to look at the output of kaniko


bithavoc:
This is cool, is there a way to change the log level altogether? #633

PR #627, opened 2019-03-28
skip verify repo tls on checking push permission if configured in opts

zhleonix:
When checking push permission, the skip-tls-verify options are not used, which causes failed operation on our private registry without importing root cert into the kaniko image.:
`error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "private.repo/test/kaniko_demo:latest": Get https://private.repo/v2/: x509: certificate signed by unknown authority
`

A fix is required to add skipTlsVerify into httpTransport used in CheckPushPermissions.


zhleonix:
I have signed it!

zhleonix:
will make another commit and submit again.

Issue #628, opened 2019-03-28
--skip-tls-verify option is not used on checking push permission 


zhleonix:
**Actual behavior**
When checking push permission, the --skip-tls-verify options are not used, which causes failed operation on our private registry without importing root cert into the kaniko image.:
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "private.repo/test/kaniko_demo:latest": Get https://private.repo/v2/: x509: certificate signed by unknown authority

**Expected behavior**
--skip-tls-verify should be considered even on CheckPushPermissions.

**To Reproduce**
1. Use an internal private repository without valid public certs
2. Use kaniko to build the image and push it to private repository
3. Before push, kaniko checks the permission but cannot verify its certificate

**Additional Information**
 latest version on master branch

lolychee:
`--insecure` also not work

PR #629, opened 2019-03-28
Fix checking push permission without skipping tls on configured (#628)

zhleonix:
When checking push permission, the skip-tls-verify options are not used, which causes failed operation on our private registry without importing root cert into the kaniko image.:
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "private.repo/test/kaniko_demo:latest": Get https://private.repo/v2/: x509: certificate signed by unknown authority

A fix is required to add skipTlsVerify into httpTransport used in CheckPushPermissions.

### DIFF
--- a/pkg/executor/push.go
+++ b/pkg/executor/push.go
@@ -63,7 +63,15 @@ func CheckPushPermissions(opts *config.KanikoOptions) error {
 		if checked[destRef.Context().RepositoryStr()] {
 			continue
 		}
-		if err := remote.CheckPushPermission(destRef, creds.GetKeychain(), http.DefaultTransport); err != nil {
+
+		// Create a transport to set our user-agent.
+		tr := http.DefaultTransport
+		if opts.SkipTLSVerify || opts.SkipTLSVerifyRegistries.Contains(destRef.Repository.Registry.Name()) {
+			tr.(*http.Transport).TLSClientConfig = &tls.Config{
+				InsecureSkipVerify: true,
+			}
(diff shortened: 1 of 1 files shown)


dbolshak:
@ImJasonH  Could you please review this #629 ?

Also looks like there are too many options to manage TLS issues, is there any guide about how to do this properly?

pkg/executor/push.go
```diff
@@ -63,7 +63,15 @@ func CheckPushPermissions(opts *config.KanikoOptions) error {
 		if checked[destRef.Context().RepositoryStr()] {
 			continue
 		}
-		if err := remote.CheckPushPermission(destRef, creds.GetKeychain(), http.DefaultTransport); err != nil {
+
+		// Create a transport to set our user-agent.
+		tr := http.DefaultTransport
```
ImJasonH: Should there be (is there already?) some kind of `GetTransport()` method that does this in a shared location, sort of like how `creds.GetKeychain()` does? I'm not familiar with all of the available `opts`, and even if I were, if a new opt was added that configured HTTP requests like this, it'd be nice to have that applied in one common location.

I don't think it should necessarily block this PR, but it'd be nice to get feedback from someone more familiar with kaniko.

abergmeier:
IMO it would be better to extract this into a function. Did that in #663.

tejal29:
@zhleonix would like to rebase this ?

PR #630, opened 2019-03-28
Cut 0.10.0 release.

dlorenc:


discordianfish:
Maybe have a quick look at #631 - probably some edge case or an issue on my side, but given that my usage seem straight forward it might be a release blocker?

priyawadhwa:
done with #701

Issue #631, opened 2019-03-28
Regression: #525 causes missing files when caching is enabled in some situations


discordianfish:
**Actual behavior**
First I observed that after enabling caching and building once, subsequent builds appear to miss the files the cached steps should have created. Disabling the cache fixes the problem.

When running the builds with the debug image, I'm see that it uses the cache here as expected:
```
INFO[0001] Using caching version of cmd: RUN mix local.hex --force
```
This should have created files that appears to be missing the the next steps, since they fail with:
```
NFO[0045] Using files from context: [/src]
INFO[0045] COPY . /app
INFO[0046] Taking snapshot of files...
INFO[0046] RUN mix compile
INFO[0046] cmd: /bin/sh
INFO[0046] args: [-c mix compile]
"File operation error: badarg. Target: .. Function: read_file_info. Process: code_server."
=ERROR REPORT==== 28-Mar-2019::15:55:29 ===
File operation error: badarg. Target: .. Function: read_file_info. Process: code_server.
15:55:29.584 [error] 'File operation error: badarg. Target: /root/.mix/archives/hex-0.19.0/hex-0.19.0/ebin. Function: read_file_info. Process: code_server.' Could not find Hex, which is needed to build dependency :phoenix
2019-03-28 15:55:29 std_error           ** (Mix) Could not find an SCM for dependency :phoenix from Gateway.Mixfile
Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn] error building image: error building stage: waiting for process to exit: exit status 1
```

The Dockerfile is pretty generic:
```
FROM elixir:1.6.1
MAINTAINER redacted

ENV MIX_ENV prod
ENV PORT 3000
ENV METRICS_PORT 3001

RUN mix local.hex --force
RUN mix local.rebar --force

WORKDIR /app

ADD mix.exs /app/mix.exs
ADD mix.lock /app/mix.lock
ADD config /app/config
RUN mix deps.get && mix compile

ADD . /app
RUN mix compile

CMD mix phoenix.server
```

**Expected behavior**
It should just work™️ 

**To Reproduce**
Haven't found an good way to reproduce (yet).

**Additional Information**
 - Kaniko Image (fully qualified with digest): gcr.io/kaniko-project/executor@sha256:d9fe474f80b73808dc12b54f45f5fc90f7856d9fc699d4a5e79d968a1aef1a72

discordianfish:
[...]

PR #639, opened 2019-04-03
Fix kaniko caching

discordianfish:
Three changes to help with the caching:

- Had to revert #525, see #631 
- Add back debug logging of cache key, I think this is useful
- Remove build args in cache key, This *should* be save, given that the commands will have the args included when the cache key gets built.

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_cache
+++ b/integration/dockerfiles/Dockerfile_test_cache
@@ -16,15 +16,7 @@
 # If the image is built twice, /date should be the same in both images
 # if the cache is implemented correctly
 
-FROM alpine as base_stage
-
-RUN mkdir foo && echo base_stage > foo/base
-
-FROM base_stage as cached_stage
-
-RUN echo cached_stage > foo/cache
-
-FROM cached_stage as bug_stage
-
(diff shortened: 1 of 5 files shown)


dlorenc:
Sorry for the delay!

orf:
We might be experiencing an issue relating to this: https://github.com/GoogleContainerTools/kaniko/issues/690

> Remove build args in cache key, This should be save, given that the commands will have the args included when the cache key gets built.

Are you *sure* this is correct? It seems from what I am seeing that it is indeed not safe.

discordianfish:
@orf I'm not sure - I just thought it would be save given that the build arg modifies the hash of the command. But I haven't found a obvious way to verify it. I hoped I'd get input from the maintainers to discuss that.

orf:
I’m quite concerned a new release went out with this. @priyawadhwa did you see my issue or my comment? This seems to explicitly break caching :/

8ear:
Is there any news?

priyawadhwa:
Hey @8ear @orf sorry for the delay, I haven't had a whole lot of time recently to look into this. If either of you is willing to open a PR, please cc me and I can take a look at it!

Issue #303, opened 2018-08-23
Info log level is too verbose


lubo:
I've been building an image containing many files and directories and my build log is pretty much just `Not adding /foo/bar because it was added by a prior layer`, `Copying file /foo/bar to /bar/foo` and `Taking snapshot of files [<thousands of filenames>]...` messages. How about dropping log level for these messages to `debug`?

RafalSkolasinski:
This still seems to be a problem when building images using `gcr.io/kaniko-project/executor:debug`.
Any clue when the fix will propagate?

jobec:
Same here.
I'm also using the debug version (as specified here: https://docs.gitlab.com/ee/ci/docker/using_kaniko.html) and it is causing almost unusable build logs.

slavoutich:
Kaniko executor has `--verbosity` flag, I assume setting it to `warn` will fix the issue for you (doing it now).

jobec:
This line is the problem:

https://github.com/GoogleContainerTools/kaniko/blob/c8fabdf6e43b19f6a223f1d0b06e127d0774bd7e/pkg/snapshot/snapshot.go#L181

It was changed from `debug` to `info` in this commit: 
https://github.com/GoogleContainerTools/kaniko/commit/1bf4421047570790ced358f60118b58709628c15#diff-736acbb393b480a7c2fa1ab622672582R181

@priyawadhwa can this be reopened?
If not fixed, this is going to explode in the next release.

priyawadhwa:
@jobec thanks for pointing this out, looks like @sharifelgamal is on it!

petervandenabeele:
FYI, it's exploding now, I cannot build anymore: on gitlab, I get: 
`Job's log exceeded limit of 4194304 bytes`

I will try the work-around like `--verbosity warn` (Ref. https://github.com/GoogleContainerTools/kaniko/issues/638)

sharifelgamal:
This should now be fixed at HEAD.

PR #640, opened 2019-04-12
this is quite spammy in my multistage build

scones:
.. and as i am surely not the only one, move it down to debug.


scones:
I signed it!

sharifelgamal:
Fixes #303

Thanks for fixing this!

PR #646, opened 2019-04-22
Prefer platform that is currently running

lisa:
Prefer platform that is currently running

The goal of this changeset is to prefer the current platform (`v1.Platform`, and `GOARCH`) to better support multiple (non-amd64) architectures. There are two targets for this: 1) the Makefile (and Dockerfiles), and 2) select the appropriate image manifest where possible.

1. Makefile section Previously, the Makefile had set `GOARCH` to `amd64`, which is problematic for portability. Setting it to the value of `go env GOARCH` if not already set seems a good compromise.

**Note**: This changeset does *not* address the other inherent portability issues in the Dockerfiles (a separate PR may be forthcoming), but be aware that while the `Makefile` and code will support the Platform and GOARCH "auto-detection", the Dockerfiles still do not. All this section does is make a `GOARCH` kaniko binary.

2. For container images that have multiple manifests, select the one for my platform

If an image has multiple manifests (such as debian:stretch, whose manifest list is `linux/amd64, linux/arm/v5, linux/arm/v7, linux/arm64, linux/386, linux/ppc64le, linux/s390x`), the previous behaviour was that kaniko, by way of go-containerregistry's [current](https://github.com/google/go-containerregistry/blob/8621d738a07bc74b2adeafd175a3c738423577a0/pkg/v1/remote/image.go#L92-L97) [implementation](https://github.com/google/go-containerregistry/blob/8621d738a07bc74b2adeafd175a3c738423577a0/pkg/v1/remote/image.go#L36-L39), is to use the `linux/amd64`, image manifest if it exists.

This change uses `runtime` to construct a `v1.Platform` reference which is used as a [platform option](https://github.com/google/go-containerregistry/blob/8621d738a07bc74b2adeafd175a3c738423577a0/pkg/v1/remote/options.go#L59-L64) to `remote.Image` so that go-containerregistry will attempt to use the current platform's manifest, if it exists.

Motivation behind this change is to bring arm64 support to kaniko.


pkg/util/util.go
```diff
@@ -127,3 +132,37 @@ func SHA256(r io.Reader) (string, error) {
 	}
 	return hex.EncodeToString(hasher.Sum(make([]byte, 0, hasher.Size()))), nil
 }
+
+type ReproducibleManifest struct {
+	Layers []v1.Descriptor
+	Config v1.Config
+}
+
+func ReproducibleDigest(img partial.WithManifestAndConfigFile) (string, error) {
```
tejal29: I dont think this function is used anywhere.

tejal29:
@lisa Are you still interested in pursuing this PR?
Please rebase.

Thanks
Tejal

lisa:
> @lisa Are you still interested in pursuing this PR?
> Please rebase.
> 
> Thanks
> Tejal

Apologies, yes, I am. I'll try and tend to it this weekend.

lisa:
@tejal29 The conflicts should be resolved.

sgielen:
I am running Kaniko on arm64 and building images for arm64. Without the changes of this Diff applied, a simple build of the following Dockerfile:

```
FROM ubuntu
ENTRYPOINT ["/bin/bash", "-c", "echo hello"]
```

...results in a successful build but the following error when starting the resulting image on arm64:

```
$ kubectl logs kaniko-output
standard_init_linux.go:211: exec user process caused "exec format error"
```

However, with the changes to `util/` applied to my local version of Kaniko, I can confirm that the image built from the same Dockerfile now results in a proper arm64 image:

```
$ kubectl logs kaniko-output
hello
```

lisa:
I noticed this needs another rebase, so I'll take care of that when I have an opportunity.

PR #651, opened 2019-04-30
Remove test foo file from git repository

dmathieu:
This empty file is only created by tests, and doesn't need to be tracked in git.


dmathieu:
Duplicate of GoogleContainerTools/kaniko/pull/635

PR #652, opened 2019-04-30
Make the destination image optional

dmathieu:
When debugging builds with docker, we may not want to push the image anywhere.
If no destination is specified, the `--no-push` parameter will then automatically be set.


tejal29:
This is not going to work since cache and destination both are optional now.

PR #663, opened 2019-05-15
Add SkipVerify support to CheckPushPermissions.

### LINKED ISSUE #628: --skip-tls-verify option is not used on checking push permission

abergmeier:
Extract makeTransport, which allows using the current mechanism used for
pushing.

Fixes #628.

### DIFF
--- a/pkg/executor/push.go
+++ b/pkg/executor/push.go
@@ -64,7 +64,10 @@ func CheckPushPermissions(opts *config.KanikoOptions) error {
 		if checked[destRef.Context().RepositoryStr()] {
 			continue
 		}
-		if err := remote.CheckPushPermission(destRef, creds.GetKeychain(), http.DefaultTransport); err != nil {
+
+		registryName := destRef.Repository.Registry.Name()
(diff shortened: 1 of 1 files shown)


orodbhen:
I've pulled master and applied this patch, and can confirm that it works.

sharifelgamal:
Thanks for the contribution! Sorry for the delay getting it merged.

PR #668, opened 2019-05-18
docs: fix document on DoBuild

orisano:
`dockerfile.Stages(opts)` do not unpack base image to root.


orisano:
improve PR description

PR #692, opened 2019-06-04
refactor: avoid twice parse Dockerfile

orisano:
avoid twice parse Dockerfile

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -361,6 +361,10 @@ func CalculateDependencies(opts *config.KanikoOptions) (map[int][]string, error)
 	if err != nil {
 		return nil, err
 	}
+	return calculateDependencies(stages, opts)
+}
+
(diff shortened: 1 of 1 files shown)


antechrestos:
@orisano @tejal29 I can contribute if needed. I understand what @orisano did

In the beginning of `DoBuild`, we have

`stages, err := dockerfile.Stages(opts)`

then couple of lines later

`crossStageDependencies, err := CalculateDependencies(opts)`

then at the beginning of `CalculateDependencies`: `stages, err := dockerfile.Stages(opts)`

Hence, making `CalculateDependencies` private and passing `stages` as parameters is a good option.

I can contribute if needed and find a way to enforce the face that the file gets loaded only once.

antechrestos:
@orisano This [pull request](https://github.com/GoogleContainerTools/kaniko/pull/1174/) made the job

PR #693, opened 2019-06-04
refactor: avoid waste code

orisano:
What is the meaning of this code? Is it a waste code?
```go
for p := range l {
    if strings.HasPrefix(filepath.Base(p), ".wh.") {
        delete(paths, p)
    }
    paths[p] = struct{}{}
}
```

### DIFF
--- a/pkg/snapshot/layered_map.go
+++ b/pkg/snapshot/layered_map.go
@@ -20,8 +20,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"path/filepath"
-	"strings"
 
 	"github.com/GoogleContainerTools/kaniko/pkg/timing"
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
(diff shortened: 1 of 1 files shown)


priyawadhwa:
Hi @orisano, this isn't waste code. If a file is added in a layer, and then deleted in a subsequent layer, it's prefixed with `.wh.` so indicate that the file should be deleted. `kaniko` uses this to extract the filesystem of the base image correctly.

orisano:
Thanks, @priyawadhwa.
Current code is removing whiteout file and adding whiteout file immediately

Is the code below correct?
```go
for p := range l {
    if strings.HasPrefix(filepath.Base(p), ".wh.") {
        delete(paths, strings.Trim(p, ".wh."))
    }
    paths[p] = struct{}{}
}
```

priyawadhwa:
Hey @orisano given 

```
p = /tmp/.wh.file
```
, this code is actually removing `/tmp/file` from paths and adding `/tmp/.wh.file`.

orisano:
sorry, @priyawadhwa.
please tell me the expected behavior.

cvgw:
Please open issues for discussions and questions related to the code.

Issue #575, opened 2019-02-21
Unreliable caching for multistage dockerfiles when stages are based on previous stages


twuttk:
**Actual behavior**
Caching for multistage dockerfiles with stages based on previous stages is unreliable and often doesn't work properly. Minimal working example below.

Caching fails at stage 2:
```
FROM debian:stable-slim AS stage1
RUN apt-get update
RUN touch stageone_file

FROM stage1 As stage2
RUN apt-get update
RUN touch stagetwo_file
```

Caching works:
```
FROM debian:stable-slim AS stage1
#RUN apt-get update
RUN touch stageone_file

FROM stage1 As stage2
RUN apt-get update
RUN touch stagetwo_file
```

Caching works:
```
FROM debian:stable-slim AS stage1
RUN apt-get update
RUN touch stageone_file

FROM stage1 As stage2
#RUN apt-get update
RUN touch stagetwo_file
```

Please note that this is a different issue than described and fixed in https://github.com/GoogleContainerTools/kaniko/issues/458

**Expected behavior**
Caching should always function properly, regardless of the amount of stages and their relation to each other.

**To Reproduce**
Build images with dockerfiles quoted above

**Additional Information**
- this was tested with the latest debug image.

dlorenc:
Hey,

Thanks for the report! I'm able to reproduce this with the images tagged :latest, but not with any built from HEAD. Could you try using one of these images and see if it still repros?

gcr.io/kaniko-project/executor:4b7e2b3a064fe0d558d2a87173bc50e7de31f441
gcr.io/kaniko-project/executor:debug-4b7e2b3a064fe0d558d2a87173bc50e7de31f441

I'm not sure off the top of my head what recent change would have fixed this, though.

temiwottku:
Hi, 

thanks for the quick answer. For me, using the images built from HEAD doesn't seem to change anything, caching for the example from above still fails at stage 2, while the other to examples work fine.

I'm running kaniko within minikube and mount my code into the container as separate volume. Here is my pod.yaml:

```
apiVer
sion: v1
kind: Pod
metadata:
  name: kaniko
spec:
  containers:
    - name: kaniko
      image: gcr.io/kaniko-project/executor:4b7e2b3a064fe0d558d2a87173bc50e7de31f441
      args: ["--context=/repo",
             "--dockerfile=multistage-caching-fails.dockerfile",
             "--destination=<private registry>",
             "--cache=true"]
      volumeMounts:
[...]

PR #715, opened 2019-07-16
Don't recreate cached layers

eyusupov:
This caching optimization reuses existing layer from a cache image for a RUN command instead of unpacking it and recreating. 
It also helps with multi-stage build caching issues like this https://github.com/GoogleContainerTools/kaniko/issues/575. The issue is that when recreating cached layers, some attributes like timestamps are not recreated and the digest for the new first stage image was different, breaking caching for subsequent stages.

### DIFF
--- a/pkg/commands/workdir.go
+++ b/pkg/commands/workdir.go
@@ -37,9 +37,7 @@ type WorkdirCommand struct {
 // For testing
 var mkdir = os.MkdirAll
 
-func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.BuildArgs) error {
-	logrus.Info("cmd: workdir")
-	workdirPath := w.cmd.Path
+func updateWorkdir(workdirPath string, config *v1.Config, buildArgs *dockerfile.BuildArgs) error {
 	replacementEnvs := buildArgs.ReplacementEnvs(config.Env)
 	resolvedWorkingDir, err := util.ResolveEnvironmentReplacement(workdirPath, replacementEnvs, true)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


eyusupov:
I signed it!

pkg/commands/workdir.go
```diff
@@ -58,6 +58,9 @@ func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile
 		logrus.Infof("Creating directory %s", config.WorkingDir)
 		w.snapshotFiles = append(w.snapshotFiles, config.WorkingDir)
 		return mkdir(config.WorkingDir, 0755)
+	} else {
+		// Force creation of empty cache image to avoid unpacking this layer on reruns
+		w.snapshotFiles = nil
```
abayer: There's a test failing expecting to get `[]string{}`, not a nil, presumably as a result of this change.

eyusupov: :+1: Working on it, I'm seeing some cache misses on the image that I'm trying to build. I'll add some unit and integration tests for caching and FS unpacking.

eyusupov: Fixed

eyusupov:
There is an integration test failing for `CMD` command due to `ArgsEscaped` flag in the metadata having incorrect value. It doesn't seem related to my changes and it seems to be failing on master locally for me too. Am I missing something or is it failing for others too?
Here's the output I'm getting on master (I commented out all tests other than `Dockerfile_test_cmd` fwiw)

```
--- FAIL: TestRun (0.00s)
    --- FAIL: TestRun/test_Dockerfile_test_cmd (0.92s)
        integration_test.go:224: diff = [
              {
                "Image1": "172.17.0.2:5000/docker-dockerfile_test_cmd",
                "Image2": "172.17.0.2:5000/kaniko-dockerfile_test_cmd",
                "DiffType": "File",
                "Diff": {
                  "Adds": null,
                  "Adds": null,
                  "Dels": null,
                  "Mods": null
                }
              },
              {
                "Image1": "172.17.0.2:5000/docker-dockerfile_test_cmd",
                "Image2": "172.17.0.2:5000/kaniko-dockerfile_test_cmd",
                "DiffType": "Metadata",
                "Diff": {
                  "Adds": [
                    "ArgsEscaped: true"
                  ],
                  "Dels": [
                    "ArgsEscaped: false"
                  ]
                }
              }
            ]
        integration_test.go:227: json: cannot unmarshal string into Go struct field .Adds of type integration.fileDiff
        integration_test.go:227: []integration.diffOutput differ (-got, +want): {[]integration.diffOutput}[1].Diff.Adds[0->?]:
                -: integration.fileDiff{}
                +: <non-existent>
            {[]integration.diffOutput}[1].Diff.Dels[0->?]:
                -: integration.fileDiff{}
                +: <non-existent>
```

tejal29:
@eyusupov Thanks for the PR. Can you please add some tests?

eyusupov:
Will do soon!

tejal29:
@eyusupov This looks great! Please add tests and we will try to get this in!

eyusupov:
Sorry for not coming back to it sooner, will try to do it as soon as I can

strax:
@eyusupov Is there anything I could do to help with this? We too are affected by #575 and it would be great to get these changes shipped.

raijinsetsu:
This is huge for my team. The step of creating the new archive from exactly the same content as the prior archive is causing a significant slow-down in our build pipeline due to the number of files (this is a LARGE NodeJS project).

cvgw:
I'm going to close this. https://github.com/GoogleContainerTools/kaniko/pull/882 will be merged in favor

PR #717, opened 2019-07-19
fix unpacking archives via ADD

vm86:
Bugfix: unpacking archives with ADD cos it recieves only a full destination path and ignores WORKDIR's directions. 

Dockerfile example: 
```
FROM busybox

WORKDIR /app

ADD test_file.tar .
RUN ls /app
```


vm86:
I signed it!

vm86:
@SamyCoenen @priyawadhwa @tejal29

PR #723, opened 2019-07-24
Add checking image presence in cache prior to downloading it

### LINKED ISSUE #722: Warmer does not use existing cache
I am experimenting with kaniko in Gitlab CI pipelines. And I use caching extensively to provide faster pipeline execution. So I've tried to use kaniko-warmer for faster docker builds.
The problem is currently kaniko-warmer downloads images regardless of cache directory state.

I propose kaniko-warmer to check image existence in the cache directory prior to downloading it.

This will require minor code modifications. PR with them is coming.

valentinsoe:
This changes allow to use kaniko-warmer multiple times without unnecessary docker image downloads.
To check image presence in cache directory I'm using existing cache function that is used by kaniko-executor.
I've considered building separate function to only check image presence, but it will have pretty much the same code.
Questionable decision is to embed CacheOptions type to KanikoOptions and WarmerOptions. Probably this should be resolved by creating interface providing needed options and implement it both mentioned structs. But I've struggled to get a meaningfull name to it.
To replicate previous behaviour of downloading regardless of cache state I've added --force(-f) option.

This changes provides crucial speed-up when downloading images from remote registry is slow.

Closes #722

### DIFF
--- a/cmd/warmer/cmd/root.go
+++ b/cmd/warmer/cmd/root.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/GoogleContainerTools/kaniko/pkg/cache"
 	"github.com/GoogleContainerTools/kaniko/pkg/config"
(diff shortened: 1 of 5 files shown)


valentinsoe:
I signed it!

sharifelgamal:
Thanks for your contribution!

PR #735, opened 2019-08-08
Bailout when there is not enough input arguments

xueshanf:
Currently the code carries on even missing required arguments.


xueshanf:
@googlebot I signed it!

Issue #731, opened 2019-08-05
executor in GCB is ignoring working directory setup field


liiri:
When specifying a build step with `gcr.io/kaniko-project/executor` image, the step working directory configuration is ignored.

1. Create a repository with a Dockerfile and code nested under some `src` dir. i.e:
```
src
  |--Dockerfile
  |--file.txt
|--cloudbuild.yaml
```
2. In the Dockerfile, assert working directory is `src` and reference some files in there. i.e
```
FROM node:10
COPY file.txt /file.txt
```
2. Create a GCB definition in repo root, i.e
```
steps:
- name: 'gcr.io/kaniko-project/executor:latest'
  id: Build test docker image
  args: ['--destination=gcr.io/$PROJECT_ID/test:$SHORT_SHA', '--cache=true', '--cache-ttl=12h']
  dir: './src'
```
3. Define a trigger and trigger it.

The build fails with error stating it cannot find the files referenced in the Dockerfile.
So the Dockerfile is located, but the run context is still the root directory.

```
Step #0 - "Build test docker image": error building image: error building stage: lstat /workspace/file.txt: no such file or directory
```

There is a simple workaround, by adding the following to the args: `'--context=/workspace/src'`.

In addition, specifying a relative path in context doesn't work, and hardcoding this directory is not optimal.
For `'--context=./src'`:
```
Step #0 - "Build test docker image": error building image: error building stage: lstat src/file.txt: no such file or directory
```
(since src is search for under `/` and not under current wd `/workspace`)

tejal29:
@liiri, kaniko when no Kaniko is not aware what the GCB step.dir. Kaniko executes from the root dir "/" and does a `os.chdir`. We will ad some documentation to highlight it

huguesalary:
@tejal29 Would you mind explaining a bit what is the rationale behind doing a `Chdir("/")`? I'm asking because this is causing quite a few issues on my end and I can't seem to find a good reason for doing that.

tejal29:
This should be fixed by https://github.com/GoogleContainerTools/kaniko/pull/736

Issue #675, opened 2019-05-22
[bug] tarPath is not relative path, but it should be


kwladyka:
**Actual behavior**
`docker run --rm -w /workspace -v (pwd):/workspace gcr.io/kaniko-project/executor --tarPath=image.tar --context=auth-provider --no-push`

`image.tar` is saved who know where. If I change it to ` --tarPath=/workspace/image.tar` then it works.

**Expected behavior**
Save it in right relative path. In case about in `/workspace/image.tar`.

**Additional Information**
I need it to produce tar in relative path in Concourse (cicd). It is random path so it not constant.

Jukie:
Relative path to what? In this case image.tar is being saved at `/image.tar` because you aren't providing a path to save to, so the root `/...` is assumed.

kwladyka:
It should be saved in `/workspace/image.tar` because I am running command from `/workspace` and I use relative path `image.tar`.

kwladyka:
In practice I run it from `/tmp/fooX/barY` in Concourse (cicd) and I need to save it in artefact directory `/tmp/fooX/barY/artefact` by Concourse instead `docker run`.

But example with `docker run` is simple to recreate and easy to understand the issue.

tejal29:
this should be fixed by https://github.com/GoogleContainerTools/kaniko/pull/736

PR #736, opened 2019-08-09
Resolve relative paths to absolute paths in command line arguments

### LINKED ISSUE #732: Relative paths don't work with `--context`, `--tarPath`, and others...
# Actual behavior
Upon executing, the `executor` binary changes its root directory to `/` as seen here https://github.com/GoogleContainerTools/kaniko/blob/master/cmd/executor/cmd/root.go#L84

This prevents the use of relative directories in the `executor` command line arguments.

For example:
- Imagine a directory `/tmp/build` 
- with a subdirectory `myworkspace/` containing the `Dockerfile` (i.e. `/tmp/build/myworkspace/Dockerfile`)
- Running the following will fail:
```
cd /tmp/build
/kaniko/executor --dockerfile=Dockerfile --context=myworkspace/ --destination=image --tarPath=image.tar --cache=true --cache-dir=./cache/ --no-push
```
- With the error:
```
Error: error resolving dockerfile path: please provide a valid path to a Dockerfile within the build context with --dockerfile
```

# Expected behavior
Kaniko should allow the use of **relative paths** in its command line and honor them.

# Additional Information
The current behavior is also confusing because the `--dockerfile` argument **does** allow a relative path. 

That is, the following command line works: 
```
cd /tmp/build
/kaniko/executor --dockerfile=myworkspace/Dockerfile --destination=image --tarPath=image.tar --cache=true --cache-dir=./cache/ --no-push
```

But fails as soon as the Dockerfile tries to use a "workspace" file
```
error building image: error building stage: lstat /myworkspace/entrypoint.sh: no such file or directory
```

# Related issues
https://github.com/GoogleContainerTools/kaniko/issues/731
https://github.com/GoogleContainerTools/kaniko/issues/675

huguesalary:
The executor accepts a few arguments `--dockerfile`, `--context`, `--cache-dir` and
`--digest-file` that all represent paths. This commits allows those paths to
be relative to the working directory of the executor.

This code assumes `context` is an actual path and not a URL, which it should be when the new `resolveRelativePaths()` function introduced by this commit is called.

Fixes #732 #731 #675

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -81,6 +81,9 @@ var RootCmd = &cobra.Command{
 		if err := executor.CheckPushPermissions(opts); err != nil {
 			exit(errors.Wrap(err, "error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again"))
 		}
+		if err := resolveRelativePaths(); err != nil {
+			exit(errors.Wrap(err, "error resolving relative paths to absolute paths"))
+		}
 		if err := os.Chdir("/"); err != nil {
 			exit(errors.Wrap(err, "error changing to root dir"))
 		}
@@ -224,6 +227,37 @@ func resolveSourceContext() error {
 		return err
 	}
 	logrus.Debugf("Build context located at %s", opts.SrcContext)
+	return nil
+}
+
+func resolveRelativePaths() error {
+	optsPaths := []*string{
+		&opts.DockerfilePath,
+		&opts.SrcContext,
+		&opts.CacheDir,
+		&opts.TarPath,
+		&opts.DigestFile,
+	}
+
+	for _, p := range optsPaths {
+		// Skip empty path
+		if *p == "" {
+			continue
+		}
+		// Skip path that is already absolute
+		if filepath.IsAbs(*p) {
+			logrus.Debugf("Path %s is absolute, skipping", *p)
+			continue
+		}
+
+		// Resolve relative path to absolute path
+		var err error
+		relp := *p // save original relative path
+		if *p, err = filepath.Abs(*p); err != nil {
+			return errors.Wrapf(err, "Couldn't resolve relative path %s to an absolute path", *p)
+		}
+		logrus.Debugf("Resolved relative path %s to %s", relp, *p)
+	}
 	return nil
 }
 
--- a/integration/images.go
+++ b/integration/images.go
@@ -286,3 +286,33 @@ func (d *DockerFileBuilder) buildCachedImages(imageRepo, cacheRepo, dockerfilesP
 	}
 	return nil
 }
+
+// buildRelativePathsImage builds the images for testing passing relatives paths to Kaniko
+func (d *DockerFileBuilder) buildRelativePathsImage(imageRepo, dockerfile string) error {
+	_, ex, _, _ := runtime.Caller(0)
+	cwd := filepath.Dir(ex)
+
+	buildContextPath := "./relative-subdirectory"
+	kanikoImage := GetKanikoImage(imageRepo, dockerfile)
+
+	kanikoCmd := exec.Command("docker",
+		append([]string{"run",
+			"-v", os.Getenv("HOME") + "/.config/gcloud:/root/.config/gcloud",
+			"-v", cwd + ":/workspace",
+			ExecutorImage,
+			"-f", dockerfile,
+			"-d", kanikoImage,
+			"--digest-file", "./digest",
+			"-c", buildContextPath,
+		})...,
+	)
+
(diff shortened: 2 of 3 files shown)


cmd/executor/cmd/root.go
```diff
@@ -227,6 +230,37 @@ func resolveSourceContext() error {
...
+		if filepath.IsAbs(*p) {
+			logrus.Debugf("Path %s is absolute, skipping", *p)
+			continue
+		}
+
+		// Resolve relative path to absolute path
+		var err error
+		absp := *p
+		if *p, err = filepath.Abs(*p); err != nil {
+			return errors.Wrapf(err, "Couldn't resolve relative path %s to an absolute path", p)
```
huguesalary: On this line, `p` should be `*p` I'll push a fix asap.

huguesalary:
Sure thing, though I am unclear as to where exactly I should add my test files. Would you mind pointing this out to me?

huguesalary:
Also, I force pushed again because `logrus.Debugf("Resolved relative path %s to %s", *p, absp)` was outputting `Resolved relative path [absolute_path] to [old_relative_path]` (instead of `Resolved relative path [old_relative_path] to [absolute_path]`)

priyawadhwa:
For sure -- we have an [integration](https://github.com/GoogleContainerTools/kaniko/tree/master/integration) test folder with all of our integration tests. for example, this is the test for [caching](https://github.com/GoogleContainerTools/kaniko/blob/master/integration/integration_test.go#L352).

You could add another test here to build a kaniko image with relative paths passed in, using the [context](https://github.com/GoogleContainerTools/kaniko/tree/master/integration/context) directory for the context and using a Dockerfile in the [dockerfiles](https://github.com/GoogleContainerTools/kaniko/tree/master/integration/dockerfiles) directory to build. If the build completes succesfully, we can say the test has passed. Let me know if you have any more questions!

huguesalary:
@priyawadhwa Thanks! Should I create a new Dockerfile or should I follow the caching test model and write a test that builds every Dockerfile available in the test directory?

priyawadhwa:
Creating a new Dockerfile or using one of the existing ones and testing once should be sufficient.

cmd/executor/cmd/root.go
```diff
@@ -227,6 +230,37 @@ func resolveSourceContext() error {
 	return nil
 }
 
+func resolveRelativePaths() error {
```
tejal29: @huguesalary Thank you for your contributions. This looks really good. However can we please some unit test for this?

Can you refactor `resolveRelativePaths()` to accept opts? That ways we can easily write unit tests


```suggestion
func resolveRelativePaths(opts config.KanikoOptions{}) error {
```

Now, you can creates tests where

```
func TestResolveRelativePaths(t *testing.T) {
	tests := []struct {
		description  string
		opts config.KanikoOptions
               expected config.KanikoOptions
               shdErr  bool 
	}{
            {
                description: "dockerfile path  is absolute",
                opts: config.KanikoOptions{
                           DockerfilePath: "/abs/dockerfile",
                        },
               expected: config.KanikoOptions{
                           DockerfilePath: "/abs/dockerfile",
                        },
            }

	}
	for _, tt := range tests {
           actual := resolveRelativePaths(tt.opts)
           t.CheckDeepEqualAndError(tt.shdErr, actual, tt.expected, tt.opts)
	}
}


```

huguesalary:
Hi @tejal29 sure thing I'll add your suggestion and some unit tests

trungdq88:
Hi, thank you for your contribution to this great repo. Just to put a note that I suspect this PR broke my build this morning:

I used `--context=dir://workspace` in my `cloudbuild.yaml`.

Build logs before:
```
Step #1: WORKDIR /code                                
Step #1: cmd: workdir                                 
Step #1: Changed working directory to /code           
Step #1: No files changed in this command, skipping snapshotting. 
Step #1: Using files from context: [workspace/build-pool/main] 
Step #1: COPY build-pool/main /code                   
Step #1: Taking snapshot of files...                  
Step #1: ENTRYPOINT ./something.sh 
Step #1: No files changed in this command, skipping snapshotting. 
Finished Step #1
```

Build logs after:
```
Step #1: WORKDIR /code                                
Step #1: cmd: workdir                                 
Step #1: Changed working directory to /code           
Step #1: No files changed in this command, skipping snapshotting. 
Step #1: error building image: error building stage: lstat /workspace/workspace/build-pool/main: no such file or directory
```

My config was incorrect. After I change the config to `--context=/workspace` it worked fine.

I would be appreciate if we could make releases backward compatible, even if the old behavior is incorrect because maybe there are people depend on it. Or at least put a note in the changelog.

If this PR did not cause the broken build as I suspected: I am sorry, please ignore this.

Thanks!

Issue #754, opened 2019-08-28
$PATH is not passed


prary:
**Actual behavior**
Dockerfile 
```
FROM something/lrh7
RUN env
```
Do a docker build, output shows 
```
Step 2/3 : RUN env
 ---> Running in a7ea3e88b5a6

HOSTNAME=a7ea3e88b5a6
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/
SHLVL=1
HOME=/root
_=/usr/bin/env
```
Do skaffold build  -v=debug
```
INFO[0025] RUN env
INFO[0025] cmd: /bin/sh
INFO[0025] args: [-c env]
PWD=/
SHLVL=1
HOME=/root
_=/usr/bin/env
````
**Expected behavior**
Path variable should be retained

**Comment by prary** [comment]
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=34fbfd2c2006
TERM=xterm
HOME=/root

**Comment by prary** [comment]
If we do echo $PATH then
```
INFO[0021] RUN echo $PATH
INFO[0021] cmd: /bin/sh
INFO[0021] args: [-c echo $PATH]
/usr/local/bin:/usr/bin
```
Don't know from where this $PATH is coming

fspaniol:
Related: #748

prary:
We should set the default path if $PATH isnt defined in base image
https://github.com/GoogleContainerTools/kaniko/blob/1d3cf8a67dda19deecf6c754696e52f4f24a1198/pkg/executor/build.go#L127

Should remove this condition

Issue #650, opened 2019-04-30
Default PATH different to docker-build's


djuarezg:
**Actual behavior**
On https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/constants/constants.go#L75 there is a default PATH for scratch, which seems like an attempt to replicate docker-build default PATH as in https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb/pkg/system/path.go#L12

The problem is that this causes a change of behaviour in Kaniko. While using an image without PATH as FROM adds a default PATH in docker-build, Kaniko does not, probably because of https://github.com/GoogleContainerTools/kaniko/issues/89

**Expected behavior**

A kaniko build with a base image without PATH should add the same default PATH as with docker-build.

djuarezg:
Any update?

PR #760, opened 2019-08-30
Setting PATH

prary:
Fixes, #754, #650 
Resolving #754 , Setting PATH variable when $PATH is empty in base image, making it similar to how docker handles it.


prary:
I will modify it if Path is empty then only set it to default

tejal29:
@prary its hard to see the effect. Can we add an integration test or unit test for this?

Issue #756, opened 2019-08-28
Kaniko seg faults while building


cblage:
**Actual behavior**
Kaniko will segmentation fault while building the docker image

**Expected behavior**
Kaniko will build the docker image successfully 

**To Reproduce**

1 - Setup build step on GCB with:

```
  - name: "gcr.io/kaniko-project/executor:latest"
    args:
      - --build-arg=JS_BUILD_ID=staging-$BUILD_ID
      - --dockerfile=Dockerfile.deployment
      - --destination=us.gcr.io/$PROJECT_ID/agentrisk-web:$COMMIT_SHA
      - --destination=us.gcr.io/$PROJECT_ID/agentrisk-web:latest-staging
      - --cache=true
      - --cache-ttl=720h
    id: "Build and push image"
    waitFor: ["Run tests"]
```

2 - Trigger build on GCB

3 - It will segfault with this:

```
ERROR: build step 3 "gcr.io/kaniko-project/executor:latest" failed: exit status 2
ERROR
Finished Step #3 - "Build and push image"
Step #3 - "Build and push image": 	/go/src/github.com/GoogleContainerTools/kaniko/cmd/executor/main.go:26 +0x2d
Step #3 - "Build and push image": main.main()
Step #3 - "Build and push image": 	/go/src/github.com/GoogleContainerTools/kaniko/vendor/github.com/spf13/cobra/command.go:800 +0x2b
Step #3 - "Build and push image": github.com/GoogleContainerTools/kaniko/vendor/github.com/spf13/cobra.(*Command).Execute(0x27a6660, 0xc420601f78, 0xc4200400b8)
Step #3 - "Build and push image": 	/go/src/github.com/GoogleContainerTools/kaniko/vendor/github.com/spf13/cobra/command.go:852 +0x30a
Step #3 - "Build and push image": github.com/GoogleContainerTools/kaniko/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x27a6660, 0x0, 0x1a31f54, 0x25)
Step #3 - "Build and push image": 	/go/src/github.com/GoogleContainerTools/kaniko/vendor/github.com/spf13/cobra/command.go:766 +0x2c1
Step #3 - "Build and push image": github.com/GoogleContainerTools/kaniko/vendor/github.com/spf13/cobra.(*Command).execute(0x27a6660, 0xc420038080, 0x6, 0x6, 0x27a6660, 0xc420038080)
[...]

PR #765, opened 2019-09-09
Fix SIGSEGV on file system deletion while building

Onlinehead:
Fix issue #756 
That error happens only if you have files/directories inside a path marked as `VOLUME`
Whitelisting and delete functions don't expect you have something inside a VOLUME path and deleting content inside that dir (which is a correct behavior) breaks processing files inside the dir.
The fix adds `isExists` check for file/directory before `info.IsDir()` and skip path processing if it is not exist.

### DIFF
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -121,8 +121,8 @@ func DeleteFilesystem() error {
 	logrus.Info("Deleting filesystem...")
 	return filepath.Walk(constants.RootDir, func(path string, info os.FileInfo, _ error) error {
 		if CheckWhitelist(path) {
-			if ! isExist(path) {
-				logrus.Debugf("Path  %s whitelisted, but not exists", path)
+			if !isExist(path) {
+				logrus.Debugf("Path %s whitelisted, but not exists", path)
 				return nil
 			}
 			if info.IsDir() {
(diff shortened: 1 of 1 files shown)


Onlinehead:
@googlebot I signed it!

pkg/util/fs_util.go
```diff
@@ -137,6 +141,13 @@ func DeleteFilesystem() error {
 		return os.RemoveAll(path)
 	})
 }
+// isExists returns tru if path exists
```
jonjohnsonjr: nit: `true`

pkg/util/fs_util.go
```diff
@@ -121,6 +121,10 @@ func DeleteFilesystem() error {
 	logrus.Info("Deleting filesystem...")
 	return filepath.Walk(constants.RootDir, func(path string, info os.FileInfo, _ error) error {
 		if CheckWhitelist(path) {
+			if ! isExist(path) {
+				logrus.Debugf("Path  %s whitelisted, but not exists", path)
```
jonjohnsonjr: nit: double space after `Path`

pkg/util/fs_util.go
```diff
@@ -121,6 +121,10 @@ func DeleteFilesystem() error {
 	logrus.Info("Deleting filesystem...")
 	return filepath.Walk(constants.RootDir, func(path string, info os.FileInfo, _ error) error {
 		if CheckWhitelist(path) {
+			if ! isExist(path) {
```
jonjohnsonjr: this needs to be run through `gofmt` to get travis to pass

Rayyis:
@Onlinehead Thanks for the fix. Is there a workaround until this gets merged? or is there something to do to speedup the process of releasing it? @jonjohnsonjr

PR #773, opened 2019-09-16
Fixes caching with COPY command

### LINKED ISSUE #742: Subsequent Builds fail when using cache
**Actual behavior**
When building with `--cache`, the first build succeeds, but subsequent builds fail. 

**Expected behavior**
If a build succeeded once, I expect the build to work every single time (as long as the code didn't change)

**To Reproduce**
Dockerfile, build context, logs, and other materials can be found [here](https://gist.github.com/BenHizak/b8f295bccc792506816725b4c9116b59)

### Additional Information
The Error is always as follows. (the folder is different)
```
INFO[0001] Found cached layer, extracting to filesystem 
error building image: error building stage: extracting fs from image: open /usr/bin/somefolder: not a directory. 
ERROR
ERROR: build step 0 "gcr.io/kaniko-project/executor:debug" failed: exit status 1
```

#### Kaniko Image
This problem was replicated with both images 
```
gcr.io/kaniko-project/executor:debug@sha256:a54d167d7c4b7ce0c7a622f17dcf473c652b29341b321ca507425c8fa3525842
gcr.io/kaniko-project/executor:latest@sha256:78d44ec4e9cb5545d7f85c1924695c89503ded86a59f92c7ae658afa3cff5400
```

donmccasland:
Fixes #742 

When copying files from a snapshot or tar, when the file is listed before the parent directory, sometimes it can get copied in without the parent dir being created.


pkg/commands/copy_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/GoogleContainerTools/kaniko/pkg/dockerfile"
+	"github.com/GoogleContainerTools/kaniko/testutil"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/moby/buildkit/frontend/dockerfile/instructions"
+	"github.com/sirupsen/logrus"
+)
+
+var copyTests = []struct {
+	name           string
+	SourcesAndDest []string
```
tejal29: you can make this private.

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,118 @@
...
+		Cmd:        nil,
+		Env:        []string{},
+		WorkingDir: "../../integration/context/",
+	}
+
+	for _, test := range copyTests {
+		t.Run(test.name, func(t *testing.T) {
+			dirList := []string{}
+
+			logrus.Infof("Running test: %v", test.name)
```
tejal29: usually `go test` will generate the name case as `TestCopyExecuteCmd/copy_foo_into_tempCopyExecuteTest` and display it when test fails.
You can skip extra logging.

pkg/util/fs_util.go
```diff
@@ -188,9 +188,12 @@ func extractFile(dest string, hdr *tar.Header, tr io.Reader) error {
...
 	case tar.TypeReg:
 		logrus.Debugf("creating file %s", path)
-		// It's possible a file is in the tar before its directory.
-		if _, err := os.Stat(dir); os.IsNotExist(err) {
+		// It's possible a file is in the tar before its directory,
+		// or a file was copied over a directory prior to now
+		fi, err := os.Stat(dir)
+		if os.IsNotExist(err) || !fi.IsDir() {
 			logrus.Debugf("base %s for file %s does not exist. Creating.", base, path)
+			os.RemoveAll(dir)
```
tejal29: this is dangerous. Are we sure we want to do this?

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+			cmd := CopyCommand{
+				cmd: &instructions.CopyCommand{
+					SourcesAndDest: test.SourcesAndDest,
+				},
+				buildcontext: "../../integration/context/",
+			}
+
+			buildArgs := copySetUpBuildArgs()
+			dest := cfg.WorkingDir + test.SourcesAndDest[len(test.SourcesAndDest)-1]
```
tejal29: Is there a reason we can't use `tempDir` ? The advantage of creating a temp and writing all the files in the temp are
1. We can delete the who dir in defer. In case the test got cancelled or killed, `defer` block is always executed and dirs are cleaned up.
```
dir, err := ioutil.TempDir("dir", "prefix")
if err != nil {
       t.Errorf("error creating temp dir %s", err)
	t.Fail()
}
defer os.RemoveAll(dir)
```

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,118 @@
...
+			fstat, ferr := fi.Stat()
+			if ferr != nil {
+				t.Error()
+			}
+			if fstat.IsDir() {
+				files, ferr := ioutil.ReadDir(dest)
+				if ferr != nil {
+					t.Error()
+				}
+				for _, file := range files {
```
tejal29: @dgageot Wrote this[ temp directory testutility](https://github.com/GoogleContainerTools/skaffold/blob/master/testutil/tmp.go) for skaffold.

The tests become much readable and very easy to test.,  e.g. https://sourcegraph.com/github.com/GoogleContainerTools/skaffold@1409a9b72718099c99b360ba6f19b2a965ff8b55/-/blob/pkg/skaffold/filemon/changes_test.go#L99

I would shameless copy it :) Beauty of open source and give @dgageot credit :)

Note:`kaniko` is under same license as `skaffold` so we shd fine :)

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,118 @@
...
+			logrus.Infof("test.expectedDest: %v", test.expectedDest)
+			logrus.Infof("dirList: %v", dirList)
+
+			testutil.CheckErrorAndDeepEqual(t, false, err, test.expectedDest, dirList)
+			os.RemoveAll(dest)
+		})
+	}
+}
+
+func copySetUpBuildArgs() *dockerfile.BuildArgs {
```
tejal29: Why do we need these? Can we send empty or nil args?

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+var copyTests = []struct {
+	name           string
+	SourcesAndDest []string
+	expectedDest   []string
+}{
+	{
+		name:           "copy foo into tempCopyExecuteTest/",
+		SourcesAndDest: []string{"foo", "tempCopyExecuteTest/"},
+		expectedDest:   []string{"foo"},
```
tejal29: i am guessing you are going to fix this later and just adding tests for now?
the expectedDest should be `temp/CopyExecuteTest/foo` right?

We should also add a test case where we are copying a dir to dir.
1. COPY dir dir/  - shd copy dir into dir/
2. COPY dir2 dir - shd copy dir2 as dir

It will easy to create these file layouts with a test util.

pkg/commands/copy_test.go
donmccasland: done. thx.

pkg/util/fs_util.go
donmccasland: Agreed.  We should still try to do a os.MkdirAll here, but if that fails we should fail out entirely.

pkg/commands/copy_test.go
donmccasland: I think this is overkill for the two tests I needed to write here, and I'd like my test to read like all the other tests in the project.

pkg/commands/copy_test.go
donmccasland: Line 71, cmd.ExecuteCommand does not work without a buildArgs structure set up.  I copied this from another test.

pkg/commands/copy_test.go
donmccasland: Okay, reworked the whole thing.

pkg/commands/copy.go
```diff
@@ -134,3 +135,46 @@ func (c *CopyCommand) FilesUsedFromContext(config *v1.Config, buildArgs *dockerf
 func (c *CopyCommand) MetadataOnly() bool {
 	return false
 }
+
+func (c *CopyCommand) RequiresUnpackedFS() bool {
+	return true
```
priyawadhwa: This if for unpacking the base image, right? If so, I would guess we don't need it for the copy command (unless I'm completely forgetting something which is totally possible)

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,168 @@
...
+		expectedDest:   []string{"tempCopyExecuteTest"},
+	},
+}
+
+func setupTestTemp() string {
+	tempDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		logrus.Fatalf("error creating temp dir %s", err)
+	}
+	//defer os.RemoveAll(tempDir)
```
priyawadhwa: nit: remove this comment

pkg/util/fs_util.go
```diff
@@ -188,9 +188,11 @@ func extractFile(dest string, hdr *tar.Header, tr io.Reader) error {
...
 	case tar.TypeReg:
 		logrus.Debugf("creating file %s", path)
-		// It's possible a file is in the tar before its directory.
-		if _, err := os.Stat(dir); os.IsNotExist(err) {
-			logrus.Debugf("base %s for file %s does not exist. Creating.", base, path)
+		// It's possible a file is in the tar before its directory,
+		// or a file was copied over a directory prior to now
+		fi, err := os.Stat(dir)
+		if os.IsNotExist(err) || !fi.IsDir() {
+			logrus.Infof("base %s for file %s does not exist. Creating.", base, path)
```
priyawadhwa: WDYT of changing this log level back to `debugf` so that the output isn't too cluttered?

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,168 @@
...
+		})
+	if cperr != nil {
+		logrus.Fatalf("error populating temp dir %s", cperr)
+	}
+
+	return tempDir
+}
+func TestCopyExecuteCmd(t *testing.T) {
+	tempDir := setupTestTemp()
+
```
priyawadhwa: add `defer os.RemoveAll(tempDir)` here

pkg/commands/copy_test.go
```diff
@@ -0,0 +1,168 @@
...
+				buildcontext: tempDir,
+			}
+
+			buildArgs := copySetUpBuildArgs()
+			dest := cfg.WorkingDir + "/" + test.sourcesAndDest[len(test.sourcesAndDest)-1]
+			logrus.Infof("dest dir: %s", dest)
+
+			err := cmd.ExecuteCommand(cfg, buildArgs)
+			if err != nil {
+				t.Error()
```
priyawadhwa: We can log the error with `t.Error(err)` throughout this test

pkg/commands/copy.go
donmccasland: It turns you absolutely need it if you are copying something into a directory that is expected to exist in the base image.

So, here's the logic for COPY:
- COPY fromsomething tosomething
   + if tosomething is a directory, or it ends with '/' then copy fromsomething INTO tosomething
   + if tosomething does not exist or it is a file, then copy fromsomething OVER tosomething

So, in the test case I added, the dockerfile says:
FROM blahblah

COPY context/foo /usr/bin

If the base image does not exist, then we copy the contents of context/foo into a file called /usr/bin, which is not intended, and not how 'docker build' behaves.

pkg/commands/copy_test.go
donmccasland: nice catch!

pkg/util/fs_util.go
donmccasland: sounds fine.

pkg/commands/copy.go
priyawadhwa: ah makes sense, thanks!

priyawadhwa:
I think you forgot to check in `Dockerfile_test_cache_copy`?

pkg/commands/copy_test.go
donmccasland: Actually no, in this case the expectedDest is correct.  And yes, we should add some more tests later to make it clearer.

The situation is that:
copy thing DirThatExists/
result is DirThatExists/thing

copy thing DirThatExists
result is DirThatExists/thing

copy thing DirThatDoesNOTExist/
result is DirThatDoesNOTExist/thing

copy thing DirThatDoesNOTExist
result is thing gets renamed to DirThatDoesNOTExcist

PR #782, opened 2019-09-18
double help text about skip-verify-tls

HerrmannHinz:
removed


HerrmannHinz:
@googlebot I signed it!

PR #783, opened 2019-09-18
fix missing version bump to 0.12.0 in Makefile

dirkcjelli:
The version bump to 0.12.0 is missing in the `Makefile` in tag v0.12.0, added it


dirkcjelli:
@googlebot I signed it!

PR #784, opened 2019-09-18
fix missing version bump to 0.12.0 in Makefile

dirkcjelli:
Version bump to v0.12.0 was missing in `Makefile` for tag v0.12.0 fixing it here.


tejal29:
Thanks @dirkcjelli. will merge this in today.

Issue #592, opened 2019-03-03
camelCase stage names do not work


michas2:
**Actual behavior**
I tried to build an existing multi stage Dockerfile containing stage names in camelCase.
Up to v0.6.0 kaniko failes to copy data from such a stage.
From v0.7.0 to latest it complains about "could not parse reference".
(For details see below.)

**Expected behavior**
I expected kaniko to build Dockerfiles the same way as `docker build` does.
Especially if `docker build` is fine with camelCase names kaniko should also be fine.

**To Reproduce**

Create a `Dockerfile` containing a camelcase stage name like this:
```
from busybox as fooBar
run touch /foo

from busybox as run
copy --from=fooBar /foo .
```
Try to build it:
```
$docker run --rm -v $PWD:/workspace gcr.io/kaniko-project/executor:v0.6.0 -d x
[...]
INFO[0007] copy --from=fooBar /foo .                    
error building image: error building stage: lstat /kaniko/fooBar/foo: no such file or directory
$ docker run --rm -v $PWD:/workspace gcr.io/kaniko-project/executor:v0.7.0 -d x
error building image: could not parse reference
INFO[0000] Downloading base image fooBar                
$ docker run --rm -v $PWD:/workspace gcr.io/kaniko-project/executor:v0.9.0 -d x
error building image: could not parse reference
INFO[0000] Downloading base image fooBar      
```
With `docker build` it works fine:
```
$ docker build .
Sending build context to Docker daemon  290.3kB
Step 1/4 : from busybox as fooBar
 ---> d8233ab899d4
Step 2/4 : run touch /foo
 ---> Running in cfc8529c5b0e
Removing intermediate container cfc8529c5b0e
 ---> 284f568e0eef
Step 3/4 : from busybox as run
 ---> d8233ab899d4
Step 4/4 : copy --from=fooBar /foo .
 ---> d8b33ff2822b
Successfully built d8b33ff2822b
```

**Additional Information**
 - Dockerfile 
   see above
 - Build Context
   No other files beside Dockerfile
 - Kaniko Image (fully qualified with digest)
   see above.

michas2:
BTW, is there a good way to debug situations like "/kaniko/fooBar/foo: no such file or directory"?
(With `docker build` I have always the chance to `docker run` the intermediate image, inspect the filesystem and manually run a failing command.)

skofgar:
Were you ever able to get this running?

dinvlad:
[...]

PR #787, opened 2019-09-22
Fix #770 Build fails when FROM image AS alias has capital letters 

### LINKED ISSUE #770: Build fails when FROM image AS alias has capital letters
**Actual behavior**
Named build stages (FROM someimage as myname) fail when name begins with a Capital letter
Example:

1. `FROM python:3.7.2 AS library` ✔ builds successfully (note that there are no capital letters)
2. `FROM python:3.7.2 AS Library` ❌ fails (note that there is a capital letter)
Error shown
```
Downloading base image Library               
error building image: could not parse reference
ERROR
```
**Expected behavior**
When using named stages, capital letters should not break the build process
**To Reproduce**
Please see [this gist](https://gist.github.com/BenHizak/a792064af4dc49db07876815439aabbf)

N4rm0:
The root cause of the issue is a lowercasing of the stage name when parsing the Docker file.

Fix #770

### DIFF
--- a/pkg/dockerfile/dockerfile_test.go
+++ b/pkg/dockerfile/dockerfile_test.go
@@ -28,12 +28,18 @@ func Test_resolveStages(t *testing.T) {
 	dockerfile := `
 	FROM scratch
 	RUN echo hi > /hi
-	
+
 	FROM scratch AS second
 	COPY --from=0 /hi /hi2
-	
-	FROM scratch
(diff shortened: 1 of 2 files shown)


N4rm0:
It should fix also #592

N4rm0:
Closing this MR as the fix should be done into buildkit repo

PR #811, opened 2019-10-09
Details about --tarPath usage improved

ruzickap:
Fixes `#572`.

**Description**

`--tarPath=` parameter doesn't work properly without `--destination` as it's described in #572 .

I'm adding it to the documentation.


priyawadhwa:
Thanks for adding this!

PR #817, opened 2019-10-14
registry with .local and scheme https push error

wutongjie23hao:
when I try to push an image to a registry that ends with .local , the error occured.

the reason is that the go-containerregistry library has a special rules. This rule is that if registry ends with ``.local``  the scheme of the push registry will be appoint to ``http``.

the code snippet is as below:
```golang
	if reLocal.MatchString(r.Name()) {
		return "http"
	}
```

the reLocal defines as below:
```golang
// Detect more complex forms of local references.
var reLocal = regexp.MustCompile(`.*\.local(?:host)?(?::\d{1,5})?$`)
```

**Description**

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Bug fixes

```

### DIFF
--- a/pkg/executor/push.go
+++ b/pkg/executor/push.go
@@ -46,7 +46,7 @@ import (
 )
 
 type withUserAgent struct {
-	t http.RoundTripper
+	t        http.RoundTripper
 	insecure bool
 	registry string
 }
--- a/pkg/executor/push.go
+++ b/pkg/executor/push.go
@@ -47,6 +47,7 @@ import (
 
 type withUserAgent struct {
 	t http.RoundTripper
+	insecure bool
 }
 
 const (
@@ -59,6 +60,9 @@ func (w *withUserAgent) RoundTrip(r *http.Request) (*http.Response, error) {
 		ua = append(ua, upstream)
 	}
 	r.Header.Set("User-Agent", strings.Join(ua, ","))
+	if !w.insecure {
+		r.URL.Scheme = "https"
+	}
 	return w.t.RoundTrip(r)
 }
 
@@ -88,7 +92,8 @@ func CheckPushPermissions(opts *config.KanikoOptions) error {
 			destRef.Repository.Registry = newReg
 		}
 		tr := makeTransport(opts, registryName)
-		if err := remote.CheckPushPermission(destRef, creds.GetKeychain(), tr); err != nil {
+		rt := &withUserAgent{t: tr, insecure: opts.Insecure}
+		if err := remote.CheckPushPermission(destRef, creds.GetKeychain(), rt); err != nil {
 			return errors.Wrapf(err, "checking push permission for %q", destRef)
 		}
 		checked[destRef.Context().RepositoryStr()] = true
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
> This rule is that if registry ends with .local the scheme of the push registry will be appoint to http.

It's a little more subtle than that -- we'll always try https first and fallback to http under certain situations (e.g. in your .local case, or if it's explicit), see here:

https://github.com/GoogleContainerTools/kaniko/blob/602657930f800d3fd7871ff647206a39e94bcd81/vendor/github.com/google/go-containerregistry/pkg/v1/remote/transport/ping.go#L69-L75

We determine whether to use http or https based on the ping response (e.g. [here](https://github.com/GoogleContainerTools/kaniko/blob/602657930f800d3fd7871ff647206a39e94bcd81/vendor/github.com/google/go-containerregistry/pkg/v1/remote/transport/ping.go#L93)). It's possible that your registry is failing when we send the https ping, so we fall back to http, which succeeds, but then fails later on?

wutongjie23hao:
@jonjohnsonjr Thanks for your reply. I review the code again, because my push test fails again. 
I found the snippet you pointed out. The ping response's scheme enabled only in one authenticate case: bearer auth. Other two auth cases:``anonymous`` and ``basic`` will only use the response of reg.Scheme. Not pass function ping.
```golang
	switch pr.challenge.Canonical() {
	case anonymous:
		return t, nil
	case basic:
		return &basicTransport{inner: t, auth: auth, target: reg.RegistryStr()}, nil
	case bearer:
		// We require the realm, which tells us where to send our Basic auth to turn it into Bearer auth.
		realm, ok := pr.parameters["realm"]
		if !ok {
			return nil, fmt.Errorf("malformed www-authenticate, missing realm: %v", pr.parameters)
		}
		service, ok := pr.parameters["service"]
		if !ok {
			// If the service parameter is not specified, then default it to the registry
			// with which we are talking.
			service = reg.String()
		}
		bt := &bearerTransport{
			inner:    t,
			basic:    auth,
			realm:    realm,
			registry: reg,
			service:  service,
			scopes:   scopes,
			scheme:   pr.scheme,
		}
		if err := bt.refresh(); err != nil {
			return nil, err
		}
		return bt, nil
	default:
		return nil, fmt.Errorf("Unrecognized challenge: %s", pr.challenge)
	}
```

tejal29:
@warrenseymourx is this because you have a https registry setup which ends with `.local` ?
Is it secure registry on localhost ?

wutongjie23hao:
@tejal29  my secure registry is end with ``.local``.  But this is not the key problem. The key problem is that only ``bearer  auth`` cases will use ``https``.  Because the transport in ping.

wutongjie23hao:
@tejal29  my secure registry is end with .local. But this is not the key problem. The key problem is that ping response only work in  ``bearer auth`` case. And it will use https because of the transport in ping.
```
 schemes := []string{"https"} 
 if reg.Scheme() == "http" { 
 	schemes = append(schemes, "http") 
 } 
```
But it doesn't work at all when in the ``basic auth`` and ``anonymous``. And in this two auth cases don't use ping's transport, it use the default user-agent's transport. Thus, it will use 'http' because it ends with ``.local``, and the ping response's scheme doesn't work at all just like I comments above.

jonjohnsonjr:
> Other two auth cases:anonymous and basic will only use the response of reg.Scheme. Not pass function ping.

:man_facepalming: that's a huge oversight -- I'll fix this.

jonjohnsonjr:
This should be fixed by https://github.com/google/go-containerregistry/pull/575 if you want to just bump the dependency instead of wiring this up yourself.

wutongjie23hao:
@jonjohnsonjr Yes, it's dependencies modify is ok and best.  And modifying the dependency is my first thought. But because my use is kaniko , so I give up modifying the dependency. Now， modifying the dependency is good. Thank you for your reply

tejal29:
@wutongjie23hao closing this PR since, we need to bump go-containerregistry dep

Issue #439, opened 2018-11-08
FROM command does not honor ENV variables


mtsyganov:
**Actual behavior**
When I create a new image from the base image `registry.gitlab.com/gitlab-org/gitlab-ee/gitlab-assets-ee:11-5-stable-ee`, I get an error

**Expected behavior**
Image is downloaded and parsed successfully same as docker do.

**To Reproduce**
1. build an image from base image `registry.gitlab.com/gitlab-org/gitlab-ee/gitlab-assets-ee:11-5-stable-ee` 
2. receive error message
```
INFO[0000] Downloading base image "registry.gitlab.com/gitlab-org/gitlab-ee/gitlab-assets-ee:11-5-stable-ee" 
error building image: getting stage builder for stage 0: could not parse reference
```

**Additional Information**
 - Dockerfile 
`FROM registry.gitlab.com/gitlab-org/gitlab-ee/gitlab-assets-ee:11-5-stable-ee as assets`

priyawadhwa:
Hey @mtsyganov what version of the kaniko image are you using? I tried with `gcr.io/kaniko-project/executor:v0.6.0` and `gcr.io/kaniko-project/debug-v0.6.0` and they both build the image successfully.

mtsyganov:
@priyawadhwa you are right. 
I am sorry, I misunderstood the bug. The real problem is that kaniko dont handle correct the `"` in the FROM command when I use variables. 

following Dockerfile will fail
```
ARG ASSETS_IMAGE="registry.gitlab.com/gitlab-org/gitlab-ee/gitlab-assets-ee:11-5-stable-ee"
FROM ${ASSETS_IMAGE} as assets
```

but this Dockerfile will succeed
```
ARG ASSETS_IMAGE=registry.gitlab.com/gitlab-org/gitlab-ee/gitlab-assets-ee:11-5-stable-ee
FROM ${ASSETS_IMAGE} as assets
```

In `docker build` both cases works fine

PR #834, opened 2019-10-24
Issue #439 Strip out double quotes in ARG value

cvgw:
Fixes [#439](https://github.com/GoogleContainerTools/kaniko/issues/439)

**Description**

Strip out double quotes enclosing ARG value after parsing Dockerfile. This brings Kaniko into parity with Docker for handling ARG values. This prevents failures where things like image names would not parse correctly due to the double quotes remaining.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Strip double quotes enclosing the value of ARG commands before substituting those values in during shell expansion
```

### DIFF
--- a/pkg/dockerfile/dockerfile.go
+++ b/pkg/dockerfile/dockerfile.go
@@ -111,7 +111,31 @@ func Parse(b []byte) ([]instructions.Stage, []instructions.ArgCommand, error) {
 	if err != nil {
 		return nil, nil, err
 	}
+
+	metaArgs = stripEnclosingDoubleQuotes(metaArgs)
+
 	return stages, metaArgs, nil
+}
+
+// stripEnclosingDoubleQuotes removes double quotes enclosing the value of each instructions.ArgCommand in a slice
+func stripEnclosingDoubleQuotes(metaArgs []instructions.ArgCommand) []instructions.ArgCommand {
+	for i := range metaArgs {
+		arg := metaArgs[i]
+		v := arg.Value
+		if v != nil {
+			val := *v
+			if val[0] == '"' {
+				val = val[1:]
+			}
+
+			if val[len(val)-1] == '"' {
+				val = val[:len(val)-1]
+			}
+			arg.Value = &val
+			metaArgs[i] = arg
+		}
+	}
+	return metaArgs
 }
 
 // targetStage returns the index of the target stage kaniko is trying to build
--- a/pkg/dockerfile/dockerfile.go
+++ b/pkg/dockerfile/dockerfile.go
@@ -112,30 +112,41 @@ func Parse(b []byte) ([]instructions.Stage, []instructions.ArgCommand, error) {
 		return nil, nil, err
 	}
 
-	metaArgs = stripEnclosingDoubleQuotes(metaArgs)
+	metaArgs, err = stripEnclosingQuotes(metaArgs)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	return stages, metaArgs, nil
 }
 
-// stripEnclosingDoubleQuotes removes double quotes enclosing the value of each instructions.ArgCommand in a slice
-func stripEnclosingDoubleQuotes(metaArgs []instructions.ArgCommand) []instructions.ArgCommand {
+// stripEnclosingQuotes removes quotes enclosing the value of each instructions.ArgCommand in a slice
+// if the quotes are escaped it leaves them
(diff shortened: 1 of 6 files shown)


pkg/dockerfile/dockerfile.go
```diff
@@ -111,9 +111,33 @@ func Parse(b []byte) ([]instructions.Stage, []instructions.ArgCommand, error) {
...
 		return nil, nil, err
 	}
+
+	metaArgs = stripEnclosingDoubleQuotes(metaArgs)
+
 	return stages, metaArgs, nil
 }
 
+// stripEnclosingDoubleQuotes removes double quotes enclosing the value of each instructions.ArgCommand in a slice
+func stripEnclosingDoubleQuotes(metaArgs []instructions.ArgCommand) []instructions.ArgCommand {
```
tejal29: We can add table driven unit tests here
1. value like`"foo"`
2. For empty value `""` ? What happens then?
   maybe we shd keep quotes as is then?
3.  value `foo`
4.  invalid value `"foo` shd probably error out?

tejal29: Some other test cases to add would be
5. When `metaArgs` is empty
6. Multiple `mataArgs` instructions have `""`. (All get resolved)

priyawadhwa: just something to consider, something like `\"foo\"` would probably need to maintain the quotes

cvgw:
> I didn't look thoroughly at this, but this may be what docker uses to [resolve quotes](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/shell/lex.go#L13-L19), which we could potentially use as well to ensure we respond to this as docker does

@priyawadhwa kaniko is currently [using that same package](https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/util/command_util.go#L60) to do the variable expansion. I was surprised that it doesn't seem to be handling the quotes, but I didn't see anything in the documentation referring to that. I did notice that in a recent version (later than what kaniko seems to be using) that `RawQuotes` field was added. Perhaps that is related?

pkg/dockerfile/dockerfile.go
cvgw: > We can add table driven unit tests here for this function only
> 
> 1. value like`"foo"`
> 2. For empty value `""` ? What happens then?
>    maybe we shd keep quotes as is then?
> 3. value `foo`
> 4. invalid value `"foo` shd probably error out?

2. Docker seems to strip out the quotes and treat it as a valid, but blank string
4. Yup, Docker also seems to error when either quote isn't matched

We also might want to handle single quotes as docker seems to handle those as well

cvgw: > just something to consider, something like `\"foo\"` would probably need to maintain the quotes

Ya, it seems that Docker does not remove the quotes in that case which will result in an error for things like image name

priyawadhwa:
>  I did notice that in a recent version (later than what kaniko seems to be using) that RawQuotes field was added. Perhaps that is related?

Yah that sounds like it could be! We could definitely bump up the version if you think that would work

cvgw:
> > I did notice that in a recent version (later than what kaniko seems to be using) that RawQuotes field was added. Perhaps that is related?
> 
> Yah that sounds like it could be! We could definitely bump up the version if you think that would work

Unfortunately doesn't seem to be related
```
package buildkit_quotes_test
  
  import (
          "testing"
  
          "github.com/moby/buildkit/frontend/dockerfile/parser"
          "github.com/moby/buildkit/frontend/dockerfile/shell"
  )
  
  func TestWithDoubleQuotes(t *testing.T) {
          envs := []string{"MEOW=purr", "VAL=\"World\""}
  
          lex := shell.NewLex(parser.DefaultEscapeToken)
          val, err := lex.ProcessWord("Hello ${VAL}", envs)
          if err != nil {
                  t.Error(err)
          }   
  
          if val != "Hello World" {
                  t.Errorf("Expected Hello World but got %s", val)
          }   
  }
  
  func TestWithDoubleQuotesRawQuotesTrue(t *testing.T) {
          envs := []string{"MEOW=purr", "VAL=\"World\""}
  
          lex := shell.NewLex(parser.DefaultEscapeToken)
          lex.RawQuotes = true
  
          val, err := lex.ProcessWord("Hello ${VAL}", envs)
          if err != nil {
                  t.Error(err)
          }   
  
          if val != "Hello World" {
                  t.Errorf("Expected Hello World but got %s", val)
          }   
  }
```

PR #841, opened 2019-10-29
Added --image-name-with-digest flag

### LINKED ISSUE #814: Add --image-name-with-digest-file
I request a small new feature very similar to the existing `--digest-file` option.
The `--digest-file` option only outputs the image digest (`sha256:bc45...67e1`) and needs to be combined with image name to make a working image reference.

The `--image-name-with-digest-file` should output the full image name including the digest: `my-org/my-image@sha256:bc45...67e1`

wripley:
Resolves #814.

**Description**
Added --image-name-with-digest to write image@sha256:... names based on destinations specified.

New flag will write a line for each destination specified. This is the full image path (ie, a docker registry will start with index.docker.io/{USER}/{IMAGE}@sha256:__)

**Submitter Checklist**

review them:

- [x] Includes unit tests

**Release Notes**

Added --image-name-with-digest flag
Description of flag: Specify a file to save the image name w/ digest of the built image to.


pkg/executor/push.go
```diff
@@ -128,9 +140,21 @@ func DoPush(image v1.Image, opts *config.KanikoOptions) error {
 		if err != nil {
 			return errors.Wrap(err, "getting tag for destination")
 		}
+		if opts.ImageNameDigestFile != "" {
```
tejal29: this implies `--digest-file` required for `--image-name-with-digest-file` ?
it should be fine to just use `--image-name-with-digest-file`

pkg/executor/push.go
```diff
@@ -128,9 +140,21 @@ func DoPush(image v1.Image, opts *config.KanikoOptions) error {
...
+			builder.Write(append(imageName, digestByteArray...))
+			builder.WriteString("\n")
+		}
 		destRefs = append(destRefs, destRef)
 	}
 
+	if opts.ImageNameDigestFile != "" {
+		err := ioutil.WriteFile(opts.ImageNameDigestFile, []byte(builder.String()), 0644)
+		if err != nil {
+			return errors.Wrap(err, "writing digest to file failed")
```
tejal29:
```suggestion
			return errors.Wrap(err, "writing image name with digest to file failed")
```

pkg/executor/push.go
wripley: Not sure I am following why line 143 implies --digest-file is required. The intent was to build up the list of images for each destination specified. Could you elaborate?

Issue #682, opened 2019-05-30
Caching fails in multi-stage build


apenney:
**Actual behavior**
When I build a multi-stage docker file the cache is blown every time, no matter what I do.  The dockerfile (first two stages, where I see the problem, there are more stages):

```
FROM python:2.7-slim AS base

# Make man page directories so that postgresql-client can install
RUN seq 1 8 | xargs -I{} mkdir -p /usr/share/man/man{} \
    && apt-get update \
    && apt-get install -y curl \
    && curl -sL https://deb.nodesource.com/setup_10.x | bash - \
    && apt-get install -y build-essential \
                          checkinstall \
                          default-libmysqlclient-dev \
                          libbz2-dev \
                          libffi-dev \
                          libgflags-dev \
                          libgflags2v5 \
                          liblz4-dev \
                          libncurses5-dev \
                          libpq-dev \
                          libsnappy-dev \
                          libsnappy1v5 \
                          libxml2 \
                          libxml2-dev \
                          libxmlsec1-dev \
                          libxslt1.1 \
                          libzstd-dev \
                          libzstd1 \
                          mariadb-client \
                          mysql-client \
                          nodejs \
                          postgresql-client-9.6 \
                          unixodbc \
                          unixodbc-dev \
                          wget \
                          zlib1g-dev \
    && rm -rf /var/lib/apt/lists/*

##
## RocksDB base image
##
## We don't expect this to change frequently, so we make it a stage all by
## itself so that we never rebuild it unless we change the version of rocksdb.
##
FROM base AS rocksdb-build

RUN curl -L https://github.com/facebook/rocksdb/archive/v5.18.3.tar.gz \
  | tar -xzC /tmp \
  && PORTABLE=1 make -C /tmp/rocksdb-5.18.3 -j 4 install-shared INSTALL_PATH=/install
```

Each build will use the cache for the first RUN, but always result in looking for a different cache for the second RUN, no matter what I do.

I run with:

```
/kaniko/executor --build-arg tests=false --dockerfile=`pwd`/Dockerfile -c `pwd` --cache=true --destination=x.dkr.ecr.us-east-1.amazonaws.com/web:latest --cache-repo=x.dkr.ecr.us-east-1.amazonaws.com/build_cache
```

Here's a snippet from a run (sorry, messy logs as I cut and paste from jenkins):

```
[36mINFO[0m[0000] Resolved base name python:2.7-slim to python:2.7-slim 
[...]

PR #882, opened 2019-11-25
Do not recompute layers retrieved from cache

cvgw:
**Description**
Updates kaniko to reuse cached layers rather than recreating the layer.

Previously kaniko would retrieve a layer from the cache, unpack the tar, snapshot the file system, and then repack the tar, recreate the layer. This caused unneeded work repacking the ta. It also resulted in layer and image digests that would change every build due to the tar being repacked.

This change updates kaniko so that the new behavior is
* Retrieve layer from cache
* Unpack tar
* Snapshot filesystem
* Reuse layer retrieved from cache

Note that for the moment it is only implemented for RUN and COPY commands.

This eliminates the extra steps of repacking the tar. It also results in layer and image digest which do not change for cached layers.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**
```
- Improve performance when caching is enabled
- Improve caching behavior for FROM commands which reference a previous stage
```

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -326,29 +326,47 @@ func (s *stageBuilder) build() error {
 			continue
 		}
 
-		tarPath, err := s.takeSnapshot(files)
-		if err != nil {
-			return errors.Wrap(err, "failed to take snapshot")
+		fn := func() bool {
+			switch v := command.(type) {
+			case commands.Cached:
+				return v.ReadSuccess()
+			default:
+				return false
+			}
 		}
 
-		logrus.Debugf("build: composite key for command %v %v", command.String(), compositeKey)
-		ck, err := compositeKey.Hash()
-		if err != nil {
-			return errors.Wrap(err, "failed to hash composite key")
-		}
+		if fn() {
+			v := command.(commands.Cached)
+			layer := v.Layer()
+			if err := s.saveLayerToImage(layer, command.String()); err != nil {
+				return err
+			}
+		} else {
+			tarPath, err := s.takeSnapshot(files)
+			if err != nil {
+				return errors.Wrap(err, "failed to take snapshot")
+			}
 
-		logrus.Debugf("build: cache key for command %v %v", command.String(), ck)
+			logrus.Debugf("build: composite key for command %v %v", command.String(), compositeKey)
+			ck, err := compositeKey.Hash()
+			if err != nil {
+				return errors.Wrap(err, "failed to hash composite key")
+			}
 
-		// Push layer to cache (in parallel) now along with new config file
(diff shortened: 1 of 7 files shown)


pkg/executor/build.go
```diff
@@ -261,23 +261,48 @@ func (s *stageBuilder) build() error {
 			continue
 		}
 
-		tarPath, err := s.takeSnapshot(files)
-		if err != nil {
-			return err
+		fn := func() bool {
+			switch v := command.(type) {
+			case commands.CachedCommand:
+				if !v.ReadSuccess() {
```
tejal29: return v.ReadSuccess()

cvgw:
This actually will fix the specific case in #682 but not the underlying issue. https://github.com/GoogleContainerTools/kaniko/pull/891 should be a more complete fix.

This change may still be worth doing to remove the extra work of recreating the tar.

tejal29:
@cvgw Would be able to pick this up for release tomorrow? if not i will rebase and merge.

cvgw:
@tejal29 do we still want to make this change? I'm not sure what the motivation is now that we've made all the caching fixes. I suppose there could be a performance win here. 

WDYT?

tejal29:
Since you have spent cycles on it, we can get this in. Upto you!

cvgw:
This has a LOT of conflicts with master and some substantial overlap with https://github.com/GoogleContainerTools/kaniko/pull/973

I'm going to wait until #973 is merged and then cherry-pick the changes over

raijinsetsu:
I believe this PR and/or #715 are contributing to significant time in our builds. Is there an ETA when these changes may be available?
Note: #973 has been merged as of 2 weeks ago.

tejal29:
We should look into how we can measure performance increase due to this PR. 
One think we can do is, 
1. Monitor build time decrease for GCB (internally)

raijinsetsu:
@tejal29 
Based on the time I am seeing in my build spent re-archiving and sending the archive to Cloud Storage, my build time would be reduced by about 10 minutes in the "cache hit" scenario, or 33%. If you think it would help, I could try to build and use master to give you actual numbers.

Note: early in my Dockerfile is a "RUN npx lerna bootstrap --ci" which installs all the dependencies for the mono-repository. Installing the dependencies takes about 10 minutes, which includes building 15+ local modules. The subsequent archiving takes about 10 minutes. As the dependencies do not change often, having the cache work correctly would trim about 20 minutes from most builds.

raijinsetsu:
Just for giggles, I used the "gcr.io/kaniko-project/executor:debug-8d9e6b8ea54274f73517f11c113c13cd03d26349" image which appears to have this change.
My build went from 20 minutes to 2 minutes. 👍

Issue #870, opened 2019-11-20
False-positive cache-hit on COPY --from


haizaar:
**Actual behavior**
I have a two stage docker image that first build sphinx docs and then copies it to the second nginx image:
```dockerfile
FROM python:3.7.5-alpine3.10 AS builder
RUN apk add --no-cache build-base libxml2-dev libxslt-dev graphviz
COPY requirements.txt /
RUN pip install --upgrade -r /requirements.txt
WORKDIR /build
COPY . ./
RUN make html
########################
FROM nginx:1.17.5-alpine AS runner
COPY --from=builder /build/_build/html /usr/share/nginx/html
```

I build it on Google Cloud Build using `gcr.io/kaniko-project/executor:latest` builder. 

The problem is the last copy instruction - even through generated HTML files differ from the previously built revision, kaniko still prefers to use the cached version:
```
Step #0: build succeeded.
Step #0:
Step #0: The HTML pages are in _build/html.
Step #0:
Step #0: Build finished. The HTML pages are in _build/html.
Step #0: INFO[0062] Taking snapshot of full filesystem...
Step #0: INFO[0064] Pushing layer gcr.io/xxx/docs-server/cache:cd602b7af9c972414b3241eb3a192a606755174cf665f9806155078915975932 to cache now
Step #0: INFO[0068] Saving file /build/_build/html for later use.
Step #0: INFO[0068] Deleting filesystem...
Step #0: INFO[0068] Downloading base image nginx:1.17.5-alpine
Step #0: INFO[0069] Error while retrieving image from cache: getting file info: stat /cache/sha256:b4c0378c841cd76f0b75bc63454bfc6fe194a5220d4eab0d75963bccdbc327ff: no such file or directory
Step #0: INFO[0069] Downloading base image nginx:1.17.5-alpine
Step #0: INFO[0069] Checking for cached layer gcr.io/xxx/docs-server/cache:5de22bf6a7763f10cd0ffa0592b8f51c876286a1d0f50564cf2f21bfd6262dc4...
Step #0: INFO[0069] Using caching version of cmd: COPY --from=builder /build/_build/html /usr/share/nginx/html
Step #0: INFO[0069] Skipping unpacking as no commands require it.
Step #0: INFO[0069] Taking snapshot of full filesystem...
Step #0: INFO[0070] COPY --from=builder /build/_build/html /usr/share/nginx/html
Step #0: INFO[0070] Found cached layer, extracting to filesystem
```

[...]

PR #883, opened 2019-11-25
Include source stage cache key in cache key for COPY commands using --from

cvgw:
Address one part of the issue in #870

**Description**

Include the source stage cache key in the cache key for COPY commands which use --from. This will insure when the source stage for the COPY command changes the cache for said COPY command is invalidated.

```
FROM a as a
RUN break-the-cache
FROM b as b
RUN something  <-- this will still be cached
COPY --from a a/file   <--here the cache is disabled
```

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Improve cache invalidation for COPY commands using the --from option
```

### DIFF
--- a/pkg/commands/copy.go
+++ b/pkg/commands/copy.go
@@ -161,6 +161,10 @@ func (c *CopyCommand) CacheCommand(img v1.Image) DockerCommand {
 	}
 }
 
+func (c *CopyCommand) From() string {
+	return c.cmd.From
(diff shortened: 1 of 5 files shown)


cvgw:
This PR also requires https://github.com/GoogleContainerTools/kaniko/issues/899 to be fixed in order to completely address the issue

PR #888, opened 2019-11-27
add unit tests for caching run and copy

cvgw:
**Description**

Adds unit tests for caching copy and run

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

### DIFF
--- a/pkg/commands/copy.go
+++ b/pkg/commands/copy.go
@@ -17,6 +17,7 @@ limitations under the License.
 package commands
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
@@ -156,8 +157,9 @@ func (c *CopyCommand) ShouldCacheOutput() bool {
 func (c *CopyCommand) CacheCommand(img v1.Image) DockerCommand {
 
 	return &CachingCopyCommand{
-		img: img,
-		cmd: c.cmd,
+		img:       img,
+		cmd:       c.cmd,
+		extractFn: util.ExtractFile,
 	}
 }
 
@@ -170,16 +172,23 @@ type CachingCopyCommand struct {
 	img            v1.Image
 	extractedFiles []string
 	cmd            *instructions.CopyCommand
+	extractFn      util.ExtractFunction
 }
 
 func (cr *CachingCopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.BuildArgs) error {
 	logrus.Infof("Found cached layer, extracting to filesystem")
 	var err error
-	cr.extractedFiles, err = util.GetFSFromImage(RootDir, cr.img)
+
+	if cr.img == nil {
+		return errors.New(fmt.Sprintf("cached command image is nil %v", cr.String()))
+	}
+	cr.extractedFiles, err = util.GetFSFromImage(RootDir, cr.img, cr.extractFn)
+
 	logrus.Infof("extractedFiles: %s", cr.extractedFiles)
 	if err != nil {
 		return errors.Wrap(err, "extracting fs from image")
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


cvgw:
I think I may have found a bug, which is that CachingCopy and Copy return different values for `FilesUsedFromContext`. The result of `FilesUsedFromContext` is included in the cache key which results in CachingCopy and Copy generating different cache keys. This also seem to be true of CachingRun and Run.

https://github.com/GoogleContainerTools/kaniko/pull/888/files#diff-44df3b0bdda7ba139f5aeadce2431721R301

Here it will be the non caching version of the command https://github.com/GoogleContainerTools/kaniko/blob/bfd85628559f0dae5b718754e1aab2f6773623a2/pkg/executor/build.go#L152

Here it will be the caching version of the command
https://github.com/GoogleContainerTools/kaniko/blob/bfd85628559f0dae5b718754e1aab2f6773623a2/pkg/executor/build.go#L243

cvgw:
> I think I may have found a bug, which is that CachingCopy and Copy return different values for `FilesUsedFromContext`. The result of `FilesUsedFromContext` is included in the cache key which results in CachingCopy and Copy generating different cache keys. This also seem to be true of CachingRun and Run.
> 
> https://github.com/GoogleContainerTools/kaniko/pull/888/files#diff-44df3b0bdda7ba139f5aeadce2431721R301
> 
> Here it will be the non caching version of the command
> 
> https://github.com/GoogleContainerTools/kaniko/blob/bfd85628559f0dae5b718754e1aab2f6773623a2/pkg/executor/build.go#L152
> 
> Here it will be the caching version of the command
> https://github.com/GoogleContainerTools/kaniko/blob/bfd85628559f0dae5b718754e1aab2f6773623a2/pkg/executor/build.go#L243

After thinking about this more; I don't think it is technically a bug since CachingCopy and CachingRun commands are pushed to the cache (since they are already there) so the cache keys produced for the caching and non-caching version may have never been intended to be the same. Where this becomes an issue are possible changes where we start using the cache key of layer A in the cache key of layer B (IE COPY --from)

cvgw:
> > I think I may have found a bug, which is that CachingCopy and Copy return different values for `FilesUsedFromContext`. The result of `FilesUsedFromContext` is included in the cache key which results in CachingCopy and Copy generating different cache keys. This also seem to be true of CachingRun and Run.
> > https://github.com/GoogleContainerTools/kaniko/pull/888/files#diff-44df3b0bdda7ba139f5aeadce2431721R301
> > Here it will be the non caching version of the command
> > https://github.com/GoogleContainerTools/kaniko/blob/bfd85628559f0dae5b718754e1aab2f6773623a2/pkg/executor/build.go#L152
> > 
> > Here it will be the caching version of the command
> > https://github.com/GoogleContainerTools/kaniko/blob/bfd85628559f0dae5b718754e1aab2f6773623a2/pkg/executor/build.go#L243
> 
> After thinking about this more; I don't think it is technically a bug since CachingCopy and CachingRun commands are not pushed to the cache (since they are already there) so the cache keys produced for the caching and non-caching version may have never been intended to be the same. Where this becomes an issue are possible changes where we start using the cache key of layer A in the cache key of layer B (IE COPY --from)
> 
> So probably not a bug, but perhaps unexpected.

So this actually is a bug in the case of caching copy commands.

CopyCommand and CachingCopyCommand produce a different cachekey. This means that the cache key used to look up the cached layer (in `stagebuilder.optimize`) is different than the cache key used to write the layer to the cache (in `stagebuilder.build`).

The bug becomes apparent when you have a dockerfile that contains a cached copy command followed by other commands like RUN. This is due to the fact that we continuously append the composite cache with new keys. This means that after the cached copy command, optimize and build produce different cache keys. Example below

```
FROM foo
COPY . ./ 
RUN apt-get install -y bar
```
```
optimize ->
COPY . ./  <- is CopyCommand, cache key includes FilesUsedFromContext
RUN apt-get install bar <- cache key includes COPY . ./ FilesUsedFromContext

build ->
COPY . ./ is CachingCopyCommand, cache key does not include FilesUsedFromContext
RUN apt-get install bar <- cache key does not include COPY . ./ FilesUsedFromContext
```

This results in `RUN apt-get install bar` producing a different cache key for read (in optimize) vs write (in build). It never finds the cache key it wrote.

PR #898, opened 2019-12-02
Only build required docker images for integration tests

### LINKED ISSUE #897: Integration suite should build images only for tests that will actually be run
The way we currently setup the integration suite means that we build every docker image before a single test is run. This means that when the user specifies a specific pattern for tests like `-run TestLayers/test_layer_Dockerfile_test_copy_bucket` a lot of unnecessary images are built.

The integration test suite should be updated so that only the images explicitly required for the tests which will execute are built.

cvgw:
Fixes #897 
**Description**

Only build docker images for the integration tests that actually run rather than for all integration tests regardless of which will run.

This will greatly reduce the time it takes to run an individual integration test.

Looking at the kokoro history, this change doesn't seem to impact performance when running the entire test suite.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**
```
- Only build docker images for the integration tests being run
````

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -140,17 +140,6 @@ func TestMain(m *testing.M) {
 	}
 	imageBuilder = NewDockerFileBuilder(dockerfiles)
 
-	//g := errgroup.Group{}
-	//for dockerfile := range imageBuilder.FilesBuilt {
-	//        df := dockerfile
-	//        g.Go(func() error {
-	//                return imageBuilder.BuildImage(config.imageRepo, config.gcsBucket, dockerfilesPath, df)
-	//        })
-	//}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: integration/integration_test.go [hunk 1]
```diff
@@ -188,19 +140,20 @@ func TestMain(m *testing.M) {
...
-		df := dockerfile
-		g.Go(func() error {
-			return imageBuilder.BuildImage(config.imageRepo, config.gcsBucket, dockerfilesPath, df)
-		})
-	}
-	if err := g.Wait(); err != nil {
-		fmt.Printf("Error building images: %s", err)
-		os.Exit(1)
-	}
+	//g := errgroup.Group{}
```
tejal29: please uncomment or remove this.

PR #906, opened 2019-12-09
updated readme

Pweetoo:
Added missing argument.

**Description**

Added missing argument -n for echo command, which corrects wrong base64 string being generated for config.json file. That resulted in authentication error when pushing image to dockerhub.


Pweetoo:
@googlebot I signed it!

PR #914, opened 2019-12-15
Fix #899 cached copy results in inconsistent key

### LINKED ISSUE #899: Cache keys don't match for builds with a cached copy command
**Actual behavior**
When a build contains a cached layer using the `COPY` command all following layers in the build will produce incorrect cache keys. Specifically, the cache key used to read from the cache and the cache key used to write to the cache will be different for each layer following the copy command. 

Layers affected by this will never be able to read from the cache because the key used to write the layer to the cache is different than the key used to read from the cache.

The cause of this bug is due to CachingCopyCommand and CopyCommand returning different values for `FilesUsedFromContext`. The result of `FilesUsedFromContext` is added to the cache key.

**Expected behavior**
All commands always produce the same cache key during read and during write given the same input.

**To Reproduce**
1. Any build with a cached copy command followed by any other caching command
2. Observe that commands following the cached copy command are never cached

**Additional Information**
 - Dockerfile 
```
FROM ubuntu:16.04
COPY foo.txt bar.txt
RUN apt-get update -y
```
 - Build Context
   `echo 'meow' > foo.txt`
 
cvgw:
Fixes #899 

**Description**

* Update cached copy command to return the same result for
files used from context so that cached and uncached copy
commands produce the same cache key
* Update tests for fix
* Add test for cached run command key consistency

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
Fix caching issue for builds with a cached copy command
```

### DIFF
--- a/pkg/commands/copy.go
+++ b/pkg/commands/copy.go
@@ -121,24 +121,7 @@ func (c *CopyCommand) String() string {
 }
 
 func (c *CopyCommand) FilesUsedFromContext(config *v1.Config, buildArgs *dockerfile.BuildArgs) ([]string, error) {
-	// We don't use the context if we're performing a copy --from.
-	if c.cmd.From != "" {
-		return nil, nil
-	}
-
-	replacementEnvs := buildArgs.ReplacementEnvs(config.Env)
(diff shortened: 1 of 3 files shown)


cvgw:
After some testing it looks like this will fix #870

PR #943, opened 2020-01-03
Fix #942 COPY or ADD to symlink destination breaks image

### LINKED ISSUE #942: ADD or COPY with symbolic link breaks image
**Actual behavior**
Image is broken when there is a symbolic link at the destination when ADD or COPY

**Expected behavior**
build should succeed and sources should be in destination

**To Reproduce**
1.  create Dockerfile
```dockerfile
FROM ubuntu
RUN mkdir -p /root/real && ln -sf /root/real /root/symlink
## cases
# case1: build success, but file path is wrong:  `/root/root/real/test-file`
COPY test-file /root/symlink/test-file 
# case2: image broken
COPY test-file /root/symlink/inner/test-file 
# case3: build fail
COPY test-dir/ /root/symlink 
# case4: image broken 
COPY test-dir/ /root/symlink/test-dir/ 
# case5: image broken
COPY test-dir/ /root/symlink/inner/test-dir/ 
```

2. image build with kaniko
```bash
$ mkdir test-dir && echo "test" > test-file && cp test-file test-dir/test-dir-file
$ /kaniko/executor --destination=shortstories/kaniko-test:case1 -v=trace
```

3. docker run
```bash
$ docker run --rm -it --entrypoint='/bin/bash' shortstories/kaniko-test:case4
Unable to find image 'shortstories/kaniko-test:case4' locally
test: Pulling from shortstories/kaniko-test
2746a4a261c9: Pull complete
4c1d20cdee96: Pull complete
0d3160e1d0de: Pull complete
c8e37668deea: Pull complete
ad9c552434bd: Pull complete
780db71f2ce6: Extracting [==================================================>]     257B/257B
docker: failed to register layer: Error processing tar file(exit status 1): mkdir /root/symlink: file exists.
See 'docker run --help'.
```

- test images
     - `shortstories/kaniko-test:case1`
     - `shortstories/kaniko-test:case2`
     - `shortstories/kaniko-test:case4`
     - `shortstories/kaniko-test:case5`

**Additional Information**
 - Dockerfile 
     - See `To Reproduce`
 - Build Context
     - See `To Reproduce`
     - `gcr.io/kaniko-project/executor:v0.15.0` (digest: `sha256:630f263d9123266b9f5420d9bc130e6a79306dbc312e5a7d35d922df391192bb`)

shortstories:
Fixes #942 

**Description**

Fixes an issue where the image is broken if there are symlink in the destination path when ADD or COPY

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Fixes an issue where symlinks resolved incorrectly
```

### DIFF
--- a/pkg/commands/copy.go
+++ b/pkg/commands/copy.go
@@ -18,14 +18,13 @@ package commands
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
-
+	"github.com/GoogleContainerTools/kaniko/pkg/constants"
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-
-	"github.com/GoogleContainerTools/kaniko/pkg/constants"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/GoogleContainerTools/kaniko/pkg/dockerfile"
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
@@ -63,6 +62,7 @@ func (c *CopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
 		if cwd == "" {
 			cwd = constants.RootDir
 		}
+
 		destPath, err := util.DestinationFilepath(src, dest, cwd)
 		if err != nil {
 			return err
@@ -76,11 +76,7 @@ func (c *CopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
 		}
 
 		if fi.IsDir() {
-			if !filepath.IsAbs(dest) {
-				// we need to add '/' to the end to indicate the destination is a directory
-				dest = filepath.Join(cwd, dest) + "/"
-			}
-			copiedFiles, err := util.CopyDir(fullPath, dest, c.buildcontext)
+			copiedFiles, err := util.CopyDir(fullPath, destPath, c.buildcontext)
 			if err != nil {
 				return err
 			}
@@ -197,21 +193,43 @@ func (cr *CachingCopyCommand) From() string {
 }
 
 func resolveIfSymlink(destPath string) (string, error) {
-	baseDir := filepath.Dir(destPath)
-	if info, err := os.Lstat(baseDir); err == nil {
+	if !filepath.IsAbs(destPath) {
+		return "", errors.New("dest path must be abs")
+	}
+
+	pathPrefix := "/"
+	pathTokens := strings.Split(destPath, string(os.PathSeparator))
+	for i, pathToken := range pathTokens {
+		currentPath := filepath.Join(pathPrefix, pathToken)
+		info, err := os.Lstat(currentPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				pathPrefix = filepath.Join(append([]string{currentPath}, pathTokens[i+1:]...)...)
+				break
+			} else {
+				return "", errors.Wrap(err, "failed to lstat")
+			}
+		}
+
 		switch mode := info.Mode(); {
 		case mode&os.ModeSymlink != 0:
-			linkPath, err := os.Readlink(baseDir)
+			linkPath, err := os.Readlink(currentPath)
 			if err != nil {
-				return "", errors.Wrap(err, "error reading symlink")
+				return "", errors.Wrap(err, "failed to read symlink")
 			}
-			absLinkPath := filepath.Join(filepath.Dir(baseDir), linkPath)
-			newPath := filepath.Join(absLinkPath, filepath.Base(destPath))
-			logrus.Tracef("Updating destination path from %v to %v due to symlink", destPath, newPath)
-			return newPath, nil
+			if filepath.IsAbs(linkPath) {
+				pathPrefix = linkPath
+			} else {
+				pathPrefix = filepath.Join(pathPrefix, linkPath)
+			}
+		default:
+			pathPrefix = filepath.Join(pathPrefix, pathToken)
 		}
 	}
-	return destPath, nil
+	if destPath != pathPrefix {
+		logrus.Tracef("Updating destination path from %v to %v due to symlink", destPath, pathPrefix)
+	}
+	return filepath.Clean(pathPrefix), nil
 }
 
 func copyCmdFilesUsedFromContext(
--- a/pkg/commands/copy.go
+++ b/pkg/commands/copy.go
@@ -18,13 +18,14 @@ package commands
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/GoogleContainerTools/kaniko/pkg/constants"
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


shortstories:
@googlebot I signed it!

pkg/commands/copy.go
```diff
@@ -197,21 +197,43 @@ func (cr *CachingCopyCommand) From() string {
 }
 
 func resolveIfSymlink(destPath string) (string, error) {
-	baseDir := filepath.Dir(destPath)
-	if info, err := os.Lstat(baseDir); err == nil {
+	if !filepath.IsAbs(destPath) {
```
cvgw: Why is this check needed?

pkg/commands/copy.go
```diff
@@ -197,21 +197,43 @@ func (cr *CachingCopyCommand) From() string {
...
+		return "", errors.New("dest path must be abs")
+	}
+
+	pathPrefix := "/"
+	pathTokens := strings.Split(destPath, string(os.PathSeparator))
+	for i, pathToken := range pathTokens {
```
cvgw: Can you explain the idea behind this code. It's not clear what it accomplishes.

pkg/commands/copy.go
```diff
@@ -59,11 +59,15 @@ func (c *CopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
 		if err != nil {
 			return err
 		}
+		if fi.IsDir() && !strings.HasSuffix(fullPath, string(os.PathSeparator)) {
```
cvgw: This should probably use `filepath.IsAbs(fullPath)`

pkg/commands/copy.go
```diff
@@ -59,11 +59,15 @@ func (c *CopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
...
+			fullPath += "/"
```
cvgw: fullPath = filepath.Clean(fullPath)

pkg/commands/copy.go
```diff
@@ -59,11 +59,15 @@ func (c *CopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
...
+		}
 		cwd := config.WorkingDir
 		if cwd == "" {
 			cwd = constants.RootDir
 		}
-		destPath, err := util.DestinationFilepath(src, dest, cwd)
+
+		destPath, err := util.DestinationFilepath(fullPath, dest, cwd)
```
cvgw: Why `fullPath` instead of `src`?

pkg/commands/copy.go
shortstories: It is suffix, not prefix. I put this code in here to easily check if source is a directory.

pkg/commands/copy.go
shortstories: `filepath.Clean()` removes trailing slash.

pkg/commands/copy.go
shortstories: To check if source is a directory. Would it be right to call `IsDestDir(src)` once again inside `util.DestinationFilePath()`?

pkg/commands/copy.go
shortstories: If it is rel path, it is not return value of `util.DestinationFilepath()`. Would it be correct to append the working directory to destPath in this function?

pkg/commands/copy.go
shortstories: To resolve all symlinks in destination path.

The previous code failed to resolve unless symlink was a parent path. For example, when the destination is `/root/symlink/inner/target`, we expect it to be resolved to`/root/real/inner/target`, but previous code didn't work that way, it causing problems.
Similarly, if more than one symlink was found in the destination, it did not resolved correctly.
Problems also occur when symlink points to abs path.

So I split path into path elements and then re-joined them one by one. if symlinks is found in this process, it is replaced with a pointed path.
Then, if a path that does not exist, all remain path elements are need to be created, so they are just joined and returned.
This way, no matter how many symlinks are in the destination, no problem.

cvgw: Ah ok, thanks for that explanation. I think I better understand the problem now.

Would https://godoc.org/path/filepath#EvalSymlinks accomplish that goal?

pkg/commands/copy.go
cvgw: woops, yup I see that mistake now. I guess I'm slightly confused about this check. So it is `true` for a fi that is a directory but whose `fullPath` does not end in a `/` ?

Is the goal of this check to make sure `fullPath` is formatted as expected?

pkg/commands/copy.go
cvgw: :+1: my mistake

pkg/commands/copy.go
shortstories: Yeah. At first I tried to solve this problem with `filepath.EvalSymlinks`. However, `file path.EvalSymlinks` could not be used because it returns a zero value string and error when` destPath` does not exist. We need to create a new directory even if `destPath` does not exist.

pkg/commands/copy.go
shortstories: The goal of this code is what you say. I thought it would be inefficient to call `s.Lstat()` or` os.Stat()` every time when I pass src directly to a function.

pkg/commands/copy.go
cvgw: Ah ok, that makes sense.

Would it be possible to use `EvalSymlink` by resolving the path from the longest to shortest segment?

Something like
```
for path != "/" {
  info, err := os.Lstat(path)
  if err != nil {
    if os.IsNotExist(err) {
      path = filepath.Dir(path)
      continue
    }
    // handle fatal err
  }
  path = path.EvalSymlinks(path)
}
```

WDYT?

shortstories: LGTM.

like this?
```go
	var nonexistentPaths []string

	newPath := destPath
	for newPath != "/" {
		_, err := os.Lstat(newPath)
		if err != nil {
			if os.IsNotExist(err) {
				dir, file := filepath.Split(newPath)
				newPath = filepath.Clean(dir)
				nonexistentPaths = append(nonexistentPaths, file)
				continue
			} else {
				return "", errors.Wrap(err, "failed to lstat")
			}
		}

		newPath, err = filepath.EvalSymlinks(newPath)
		if err != nil {
			return "", errors.Wrap(err, "failed to eval symlinks")
		}
		break
	}

	for i := len(nonexistentPaths)-1; i >= 0; i-- {
		newPath = filepath.Join(newPath, nonexistentPaths[i])
	}

	return filepath.Clean(newPath), nil

```

cvgw: Yup :)

Though a minor suggestion
```
for i := len(nonexistentPaths)-1; i >= 0; i-- {
  newPath = filepath.Join(newPath, nonexistentPaths[i])
}
```
to
```
sort.Reverse(nonexisitentPaths)
newPath = filepath.Join(newPath, nonexistentPaths...)
```

WDYT?

shortstories: In that case, the code actually looks like this:
```go
sort.Sort(sort.Reverse(sort.StringSlice(nonexistentPaths)))
newPath = filepath.Join(append([]string{newPath}, nonexistentPaths...)...)
```
Umm... I think it's a bit harder to read..

PR #957, opened 2020-01-08
Upgrade go-containerregistry third-party library

antechrestos:
Fixes #808

**Description**

In this change I updated `go-containerregistry` dependency to a more recent version. It was needed to get the fix brought by the [following pull request on the third party project](https://github.com/google/go-containerregistry/pull/604)
By doing so, it forced me to upgrade to newer dependency and/or versions.

One of them is `moby/buildkit`. This late one is developed in go **1.11**. Starting version 1.13, versioning does not allow  invalid pseudo-version. The invalid pseudo versions used by `moby/buildkit` (or dependency using itself an invalid pseudo version) are the one that are declared in the `replace`section.

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes **Not needed**
- [ ] Adds integration tests if needed  **Not needed**

**Reviewer Notes**

- [ ] The code flow looks good. 
- [ ] Unit tests and or integration tests added.

**Release Notes**

Describe any changes here so maintainer can include it in the release notes, or delete this block.

```
Fix issues when pushing images on some registry built from a base image hosted on a registry different than docker hub and target registry
```

PR #959, opened 2020-01-08
Use cache instead of vendor

antechrestos:
**Description**

This work is a proposal. In another [pull request](https://github.com/GoogleContainerTools/kaniko/pull/957), I had more than 1K files modified just by changing a dependency version.

I took sight it might also be a good way to introduce an undesired/unhandled/malicious modification in vendor files as currently travis does not check consistency between `vendor` directory and `go.mod` file.

In this change I did the following things

- remove the `vendor` directory
- create a `.travis/go-cache`directory:  to create this, I cleaned my own cache directory ( `go clean --modcache`), deleted the `vendor` directory, and did a `go mod vendor`. Then I created the `.travis/go-cache` directory from `$GOPATH/pkg/mod`directory
- updated `.travis.yml`, `integration-test.sh` to copy cache to `$GOPATH/pkg/mod` before running a `go mod vendor`
- Updated `DEVELOPMEND.md` file to tell any neww developer to run a `go mod vendor` after cloning repository.
- update the two script `hack/boilerplate.sh` and `hack/gofmt.sh` to  ignore the `.travis` directory
- added the `vendor`directory to the `.gitignore` file

I cut my laptop from network and made a `cp -r ./.travis/go-cache $GOPATH/pkg/mod && go mod vendor && make out/executor`and it worked perfectly.

This way, it may be a good thing to let the core team be the protector of the `.travis/go-cache` directory (pull requests will download new dependencies in their jobs but cache will be updated by core team once it is merged).

What do you think of it @cvgw @tejal29 ?

**Submitter Checklist**

review them:

- [ ] Includes [unit tests]**not needed**
- [ ] Adds integration tests if needed. **not needed**

**Reviewer Notes**


tejal29:
@antechrestos Thank you for your PR.
I took a quick glance and have couple of questions.

1. IIUC, we have just moved all the files in `vendor` to `.travis/go-cache` right?
2. What happens when a developer adds a new dependency? They will have to add that to `.travis/go-cache` now or else travis ci will end up downloading this new dependency for every single run? So we will still have the same issue of large diffs when adding a new dependency?


Thanks
Tejal

antechrestos:
@tejal29 

1. More or less. In fact I just put in `.travis/go-cache` the cache generated by `go` when you generate `vendor`directory with `go mod vendor `command`. On your own computer you can see that in `$GOPATH/pkg/mod`there are both go sources and additional database used by `go`
2. On this point I would suggest that external contributors  (by it I mean out of the collaborators) only change the `go.mod` file. Travis would end downloading new dependencies as long as collaborators does not update the local cache. This will lower the number of file changed in pull requests and prevent any malicious use of `vendor` libraries.

cvgw:
I'm not sure I follow the intent of this PR. Can we open an issue to discuss this further?

tejal29:
@antechrestos At this time, i don't mind reviewing the larger diffs when dependencies are added Vs  adding additional maintenance load of updating the travis cache.

In future we can re-evaluate this decision. 
Sorry and thank you for bringing this up!

Thanks
Tejal

antechrestos:
@tejal29 Thank you for this feedback :smile: . 
Good to know that there is a vigilance about dependencies.

Cheers

Issue #579, opened 2019-02-22
Copy From with Chown on Multistage does not chown


kferrone:
**Actual behavior**
When I use a multi stage build which copies from the first stage to the second with a `--chown`, the directories are not actually `chowned`. 

**Expected behavior**
The chown on a `copy` should actually chown to the selected user. 

**To Reproduce**
1. Use this simple NodeJS DockerFile with any node project. 

```
FROM node:10.11.0 as builder

WORKDIR /home/node

COPY . .

RUN npm install --only=production --verbose

FROM node:10.11.0-alpine as runner

ENV NPM_CONFIG_LOGLEVEL=info \
	NODE_ENV=production

WORKDIR /home/node

COPY --from=builder --chown=node:node /home/node .

EXPOSE 3500

USER node

RUN mkdir -p logs

CMD [ "npm", "start" ]
```

2. When you build this with docker, everything works. When built with kaniko, the line with `RUN mkdir -p logs` will fail with permission denied and the whole `/home/node` dir is owned by root still. 

**Additional Information**
 
Here is what I had to do to the Dockerfile to make it work in the mean time:

```
FROM node:10.11.0 as builder

WORKDIR /home/node

COPY . .

RUN npm install --only=production --verbose

FROM node:10.11.0-alpine as runner

ENV NPM_CONFIG_LOGLEVEL=info \
	NODE_ENV=production

WORKDIR /home/node

COPY --from=builder --chown=node:node /home/node .

EXPOSE 3500

RUN chown -R node:node /home/node && \
	mkdir -p logs

USER node

CMD [ "npm", "start" ]
```

cvgw:
related issue https://github.com/GoogleContainerTools/kaniko/issues/57

tejal29:
@kferrone I verified your dockerfile on #962 
```
INFO[0000] Resolved base name node:10.11.0 to node:10.11.0 
INFO[0000] Resolved base name node:10.11.0-alpine to node:10.11.0-alpine 
INFO[0000] Using dockerignore file: /workspace/.dockerignore 
INFO[0000] Resolved base name node:10.11.0 to node:10.11.0 
INFO[0000] Resolved base name node:10.11.0-alpine to node:10.11.0-alpine 
INFO[0000] Retrieving image manifest node:10.11.0       
INFO[0001] Retrieving image manifest node:10.11.0       
INFO[0002] Retrieving image manifest node:10.11.0-alpine 
[...]

Issue #550, opened 2019-02-02
COPY changes ownership of target directory


everflux:
**Actual behavior**
Copying everything to the workdir using
`COPY . .`
changes the file ownernship of the target directory to root.
For images that use a non-root user and a workdir that is owned by the this user, but not root, this leads to permission problems.

**Expected behavior**
Target directory is kept as is, but files are added.

**To Reproduce**
1. Use a container image with USER 1000 as last instruction in base image
2. Have in base image RUN mkdir /app && chown 1000:1000 /app
3. In Dockerfile used by kaniko COPY . .

**Additional Information**
Dockerfile
```
FROM trion/ng-cli:latest AS ngcli

RUN ls -la; touch foo; ls -la
COPY . .
RUN ls -la; rm foo; ls -la
```

**Output**
```
INFO[0083] args: [-c pwd; ls -la; touch foo; ls -la]    
/app
total 8
drwxr-xr-x 2 node root 4096 Feb  2 20:56 .
drwxr-xr-x 1 root root 4096 Feb  2 21:42 ..
total 8
drwxr-xr-x 2 node root 4096 Feb  2 21:42 .
drwxr-xr-x 1 root root 4096 Feb  2 21:42 ..
-rw-r--r-- 1 node root    0 Feb  2 21:42 foo
INFO[0083] Taking snapshot of full filesystem...        

INFO[0107] COPY . .
INFO[0108] Taking snapshot of files...           

INFO[0108] RUN ls -la; rm foo; ls -la                   
INFO[0108] cmd: /bin/sh                                 
INFO[0108] args: [-c ls -la; rm foo; ls -la]            
total 524
drwxr-xr-x 8 root root   4096 Feb  2 21:43 .
drwxr-xr-x 1 root root   4096 Feb  2 21:42 ..
```

**Workaround**
This is how I managed to get the build working, though it feels like a crutch

```
COPY . /tmp/src
RUN cp -r /tmp/src/* .; cp /tmp/src/.* .
```

hobti01:
`COPY --chown` has no effect

Dockerfile
```
FROM alpine
RUN adduser -u 1000 -G root -D pipeline
USER pipeline
COPY --chown=pipeline:root . /home/pipeline/
```

docker
```
drwxr-sr-x    1 pipeline root          4096 Mar  6 14:34 .
drwxr-xr-x    1 root     root          4096 Mar  6 14:03 ..
[...]

PR #962, opened 2020-01-10
Support COPY --chown flag

### LINKED ISSUE #9: Add support for --chown flag to COPY command
See #1 for details

xanonid:
This PR fixes #9, #550, #579 

**Description**

This PR adds the --chown flag to kaniko. This allows to build e.g. root-less containers more easily and avoids unnecessary additional layers from manual `chown` calls.

The PR does not yet contain integration tests as the container-diff cannot yet check file ownership (https://github.com/GoogleContainerTools/container-diff/issues/308). 

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

kaniko supports now the --chown flag for the COPY command.
```

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -215,7 +215,7 @@ func resolveEnvironmentBuildArgs(arguments []string, resolver func(string) strin
 // copy Dockerfile to /kaniko/Dockerfile so that if it's specified in the .dockerignore
 // it won't be copied into the image
 func copyDockerfile() error {
-	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, ""); err != nil {
+	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, "", -1, -1); err != nil {
 		return errors.Wrap(err, "copying dockerfile")
 	}
 	opts.DockerfilePath = constants.DockerfilePath
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -215,7 +215,7 @@ func resolveEnvironmentBuildArgs(arguments []string, resolver func(string) strin
 // copy Dockerfile to /kaniko/Dockerfile so that if it's specified in the .dockerignore
 // it won't be copied into the image
 func copyDockerfile() error {
-	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, "", -1, -1); err != nil {
+	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, "", util.DoNotChangeUID, util.DoNotChangeGID); err != nil {
 		return errors.Wrap(err, "copying dockerfile")
 	}
 	opts.DockerfilePath = constants.DockerfilePath
--- a/pkg/util/command_util.go
+++ b/pkg/util/command_util.go
@@ -328,9 +328,20 @@ Loop:
 }
 
 // Extract user and group id from a string formatted 'user:group'.
-// If gidEqualToUIDIfNotGiven is set, the gid is equal to uid if the group is not specified
+// If fallbackToUID is set, the gid is equal to uid if the group is not specified
 // otherwise gid is set to zero.
-func GetUIDAndGIDFromString(uidStr string, gidStr string) (uint32, uint32, error) {
+func GetUIDAndGIDFromString(userGroupString string, fallbackToUID bool) (uint32, uint32, error) {
+	userAndGroup := strings.Split(userGroupString, ":")
+	userStr := userAndGroup[0]
+	var groupStr string
+	if len(userAndGroup) > 1 {
+		groupStr = userAndGroup[1]
+	}
+
+	uidStr, gidStr, err := GetUserFromUsername(userStr, groupStr, fallbackToUID)
+	if err != nil {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by tejal29** [comment]
@xanonid Can you explain more on "This allows to build e.g. root-less containers more easily and avoids unnecessary additional layers from manual chown calls." ?

File: cmd/executor/cmd/root.go [hunk 1]
```diff
@@ -200,7 +200,7 @@ func resolveDockerfilePath() error {
 // copy Dockerfile to /kaniko/Dockerfile so that if it's specified in the .dockerignore
 // it won't be copied into the image
 func copyDockerfile() error {
-	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, ""); err != nil {
+	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, "", -1, -1); err != nil {
```
cvgw: lets make the two integers into constants so that there meaning is more clear to readers

pkg/util/command_util.go
```diff
@@ -321,7 +322,44 @@ Loop:
...
+	userAndGroup := strings.Split(userGroupString, ":")
+	userStr := userAndGroup[0]
+	var groupStr string
+	if len(userAndGroup) > 1 {
+		groupStr = userAndGroup[1]
+	}
+
+	uidStr, gidStr, err := GetUserFromUsername(userStr, groupStr, !gidEqualToUIDIfNotGiven)
+	if err != nil {
+		return 0, 0, err
```
cvgw: Making these constants would also be great

pkg/util/fs_util.go
```diff
@@ -513,9 +514,12 @@ func CopyDir(src, dest, buildcontext string) ([]string, error) {
 			logrus.Tracef("Creating directory %s", destPath)
 
 			mode := fi.Mode()
-			uid := int(fi.Sys().(*syscall.Stat_t).Uid)
-			gid := int(fi.Sys().(*syscall.Stat_t).Gid)
-
+			if uid < 0 {
+				uid = int64(int(fi.Sys().(*syscall.Stat_t).Uid))
```
cvgw: Is the `int(` call redundant now?

pkg/util/fs_util.go
```diff
@@ -575,9 +579,13 @@ func CopyFile(src, dest, buildcontext string) (bool, error) {
 		return false, err
 	}
 	defer srcFile.Close()
-	uid := fi.Sys().(*syscall.Stat_t).Uid
-	gid := fi.Sys().(*syscall.Stat_t).Gid
-	return false, CreateFile(dest, srcFile, fi.Mode(), uid, gid)
+	if uid < 0 {
```
cvgw: This looks really similar to the code at https://github.com/GoogleContainerTools/kaniko/pull/962/files#diff-d36eb675aa49a7b471e3a2be77005b18R517

would it be possible to unify them?

cmd/executor/cmd/root.go
johannsoden: done. There is now DoNotChangeUID and DoNotChangeGID.

pkg/util/command_util.go
johannsoden: These both zeros are only dummy values, as somewhat needs to be set to return in the err case. They are not used (and should not) if err is set.
What would be best practise here?

pkg/util/fs_util.go
johannsoden: yes, removed and moved to own function.

pkg/util/fs_util.go
johannsoden: yes, moved to own function. thanks.

johannsoden:
@tejal29 I added a unit test.
Testing here is not very easy as usually the user running the tests does not have root permission or is not in more than one group, i.e. cannot create files with more than one uid/gid. 

> @xanonid Can you explain more on "This allows to build e.g. root-less containers more easily and avoids unnecessary additional layers from manual chown calls." ?

When creating container which should run with a non-root user (a security best practise), the files copied to it needs the correct ownership if they need to be modified/accessed at runtime (depending on the exact permissions).
Without this patch, the only possibility to fix the permission is `RUN chown -R ...` or similar which adds another layer in the Docker container duplicating all affected files.
With this patch, the files can directly be copied with the correct permission to the container which avoids this additional "chown" call and the needless layer.

tejal29:
Thanks @xanonid Looks like there is conflict due to a my PR that went in before. I will try to pick this up and get it in this release.

anton-johansson:
I just bumped into this issue and was very happy to find this PR, being merged and all, nice work! However, it made it to `master` **just** after the latest release - `0.17.1` was created. Any plans of creating `0.18.0` or maybe even `0.17.2` anytime soon?

I will use a `latest` version for now, but I'll be more happy to use a release soon. :)

xanonid:
@anton-johansson @tejal29 0.18.0 was released some days ago. Unfortunately, the change by this PR is not mentioned in the release notes.

Misteur-Z:
If this takes too long to release, I don't mind creating a quick PR to mention in documentation that `--chown` flag is not support on `ADD` and `COPY` commands. It's kind of a limitation?

xanonid:
@Misteur-Z This PR is merged and released, so `chown` is supported for COPY now, but not yet for ADD.

PR #964, opened 2020-01-13
Fix bug with docker compatibility ArgsEscaped

samos123:
ArgsEscaped according to Docker docs should only be set in Windows
environments: https://docs.docker.com/engine/api/v1.30/

It was causing a few integration tests to fail with following message:
```
FAIL: TestRun/test_Dockerfile_test_metadata (8.48s)
           "Diff": {
             "Adds": [
               "ArgsEscaped: true"
             ],
             "Dels": [
               "ArgsEscaped: false"
             ]
```

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

Change ArgsEscaped from true to false since this should only be used in Windows environments.

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -18,6 +18,8 @@ set -ex
 GCS_BUCKET="${GCS_BUCKET:-gs://kaniko-test-bucket}"
 IMAGE_REPO="${IMAGE_REPO:-gcr.io/kaniko-test}"
 
+docker version
+
 # Sets up a kokoro (Google internal integration testing tool) environment
 if [ -f "$KOKORO_GFILE_DIR"/common.sh ]; then
     echo "Installing dependencies..."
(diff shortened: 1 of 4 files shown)


samos123:
kokoro is failing but I believe this might be due to different behaviour with different docker versions. The latest stable version will return `ArgsEscape: false` however kokoro seems to be running an older version of Docker that returns `ArgsEscaped: true`.

integration-test.sh
```diff
@@ -18,6 +18,8 @@ set -ex
 GCS_BUCKET="${GCS_BUCKET:-gs://kaniko-test-bucket}"
 IMAGE_REPO="${IMAGE_REPO:-gcr.io/kaniko-test}"
 
+docker version
```
tejal29: just 1 nit: you can remove this here and log it in the newly added  `getDockerMajorVersion`

samos123: getDockerMajorVersion uses a template string to only get engine version. The command in script shows more info that's useful for debugging so I think should be kept as is.

PR #967, opened 2020-01-14
Clean code (Condition is always 'false' because 'err' is always 'nil' ).

lou-lan:
Clean code (Condition is always 'false' because 'err' is always 'nil' ).


lou-lan:
@googlebot I signed it!

PR #971, opened 2020-01-16
Fix Symlinks not being copied across stages

### LINKED ISSUE #915: Inconsistent behavior for symlink file COPY compared with docker build
**Actual behavior**
For docker build, copy symlink would copy the actual file.

**Expected behavior**
For kaniko build, copy symlink would copy the symlink file.

**To Reproduce**
1. ...FROM any image with /bin/sh(e.g. golang)
2. ...FROM scratch
3. ...COPY --from=0 /bin/sh /bin/sh

**Additional Information**
 - Dockerfile 
```
FROM golang:1.13
FROM gcr.io/kaniko-project/executor
COPY --from=0 /bin/sh /bin/sh
```
   Please provide either the Dockerfile you're trying to build or one that can reproduce this error.
 - Build Context
   Please provide or clearly describe any files needed to build the Dockerfile (ADD/COPY commands)
 - Personal thought
Actually, /bin/sh is a symlink point to /bin/dash, but kaniko would only copy the symlink, which makes /bin/sh fail to work. In comparison, docker would copy the /bin/dash into /bin/sh.
In kaniko, I found that the Dir symlink would be resolved [here](https://github.com/GoogleContainerTools/kaniko/blob/d72a960600f388655171eec0bf797cd5fbece666/pkg/commands/copy.go#L210)
So far, the dir symlink would be resolved but the file symlink would be ignored. Any concern about that? Why they are treated in a different way?

tejal29:
Fixes #915

**Description**
In case of symlink files across multi-stage files, kaniko would only detect change in the symlink.
Logs suggested, it only saved the symlink and not the target. Due to which user saw a build failure.
```
INFO[0000] Resolved base name alpine:3.11 to alpine:3.11
INFO[0000] Resolved base name scratch to scratch
INFO[0000] Resolved base name alpine:3.11 to alpine:3.11
....
INFO[0015] Taking snapshot of full filesystem...
INFO[0017] Saving file /usr/lib/libstdc++.so.6 for later use.
....
INFO[0018] COPY --from=0 /usr/lib/libstdc++.so.6 /usr/lib/
error building image: error building stage: failed to execute command: symlink libstdc++.so.6.0.27 /usr/lib/libstdc++.so.6: no such file or directory
```

In this change, 
1. For symlinks, we identify the symlink as well as the target as files to saved
```
/ # /kaniko/executor -f dockerfiles/Dockerfile_test_copy_symlink --context=dir:///workspace --destination=gcr.io/tejal-test/test --tarPath=image.tar
....
INFO[0004] Taking snapshot of full filesystem...        
INFO[0004] Adding whiteout for /kaniko                  
INFO[0007] Saving file /usr/lib/libstdc++.so.6.0.27 for later use 
INFO[0007] Saving file /usr/lib/libstdc++.so.6 for later use 
INFO[0007] Deleting filesystem...                       
INFO[0007] No base image, nothing to extract            
INFO[0007] Unpacking rootfs as cmd COPY --from=t /usr/lib/libstdc++.so.6 /usr/lib/ requires it. 
INFO[0008] COPY --from=t /usr/lib/libstdc++.so.6 /usr/lib/ 
INFO[0008] Taking snapshot of files... 
```
2. When a symlink is copied across mutliple stages, the target file is also copies.
Success log above.

**Submitter Checklist**

review them:

will do.
- [X] Adds integration tests if needed.

**Reviewer Notes**

**Release Notes**

```
Fix symlinks files copies across multi-stage dockerfiles.
```

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_copy_symlink
+++ b/integration/dockerfiles/Dockerfile_test_copy_symlink
@@ -1,0 +1,5 @@
+FROM alpine:3.11 as t
+RUN apk add gcc
+
+FROM scratch
+COPY --from=t /usr/lib/libstdc++.so.6 /usr/lib/
--- a/integration/dockerfiles/Dockerfile_test_copy_symlink
+++ b/integration/dockerfiles/Dockerfile_test_copy_symlink
@@ -1,5 +1,6 @@
-FROM alpine:3.11 as t
-RUN apk add gcc
+FROM busybox as t
+RUN echo "hello" > /tmp/target
+RUN ln -s /tmp/target /tmp/link
 
 FROM scratch
-COPY --from=t /usr/lib/libstdc++.so.6 /usr/lib/
+COPY --from=t /tmp/link /tmp
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -587,11 +587,8 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 
 // fileToSave returns all the files matching the given pattern in deps.
 // If a file is a symlink, it also returns the target file.
-// It first returns all the target files and then symlinks so they can copied
-// in that order to the kaniko workspace.
 func filesToSave(deps []string) ([]string, error) {
 	srcFiles := []string{}
-	symLinks := []string{}
 	for _, src := range deps {
 		srcs, err := filepath.Glob(src)
 		if err != nil {
@@ -599,14 +596,12 @@ func filesToSave(deps []string) ([]string, error) {
 		}
 		for _, f := range srcs {
 			if link, err := util.EvalSymLink(f); err == nil {
-				symLinks = append(symLinks, f)
 				srcFiles = append(srcFiles, link)
-			} else {
-				srcFiles = append(srcFiles, f)
 			}
+			srcFiles = append(srcFiles, f)
 		}
 	}
-	return append(srcFiles, symLinks...), nil
+	return srcFiles, nil
 }
 
 func fetchExtraStages(stages []config.KanikoStage, opts *config.KanikoOptions) error {
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -563,7 +563,6 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 			}
 		}
 
-		//
 		filesToSave, err := filesToSave(crossStageDependencies[index])
 		if err != nil {
 			return nil, err
@@ -599,7 +598,7 @@ func filesToSave(deps []string) ([]string, error) {
 			return nil, err
 		}
 		for _, f := range srcs {
-			if link, err := util.CanonicalizeLink(f); err == nil {
+			if link, err := util.EvalSymLink(f); err == nil {
 				symLinks = append(symLinks, f)
 				srcFiles = append(srcFiles, link)
 			} else {
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -23,8 +23,6 @@ import (
 	"strconv"
 	"time"
 
-	otiai10Cpy "github.com/otiai10/copy"
-
 	"github.com/google/go-containerregistry/pkg/v1/partial"
 
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
@@ -565,6 +563,7 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 			}
 		}
 
+		//
 		filesToSave, err := filesToSave(crossStageDependencies[index])
 		if err != nil {
 			return nil, err
@@ -574,8 +573,8 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 			return nil, err
 		}
 		for _, p := range filesToSave {
-			logrus.Infof("Saving file %s for later use.", p)
-			otiai10Cpy.Copy(p, filepath.Join(dstDir, p))
+			logrus.Infof("Saving file %s for later use", p)
+			util.CopyFileOrSymlink(p, dstDir)
 		}
 
 		// Delete the filesystem
@@ -587,16 +586,28 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 	return nil, err
 }
 
+// fileToSave returns all the files matching the given pattern in deps.
+// If a file is a symlink, it also returns the target file.
+// It first returns all the target files and then symlinks so they can copied
+// in that order to the kaniko workspace.
 func filesToSave(deps []string) ([]string, error) {
-	allFiles := []string{}
+	srcFiles := []string{}
+	symLinks := []string{}
 	for _, src := range deps {
 		srcs, err := filepath.Glob(src)
 		if err != nil {
 			return nil, err
 		}
-		allFiles = append(allFiles, srcs...)
+		for _, f := range srcs {
+			if link, err := util.CanonicalizeLink(f); err == nil {
+				symLinks = append(symLinks, f)
+				srcFiles = append(srcFiles, link)
+			} else {
+				srcFiles = append(srcFiles, f)
+			}
+		}
 	}
-	return allFiles, nil
+	return append(srcFiles, symLinks...), nil
 }
 
 func fetchExtraStages(stages []config.KanikoStage, opts *config.KanikoOptions) error {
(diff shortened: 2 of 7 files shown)


integration/dockerfiles/Dockerfile_test_copy_symlink
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.11 as t
+RUN apk add gcc
+
+FROM scratch
+COPY --from=t /usr/lib/libstdc++.so.6 /usr/lib/
```
samos123: is /usr/lib/libstdc++.so.6 a symlink in alpine?

pkg/executor/build.go
```diff
@@ -565,6 +563,7 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 			}
 		}
 
+		//
```
samos123: nit: this should probably be removed?

integration/dockerfiles/Dockerfile_test_copy_symlink
wilmardo: Yes it is, see the original issue:
```
lrwxrwxrwx    1 root     root            19 Jan 13 19:11 /usr/lib/libstdc++.so.6 -> libstdc++.so.6.0.27
```

In my opinion a test like this would suffice and be less prone to errors (libstdc++.so changing version) than the current approach:
```Dockerfile
FROM alpine
FROM scratch
COPY --from=0 /bin/sh /bin/sh
```
`/bin/sh` is a symlink to `/bin/busybox` on Alpine and is one of the core principles.

pkg/snapshot/snapshot.go
```diff
@@ -179,16 +178,19 @@ func (s *Snapshotter) scanFullFilesystem() ([]string, []string, error) {
...
 		}
 		if fileChanged {
-			logrus.Tracef("Adding %s to layer, because it was changed.", path)
-			filesToAdd = append(filesToAdd, path)
+			files, err := filesWithLinks(path)
+			if err != nil {
+				return nil, nil, err
+			}
+			logrus.Debugf("Adding files %s to layer, because it was changed.", files)
+			filesToAdd = append(filesToAdd, files...)
```
cvgw: Why add both the link and fully resolved path?

pkg/executor/build.go
```diff
@@ -587,16 +586,28 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 	return nil, err
 }
 
+// fileToSave returns all the files matching the given pattern in deps.
+// If a file is a symlink, it also returns the target file.
+// It first returns all the target files and then symlinks so they can copied
+// in that order to the kaniko workspace.
```
cvgw: Is there a requirement or benefit to copying them in order?

pkg/snapshot/snapshot.go
```diff
@@ -170,7 +169,7 @@ func (s *Snapshotter) scanFullFilesystem() ([]string, []string, error) {
 	filesToAdd := []string{}
 	for path := range memFs {
 		if util.CheckWhitelist(path) {
-			logrus.Tracef("Not adding %s to layer, as it's whitelisted", path)
+			logrus.Infof("Not adding %s to layer, as it's whitelisted", path)
```
cvgw: Maybe debug? There are a lot of whitelisted files iirc

pkg/util/fs_util.go
```diff
@@ -677,3 +672,49 @@ func CreateTargetTarfile(tarpath string) (*os.File, error) {
...
+
+func CanonicalizeLink(path string) (string, error) {
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return "", err
+	}
+	if !IsSymlink(fi) {
+		return "", ErrNotSymLink
+	}
+	return filepath.EvalSymlinks(path)
```
cvgw: iirc `EvalSymlinks` won't raise an error if you pass it a resolved path. Is there a need to check if the path is a symlink? Is the error handling used as control flow somewhere else to determine whether a path is a symlink?

cvgw:
@tejal29 how similar to https://github.com/GoogleContainerTools/kaniko/pull/943 is this?

pkg/util/fs_util.go
```diff
@@ -677,3 +672,49 @@ func CreateTargetTarfile(tarpath string) (*os.File, error) {
...
+}
+
+// otiai10Cpy.Copy in case the src file is a symlink, will copy the target
+// file at destination instead of creating a symlink. See #915 for more details.
```
cvgw: I think updating this comment with more detail instead of a reference to the issue would be great.

Something like
```
// For cross stage dependencies kaniko must persist the referenced path so that it can be used in 
// the dependent stage. For symlinks we copy the target path because copying the symlink would 
// result in a dead link
```

tejal29:
> @tejal29 how similar to #943 is this?

I looked at the PR and looks like for symlinks, it fixes copy by copying the target file when a symlink is copied. The code does not change what file is saved.
For cross stage references,  for symlink is referenced, we should save both symlink as well as target. 

WDYT? 
i will check what docker does when copying a symlink and update this section.

cvgw:
> > @tejal29 how similar to #943 is this?
> 
> I looked at the PR and looks like for symlinks, it fixes copy by copying the target file when a symlink is copied. The code does not change what file is saved.
> For cross stage references, for symlink is referenced, we should save both symlink as well as target.
> 
> WDYT?
> i will check what docker does when copying a symlink and update this section.

Ya I agree, I think they are addressing two similar but different issues.
This PR: src is a symlink
That PR: dest is a symlink

pkg/util/fs_util.go
```diff
@@ -677,3 +672,49 @@ func CreateTargetTarfile(tarpath string) (*os.File, error) {
...
 }
+
+// Returns true if a file is a symlink
+func IsSymlink(fi os.FileInfo) bool {
+	return fi.Mode()&os.ModeSymlink != 0
+}
+
+var ErrNotSymLink = fmt.Errorf("not a symlink")
+
+func CanonicalizeLink(path string) (string, error) {
```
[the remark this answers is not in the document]
tejal29: done. turns out, i need both 
- `os.Readlink` for creating a symlink when copying a symlink file.
- `filepath.EvalSymlinks` to get the destination path if file needed to be saved

I now have two separate functions.

pkg/util/fs_util.go
tejal29: this is now changed. Please take a look again.

integration/dockerfiles/Dockerfile_test_copy_symlink
tejal29: yes @samos123  it is a symlink in alpine image which is causing an error.
Thanks @wilmardo.

pkg/snapshot/snapshot.go
tejal29: sorry not an intentional change.

pkg/snapshot/snapshot.go
tejal29: We need to save both symlink and target file to kaniko workspace if symlink is referenced across stages. 
If we don't then the symlink points to a deadlink and results into an error.

pkg/util/fs_util.go
```diff
@@ -519,7 +511,7 @@ func CopyDir(src, dest, buildcontext string) ([]string, error) {
 			if err := mkdirAllWithPermissions(destPath, mode, uid, gid); err != nil {
 				return nil, err
 			}
-		} else if fi.Mode()&os.ModeSymlink != 0 {
+		} else if IsSymlink(fi) {
 			// If file is a symlink, we want to create the same relative symlink
```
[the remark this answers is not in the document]
tejal29: yes. we still create same symlink in the layer. The only thing i changed in this PR is when symlinks are references across stages, 
1. Save the target file and not just the symlink file 

Rest all remains the same.

pkg/executor/build.go
tejal29: Previously, i was using `filepath.EvalSymlink` and it was requirement then or else `EvalSymlink` returned an error "target file does not exist"
With the current implementation it is no longer a requirement
```
for _, p := range filesToSave {
			logrus.Infof("Saving file %s for later use", p)
			util.CopyFileOrSymlink(p, dstDir)
		}

----
// For cross stage dependencies kaniko must persist the referenced path so that it can be used in
// the dependent stage. For symlinks we copy the target path because copying the symlink would
// result in a dead link
func CopyFileOrSymlink(src string, destDir string) error {
	destFile := filepath.Join(destDir, src)
	if fi, _ := os.Lstat(src); IsSymlink(fi) {
		link, err := os.Readlink(src)
		if err != nil {
			return err
		}
		return os.Symlink(link, destFile)
	}
	return otiai10Cpy.Copy(src, destFile)
}
```

tejal29: Let me try not retaining the order and see if it still matters.

pkg/snapshot/snapshot.go
cvgw: Might worth adding a comment. Your call

PR #973, opened 2020-01-17
Fix issue #704

### LINKED ISSUE #704: Removed files re-appear when using cache
**Actual behavior**

Building a Dockerfile which deletes a file _initially_ produces an image without the file. But when re-built using the cache, the resulting image contains the file again.

**Expected behavior**

The deleted file should never be present in final images, regardless of whether cache was used.

**To Reproduce**

1. Build a simple Dockerfile that adds a file then removes it in a later step, using Kaniko, with caching enabled:

```
FROM alpine

RUN mkdir -p /some/dir/ && echo 'first' > /some/dir/first.txt

RUN rm /some/dir/first.txt
```

2. Pull the resulting image from your registry, and run `ls -la /some/dir` in a container using the image. Observe that the directory is empty:

```
$ docker pull my-registry/my/img:latest && docker run --rm my-registry/my/img:latest bin/sh -c "ls -la /some/dir"
latest: Pulling from my/img
921b31ab772b: Already exists
b4f1d0fd5c4c: Pull complete
f6d9b826d061: Pull complete
Digest: sha256:213ace2d10ba510eeeaefa90b255da5f48e87ff7577a560bbb262b9db1cd9945
Status: Downloaded newer image for my-registry/my/img:latest
my-registry/my/img:latest
total 8
drwxr-xr-x    1 root     root          4096 Jun 24 14:27 .
drwxr-xr-x    1 root     root          4096 Jun 24 14:27 ..
```

3. Re-build the image using Kaniko, exactly as before. Logs will show that cached layers are being used.

4. Re-pull the image, re-run a container to `ls` again. Observe that the file has re-appeared.

```
docker pull my-registry/my/img:latest && docker run --rm my-registry/my/img:latest bin/sh -c "ls -la /some/dir"
latest: Pulling from my/img
921b31ab772b: Already exists
bd7f0d50b1ce: Pull complete
7d792274c478: Pull complete
Digest: sha256:9adda5568544f015972d112a9de87829ee2f02c0f65871ca2c632b10201be096
Status: Downloaded newer image for my-registry/my/img:latest
my-registry/my/img:latest
total 12
drwxr-xr-x    1 root     root          4096 Jun 24 14:34 .
drwxr-xr-x    1 root     root          4096 Jun 24 14:34 ..
-rw-r--r--    1 root     root             6 Jun 24 14:34 first.txt
```

**Additional Information**

- Example Dockerfile above.
- Build context is empty/irrelevant.
- Tried various combinations of `ADD <file-from-context> <path>`, `ADD <archive-of-directory>.tar.gz /`, `COPY <file-from-context> <path>` and using `RUN` steps to manipulate the files (as in the example). All combinations show this behaviour.
- This behaviour has been seen with various other later steps after the deletion (fs-related stuff, non-FS things like `ENTRYPOINT` etc).
- Kaniko Image details:

```
$ docker inspect gcr.io/kaniko-project/executor:v0.10.0
[
    {
        "Id": "sha256:9000801241318d6c9b8c2b505974f22fd74c57e0adfdb09c98870aef4923bb65",
        "RepoTags": [
            "gcr.io/kaniko-project/executor:latest",
            "gcr.io/kaniko-project/executor:v0.10.0"
        ],
        "RepoDigests": [
            "gcr.io/kaniko-project/executor@sha256:78d44ec4e9cb5545d7f85c1924695c89503ded86a59f92c7ae658afa3cff5400"
        ],
        ...
```

I can see from logs that a whiteout file is being created as expected in the first (uncached) build. However, it looks like the whiteout is either not making it's way into the cached data, or isn't being honoured when the cache is used later?

Peeking inside docker's `overlay2` directory confirms the lack of a whiteout in 2nd img. Upper layer internals in 1st image, built *without* cache:

```
/ # ls -la /var/lib/docker/overlay2/596f011a6b72769df25c4bd64e247bcede1f08f43f72be21790fc5a347d43921/diff/some/dir
total 8
drwxr-xr-x    2 root     root          4096 Jun 24 14:27 .
drwxr-xr-x    3 root     root          4096 Jun 24 14:27 ..
c---------    1 root     root        0,   0 Jun 24 14:28 first.txt
```

Upper layer internals for 2nd image, built *with* cache:

```
/ # ls -la /var/lib/docker/overlay2/afd849e49a216c891156d60f0dc99e4a6eb5b2a59ece4ca4d75e3705df9edc9e/diff/some/dir
total 8
drwxr-xr-x    2 root     root          4096 Jun 24 14:34 .
drwxr-xr-x    3 root     root          4096 Jun 24 14:34 ..
```

This behaviour has been observed in multiple environments:

- Using Jenkins on Kubernetes in GCP (as per [official suggestion](https://cloud.google.com/solutions/jenkins-on-kubernetes-engine) & [example](https://github.com/jenkinsci/kubernetes-plugin/blob/kubernetes-1.14.5/examples/kaniko-declarative.groovy)), with gcr.io  as final img dest & a k8s-deployed docker registry for cache.
- On a mac, using Kaniko via docker, internal private registry as final img dest & locally-running registry for cache (for easy testing)

cvgw:
Fixes #704 

**Description**
Previously whiteout files were not extracted from cached layers and not included in the list of extracted files. This meant that if a file was deleted in a cached layer it would "reappear" in the final image as it was missing the whiteout file.

This change adds a new method `GetFSFromLayers` to which `GetFSFromImage` now delegates. `GetFSFromLayers` accepts a series of options, one of which `IncludeWhiteout` will cause whiteout files to be extracted and returned in the list of extracted files.

Additionally, this change updates `CachingRunCommand` and `CachingCopyCommand` to use `GetFSFromLayers` and pass the option to include the whiteout files. This results in the whiteout files being extracted from cached layers and the deleted file properly removed from the final image.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Fix bug with deleted files and cached run and copy commands
```


pkg/util/fs_util.go
```diff
@@ -71,25 +71,58 @@ var excluded []string
...
+
+type FSOpt func(*FSConfig)
+
+func IncludeWhiteout() FSOpt {
+	return func(opts *FSConfig) {
+		opts.includeWhiteout = true
+	}
+}
+
+func ExtractFunc(extractFunc ExtractFunction) FSOpt {
```
samos123: I couldn't wrap my head around ExtractFunc taking an ExtractFunction type with parameter name extractFunc. I don't have a recommendation either since my go skills are rather poor so just fyi.

cvgw: WDYT about `func ExtractFuncOpt(f ExtractFunction) FSOpt`

pkg/util/fs_util.go
```diff
@@ -71,25 +71,58 @@ var excluded []string
...
+type FSConfig struct {
+	includeWhiteout bool
+	extractFunc     ExtractFunction
+}
+
+type FSOpt func(*FSConfig)
+
+func IncludeWhiteout() FSOpt {
+	return func(opts *FSConfig) {
+		opts.includeWhiteout = true
```
tejal29: this is always true. Do you see any future this being configured via command line? i am wondering if we could make the code simpler and avoid this config.

cvgw: I don't see it ever being configured via command line since a single process will call FSFromLayers with different options.

This is always true because the zero value is false. So by not including this option it is set to false.

I might be misunderstanding

cvgw: I guess I could see an argument for passing the value of IncludeWhitespace as an arg to the function so it could be switched from true to false but that seems like something that could be changed later if the need arisea

pkg/util/fs_util.go
samos123: ah I know get it was well, might just be me not getting it the other night, never seen this pattern before. I think keeping as is is fine

pkg/util/fs_util.go
tejal29: actually i meant the opposite. Lets remove `opts.includeWhiteout` totally

cvgw: I don't understand. When `GetFSFromImage` calls `GetFSFromLayers` `includeWhiteout` should be false. When `CachingCopyCommand` or `CachingRunCommand` calls `GetFSFromLayers` `includeWhiteout` should be true.

so `includeWhiteout` has to be configurable in some regards. Are you objecting to having it be configurable or this specific configuration pattern?

tejal29: Ah, i think it did not occur to me `includeWhiteout` was false due to the `opts .. FSOpt` array. 
That explains.

```
	return GetFSFromLayers(root, layers, ExtractFunc(extract))
}

func GetFSFromLayers(root string, layers []v1.Layer, opts ...FSOpt) ([]string, error) {

```

Issue #903, opened 2019-12-05
Symlinks to absolute paths don't work properly


cpointner:
**Actual behavior**
Ubuntu images have a symlink `/var/run` that points to the absolute path `/run`. Any file that gets created inside `/var/run` does not end up in the image. If i change the link to be relative everything works as expected.

**Expected behavior**
Symlinks to absolute paths must be based on the image root.

**To Reproduce**

Create a temporary directory and copy the following Dockerfile to it:
```
FROM ubuntu:bionic
RUN mkdir -p /var/run/foo /run/bar
CMD [ "find", "/run" ]
```
Inside the directory run the following commands:
```
docker run --rm -it -v $(pwd):/workspace gcr.io/kaniko-project/executor:latest --dockerfile=/workspace/Dockerfile --context=/workspace --destination=symlinks:kaniko --tarPath /workspace/symlinks-kaniko.tar --no-push
docker load < symlinks-kaniko.tar
docker run --rm -it symlinks:kaniko
```

This yields the following output:

```
/run
/run/bar
/run/systemd
/run/systemd/container
/run/lock
/run/utmp
/run/mount
/run/mount/utab
```

Whereas the image built using:
```
docker build -t symlinks:docker .
docker run --rm -it symlinks:docker
```
yields the following output:

```
/run
/run/bar
/run/foo
/run/systemd
/run/systemd/container
/run/lock
/run/utmp
/run/mount
/run/mount/utab
```
Mind the missing `foo` in the first output.

At the time i tested this the kaniko image latest tag pointed to `sha256:9c40a04cf1bc9d886f7f000e0b7fa5300c31c89e2ad001e97eeeecdce9f07a29`

 
 **Triage Notes for the Maintainers**
 

 
 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li>- [ ] </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li>- [x] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [x] </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [ ] </li></ul>|

cvgw:
Hi @cpointner is it possible this is a duplicate of https://github.com/GoogleContainerTools/kaniko/issues/506 and not specific to symlinks?

MarkusTeufelberger:
[...]

PR #975, opened 2020-01-18
Fix #506 add option additional-whitelist

### LINKED ISSUE #506: Paths under /var/run ignored
**Actual behavior**
Directories and files created under /var/run do not exist in created image.

**Expected behavior**
All directories and files created in Dockerfile should exist in created image.

**To Reproduce**
Create directory /var/run/test in Dockerfile. Observe that /var/run/test doesn't exist in resulting image.

**Additional Information**
 - Dockerfile 
FROM debian
RUN mkdir -p /var/run/test

### LINKED ISSUE #704: Removed files re-appear when using cache

cvgw:
Fixes #506, #903 

**Description**

Add a new option additional-whitelist which defaults to a single entry, "/var/run". This will allow users to remove "/var/run" from the whitelist or retain the current behavior with no change.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- add new flag to customize ignored directories
```

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -38,9 +38,10 @@ import (
 )
 
 var (
-	opts     = &config.KanikoOptions{}
-	logLevel string
-	force    bool
+	opts                = &config.KanikoOptions{}
+	logLevel            string
+	force               bool
+	additionalWhitelist []string
 )
 
 func init() {
@@ -72,6 +73,16 @@ var RootCmd = &cobra.Command{
 			}
 			if len(opts.Destinations) == 0 && opts.ImageNameDigestFile != "" {
 				return errors.New("You must provide --destination if setting ImageNameDigestFile")
+			}
+
+			if additionalWhitelist == nil {
+				additionalWhitelist = []string{
+					"/var/run",
+				}
+			}
+
+			for _, path := range additionalWhitelist {
+				util.AddToWhitelist(path)
 			}
 		}
 		return nil
@@ -144,6 +155,10 @@ func addKanikoOptionsFlags() {
 	RootCmd.PersistentFlags().DurationVarP(&opts.CacheTTL, "cache-ttl", "", time.Hour*336, "Cache timeout in hours. Defaults to two weeks.")
 	RootCmd.PersistentFlags().VarP(&opts.InsecureRegistries, "insecure-registry", "", "Insecure registry using plain HTTP to push and pull. Set it repeatedly for multiple registries.")
 	RootCmd.PersistentFlags().VarP(&opts.SkipTLSVerifyRegistries, "skip-tls-verify-registry", "", "Insecure registry ignoring TLS verify to push and pull. Set it repeatedly for multiple registries.")
+
+	// We use nil as the default value so we can differentiate between the flag passed
+	// with an empty list and the flag not set
+	RootCmd.PersistentFlags().StringSliceVar(&additionalWhitelist, "additional-whitelist", nil, "Paths to whitelist. These will be ignored be kaniko to improve performance.")
 }
 
 // addHiddenFlags marks certain flags as hidden from the executor help text
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -75,12 +75,6 @@ var RootCmd = &cobra.Command{
 				return errors.New("You must provide --destination if setting ImageNameDigestFile")
 			}
 
-			if additionalWhitelist == nil {
-				additionalWhitelist = []string{
-					"/var/run",
-				}
-			}
-
 			for _, path := range additionalWhitelist {
 				util.AddToWhitelist(path)
 			}
@@ -158,7 +152,7 @@ func addKanikoOptionsFlags() {
 
 	// We use nil as the default value so we can differentiate between the flag passed
 	// with an empty list and the flag not set
-	RootCmd.PersistentFlags().StringSliceVar(&additionalWhitelist, "additional-whitelist", nil, "Paths to whitelist. These will be ignored be kaniko to improve performance.")
+	RootCmd.PersistentFlags().StringSliceVar(&additionalWhitelist, "additional-whitelist", []string{}, "Paths to whitelist. These will be ignored by kaniko to improve performance.")
 }
 
 // addHiddenFlags marks certain flags as hidden from the executor help text
--- a/pkg/util/fs_util_test.go
+++ b/pkg/util/fs_util_test.go
@@ -38,26 +38,12 @@ import (
 )
 
 func Test_DetectFilesystemWhitelist(t *testing.T) {
-	testDir, err := ioutil.TempDir("", "")
-	if err != nil {
-		t.Fatalf("Error creating tempdir: %s", err)
-	}
-	fileContents := `
-	228 122 0:90 / / rw,relatime - aufs none rw,si=f8e2406af90782bc,dio,dirperm1
-	229 228 0:98 / /proc rw,nosuid,nodev,noexec,relatime - proc proc rw
-	230 228 0:99 / /dev rw,nosuid - tmpfs tmpfs rw,size=65536k,mode=755
-	231 230 0:100 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666
-	232 228 0:101 / /sys ro,nosuid,nodev,noexec,relatime - sysfs sysfs ro`
-
-	path := filepath.Join(testDir, "mountinfo")
-	if err := os.MkdirAll(filepath.Dir(path), 0750); err != nil {
-		t.Fatalf("Error creating tempdir: %s", err)
-	}
-	if err := ioutil.WriteFile(path, []byte(fileContents), 0644); err != nil {
-		t.Fatalf("Error writing file contents to %s: %s", path, err)
+	type testcase struct {
+		desc                string
+		additionalWhitelist []string
+		expectedWhitelist   []WhitelistEntry
 	}
 
-	err = DetectFilesystemWhitelist(path)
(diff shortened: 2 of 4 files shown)


cvgw:
So this is working without caching but it doesn't work with caching, I think because `/var/run` is actually a link to `/run` on a lot of machines. My cached image ends up with `/run/test-folder` buut not `/var/run/test-folder` similar to #915 and as a result I think this PR should be merged after that is fixed.

pkg/util/fs_util_test.go
```diff
@@ -58,7 +58,6 @@ func Test_DetectFilesystemWhitelist(t *testing.T) {
 		{"/dev", false},
 		{"/dev/pts", false},
 		{"/sys", false},
-		{"/var/run", false},
```
tejal29: This test is little confusing. Seems like we are testing 
- whitelist before `AddToWhitelist` 
- whitelist after `AddToWhitelist`

Can we break this into 2 tests or only one after `AddToWhitelist` should suffice.

tejal29: Ideally, it would be great if we convert this to Test Table driven with test cases
- no addiional whitelists
- 1 additional whitelist
- 1 or more additional whitelist

cmd/executor/cmd/root.go
```diff
@@ -73,6 +74,16 @@ var RootCmd = &cobra.Command{
 			if len(opts.Destinations) == 0 && opts.ImageNameDigestFile != "" {
 				return errors.New("You must provide --destination if setting ImageNameDigestFile")
 			}
+
+			if additionalWhitelist == nil {
+				additionalWhitelist = []string{
```
tejal29: i think we should default this to empty string. 
if users want to remove `/var/run` then they would have to run kaniko as
`/kaniko/executor --additional-whitelist=""`  instead users who need it to whitelisted should explicitly specify it.

e.g. #903

cvgw: >if users want to remove /var/run then they would have to run kaniko as
/kaniko/executor --additional-whitelist="" instead users who need it to whitelisted should explicitly specify it.

I'm not strongly opposed to that but it seems like it would be a pretty major and invisible change for users. I can imagine users who don't realize they need to explicit add `/var/run` having their build times dramatically increase.

I think defaulting it to the current behavior would be a better user experience and let people opt in. I can see how an empty string might be a better experience for new users though.

WDYT

cvgw:
This should include a README update before merging. Probably makes sense to agree on the default behavior before documenting.

cmd/executor/cmd/root.go
tejal29: I also thought about being backward compatible for existing users. 
 But we want these users to eventually correct their pod config and explicitly add `/var/run` to `additional-whitelist`. 
I would want us to push them towards that direction. 
This is challenging for us especially since users don't have to change their image tag if they are pointing to `latest`

Maybe for a couple of releases we could add a visible Warning that `/var/run` needs to be explicitly whitelisted especially if their build times have increased. 
Also monitor slack closely.

WDYT?

cvgw:
Ya, I think a clear warning and good communication should handle most of
our problems. Let's do that. I'll make the change here.

cmd/executor/cmd/root.go
cvgw: linking for history https://github.com/GoogleContainerTools/kaniko/pull/975#issuecomment-577847191

pkg/util/fs_util_test.go
cvgw: done :+1:

cmd/executor/cmd/root.go
cvgw: done :+1:

helder-moreira:
@tejal29 this pull request just broke a lot of pipelines I currently manage. Here is the error:

```
error building image: error building stage: failed to get filesystem from image: error removing var/run to make way for new symlink: unlinkat /var/run/secrets/kubernetes.io/serviceaccount/..2020_01_25_01_10_27.560757320: read-only file system
```

cvgw:
> @tejal29 this pull request just broke a lot of pipelines I currently manage. Here is the error:
> 
> ```
> error building image: error building stage: failed to get filesystem from image: error removing var/run to make way for new symlink: unlinkat /var/run/secrets/kubernetes.io/serviceaccount/..2020_01_25_01_10_27.560757320: read-only file system
> ```

@helder-moreira are you using tag `latest`? Might be worth specifying a specific release of Kaniko like `v0.16.0`. Also there is also an open issue you can add your info to https://github.com/GoogleContainerTools/kaniko/issues/1002

helder-moreira:
@tejal29 Thanks. I have already moved to and older version.

I will follow that issue, thanks.

tejal29:
@helder-moreira This PR has been merged but not released.

helder-moreira:
@tejal29 Got it. I was using the `debug` tag. I have now moved to `debug-v0.16.0`. Thanks.

tejal29:
Thanks @helder-moreira. The team wasn't aware debug tag was updated on every push to master. Sorry for breaking your workflow. We will investigate and fix this soon.

Issue #966, opened 2020-01-13
Desitination using --tarPath doesn't work if base image is in gcr.io


samos123:
**Actual behavior**
I tried to use --tarPath where the base image is in gcr.io which will cause it to try to authenticate to gcr.io even for public images.

**Expected behavior**
--tarPath to work with any kind of image

**To Reproduce**
1. Run the following and watch it fail:
```
docker run \                                      
  -v $(pwd):/workspace gcr.io/kaniko-project/executor:latest \
  --dockerfile=/workspace/integration/dockerfiles/Dockerfile_test_cmd --context=dir:///workspace/integration \
--no-push --tarPath /workspace/test_add.tar --force
```

**Additional Information**
 - Dockerfile: https://github.com/GoogleContainerTools/kaniko/blob/master/integration/dockerfiles/Dockerfile_test_cmd
 - Build Context part of integration tests
 - Kaniko Image (fully qualified with digest): gcr.io/kaniko-project/executor                                
e90df5b891a7
 
 **Triage Notes for the Maintainers**
 

 
 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li>- [ ] </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li>- [ ] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [ ] </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [ ] </li></ul>|

filesnate:
See https://github.com/GoogleContainerTools/kaniko/pull/1125 which may be related

Issue #947, opened 2020-01-03
Travis, Docker, and Kokoro all use different golang versions


cvgw:
**Actual behavior**
Kokoro, Travis, and the Kaniko Dockerfiles all use different Golang versions

Travis is 1.13
https://github.com/GoogleContainerTools/kaniko/blob/eafb80227b044b521f261b3883f094da0dd29950/.travis.yml#L5

Dockerfile is 1.12
https://github.com/GoogleContainerTools/kaniko/blob/eafb80227b044b521f261b3883f094da0dd29950/deploy/Dockerfile#L17

I have no idea what Kokoro is using but it's not 1.13 because `testing.Init` is not available and that was added in 1.13

**Expected behavior**
Kokoro, Travis, and the Kaniko docker files should all use the same golang version

**To Reproduce**
N/A

**Additional Information**
 
 **Triage Notes for the Maintainers**
 

 
 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li>- [ ] </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li>- [ ] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [ ] </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [ ] </li></ul>|

tejal29:
This is being tackled by @samos123

tejal29:
@samos123 is this fixed now?

samos123:
Yes, however I will still need to update the Dockerfile to use 1.13 as well. It's still using 1.12 in the dockerfile. Let me create a PR for that.

samos123:
Actually seems the Dockerfiles have already been updated to use 1.13 in master. So this can be closed now.

PR #979, opened 2020-01-19
Run integration tests in Travis CI

samos123:
- Change integration tests to use docker hub instead of GCR due to bug
in library that requires authentication with gcr.io even for public
images. See #966 for bug tracking this.
- Make uploading to GCS bucket configurable through a flag --uploadToGCS
- Utilize a locally deployed docker registry in travis CI to remove
dependency of needing to authenticate with GCP. This requires host
networking such that we can access the registry on localhost:5000
- Use the commit that's being tested for `TestGitBuildcontext`
- Remove duplicate GitBuildContext case that's now covered by default in
travis CI.

Related to #947 

**Submitter Checklist**

review them:

**Reviewer Notes**

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_metadata
+++ b/integration/dockerfiles/Dockerfile_test_metadata
@@ -1,4 +1,4 @@
-FROM gcr.io/distroless/base@sha256:628939ac8bf3f49571d05c6c76b8688cb4a851af6c7088e599388259875bde20
+FROM debian:10.2
 CMD ["command", "one"]
 CMD ["command", "two"]
 CMD echo "hello"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,21 @@
 language: go
 os: linux
-
+dist: bionic
+env:
+  - IMAGE_REPO=localhost:5000
 go:
   - "1.13.3"
 go_import_path: github.com/GoogleContainerTools/kaniko
+before_install:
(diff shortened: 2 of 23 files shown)


integration/dockerfiles/Dockerfile_test_metadata
```diff
@@ -1,4 +1,4 @@
-FROM gcr.io/distroless/base@sha256:628939ac8bf3f49571d05c6c76b8688cb4a851af6c7088e599388259875bde20
+FROM debian:10.2
```
cvgw: ok to change this one from distroless/base to debian?

tejal29:
This is great!! 
My only concern is we have increased the Travis CI time from 6 mins to 20 mins. 
Can we create a Matrix CI job and run `make integration` be another sub job. That way we can reduce the wait time to ~14 mins ?
What do you think?

Would you be able to do this in this PR ? or want to follow up in another PR?

integration/dockerfiles/Dockerfile_test_metadata
samos123: since there is no reason to use distroless for the test case. Distroless is only available on GCR and there is an issue with GCR when you're not authenticated with gcloud.

samos123:
Let's separate that in another PR. I think 20 mins is still acceptable for a full integration testing run since previously it was only running unit tests.

samos123:
also I suggest to keep kokoro running for the the next few weeks at least. This way we can easily see whether some issues might not be catched with a local registry vs using gcr.

Issue #970, opened 2020-01-16
Snapshotting is taking > 10 minutes


nicolaferraro:
**Actual behavior**

We're using Kaniko in Camel K and some users are complaining about builds that became slow after we upgraded to v0.14.0 from 0.9.0 (https://github.com/apache/camel-k/issues/1209) especially on Minikube (towards Minikube internal addon registry).

Under the hood, we use the `fabric8/s2i-java:3.0-java8` image which is a bit heavy, but it seems the logs get stuck when doing the snapshot of some jar files in a specific dir (~50 jars):

```
camel-k-kit-bofk6gui1aal47gim840-builder kaniko INFO[0492] Downloading base image 10.111.197.108/chapter-7/camel-k-kit-bofj39ei1aal47gim830:36212
camel-k-kit-bofk6gui1aal47gim840-builder kaniko INFO[0647] Skipping unpacking as no commands require it.
camel-k-kit-bofk6gui1aal47gim840-builder kaniko INFO[0647] Taking snapshot of full filesystem...
camel-k-kit-bofk6gui1aal47gim840-builder kaniko INFO[0647] Using files from context: [/builder/kit-bofk6gui1aal47gim840/package/context]
camel-k-kit-bofk6gui1aal47gim840-builder kaniko INFO[0647] ADD . /deployments
camel-k-kit-bofk6gui1aal47gim840-builder kaniko INFO[0647] Taking snapshot of files...
```

Is there any way to speedup Kaniko builds, e.g. by installing something on the host machine?
Shouldn't the snapshot be limited to the added directory after an "ADD" command?

**Expected behavior**
It should not take ages to build an image.

**To Reproduce**
1. Install Minikube
2. Enable the registry addon
3. Install [Camel K 1.0.0-RC1](https://github.com/apache/camel-k/releases/tag/1.0.0-RC1) (`kamel install`)
4. Build an integration `kamel run https://raw.githubusercontent.com/apache/camel-k/master/examples/simple.groovy`
5. A kaniko build pod will be started, preceded by an init-container

 **Triage Notes for the Maintainers**
 

 
 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li>- [ ] </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li>- [ ] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [ ] </li></ul>|
[...]

PR #989, opened 2020-01-23
fix method documentation, better debug log and add a test case

tejal29:
Helps debug #970


tejal29:
restarting both builds.

PR #993, opened 2020-01-23
Expand build argument from environment when no value specified

### LINKED ISSUE #713: Kaniko's --build-arg accepts only literal values
Actual behavior
Using encrypted variables is working fine with cloud build image. 
Docker also supports copying a value from the environment by just specifying NAME[1].
But if I switch to kaniko, it doesn't pass the secret values as a build argument. Kaniko works only when I pass constant value, eg NAME=value. 

Expected behavior
Be able to use the same way as docker does, the environment variables on Kaniko

To Reproduce

command:
gcloud builds submit --config=cloudbuild-kaniko.yaml .

Dockerfile:
FROM busybox
ARG THE_SECRET
RUN echo "::${THE_SECRET}::"

cloudbuild-kaniko.yaml:
steps:
- id: bash
  name: 'gcr.io/cloud-builders/docker'
  entrypoint: "bash"
  args: ['-c', 'echo "$$THE_SECRET"']
  secretEnv: ['THE_SECRET']
- id: docker
  name: 'gcr.io/kaniko-project/executor:latest'
  args: 
    - --destination=gcr.io/[project-id]/image
    - --build-arg=THE_SECRET=$$THE_SECRET 
    - --build-arg=THE_SECRET=secrettttttt  
  secretEnv: ['THE_SECRET']
secrets:
- kmsKeyName: projects/[project-id]/locations/global/keyRings/tinkering/cryptoKeys/cloud-build-demo
  secretEnv:
    THE_SECRET: XXXXXX

____
[1]: https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg

antechrestos:
Fixes #713 

**Description**

This change adds the feature of getting value of build argument from environment when build argument is specified as follows

`/kaniko/executor .... --build-arg MY_ARGUMENT ...` 

**Submitter Checklist**

**Reviewer Notes**

- [x] Unit tests and or integration tests added.

**Release Notes**

```
Allow user to provide build argument from environment variable as follows:
$> export MY_ARGUMENT="the value"
$> /kaniko/executor ... --build-arg MY_ARGUMENT ...
Or
$> docker run -e MY_ARGUMENT="the value" gcr.io/kaniko-project/executor ... --build-arg MY_ARGUMENT ...
```

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -55,6 +55,7 @@ var RootCmd = &cobra.Command{
 	Use: "executor",
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 		if cmd.Use == "executor" {
+			resolveEnvironmentBuildArgs(opts.BuildArgs, os.Getenv)
 			if err := util.ConfigureLogging(logLevel); err != nil {
 				return err
 			}
@@ -198,6 +199,17 @@ func resolveDockerfilePath() error {
 		return copyDockerfile()
 	}
 	return errors.New("please provide a valid path to a Dockerfile within the build context with --dockerfile")
+}
+
+// resolveEnvironmentBuildArgs replace build args without value by the same named environment variable
+func resolveEnvironmentBuildArgs(arguments []string, resolver func(string) string) {
+	for index, argument := range arguments {
+		i := strings.Index(argument, "=")
+		if i < 0 {
+			value := resolver(argument)
+			arguments[index] = fmt.Sprintf("%s=%s", argument, value)
+		}
+	}
 }
 
 // copy Dockerfile to /kaniko/Dockerfile so that if it's specified in the .dockerignore
--- a/integration/images.go
+++ b/integration/images.go
@@ -17,6 +17,7 @@ limitations under the License.
 package integration
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -46,10 +47,11 @@ const (
 
 // Arguments to build Dockerfiles with, used for both docker and kaniko builds
 var argsMap = map[string][]string{
-	"Dockerfile_test_run":     {"file=/file"},
-	"Dockerfile_test_workdir": {"workdir=/arg/workdir"},
-	"Dockerfile_test_add":     {"file=context/foo"},
-	"Dockerfile_test_onbuild": {"file=/tmp/onbuild"},
+	"Dockerfile_test_run":        {"file=/file"},
+	"Dockerfile_test_workdir":    {"workdir=/arg/workdir"},
+	"Dockerfile_test_add":        {"file=context/foo"},
+	"Dockerfile_test_arg_secret": {"SSH_PRIVATE_KEY", "SSH_PUBLIC_KEY=Pµbl1cK€Y"},
+	"Dockerfile_test_onbuild":    {"file=/tmp/onbuild"},
 	"Dockerfile_test_scratch": {
 		"image=scratch",
 		"hello=hello-value",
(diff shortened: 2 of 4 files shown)


antechrestos:
@cvgw As promised... Do I need to add integration test?

tejal29:
Thank you so much for your contribution!

stewi2:
> won't `docker -e SOME_SECRET=secret-value run ....` output the value of the secret to the log/shell history?

At least that part can be solved via a .env file. I.e.

```
docker --env-file .secrets run ....
```

with `.secrets` containing this:

```
SOME_SECRET=secret-value
```

Granted, that file would need to be written to disk first, but this is a common way to pass along environment variables without exposing them via the command line.

antechrestos:
@tejal29 you're welcome
@cvgw I added it. To do so I added a new dedicated dockerfile. I also added a way to pass environment variable to build (with `cmd.Env` for docker build and `-e` for kaniko build as kaniko build is done inside docker) and also tested the expansion of docker variable.
For this test a check of output is provided te do a research on argument value and ensure that it is not printed.

cmd/executor/cmd/root.go
```diff
@@ -206,6 +207,17 @@ func resolveDockerfilePath() error {
 	return errors.New("please provide a valid path to a Dockerfile within the build context with --dockerfile")
 }
 
+// resolveEnvironmentBuildArgs replace build args without value by the same named environment variable
+func resolveEnvironmentBuildArgs(arguments []string, resolver func(string) string) {
```
cvgw: So this will take an argument like `FOO` and replae it with `FOO=FOO` ?

integration/images.go
```diff
@@ -71,6 +78,36 @@ var additionalKanikoFlagsMap = map[string][]string{
...
 }
 
+// output check to do when building with kaniko
+var outputChecks = map[string]func(string, []byte) error{
+	"Dockerfile_test_arg_secret": checkArgsNotPrinted,
+}
+
+// Checks if argument are not printed in output.
+// Argument may be passed through --build-arg key=value manner or --build-arg key with key in environment
+func checkArgsNotPrinted(dockerfile string, out []byte) error {
```
cvgw: So this checks to make sure the value of none of the build args are logged? It makes no distinction between whether they were added as literals or expanded from the env?

antechrestos: @cvgw I could have make the distinction in the test, however I could not find any way to make the distinction in the code as env like variable (env, arg and meta arg ) are used everywhere as a mere string `key=value` and could not make the distinction without breaking everything.
Do you want I change the test? Or everything?

cmd/executor/cmd/root.go
antechrestos: @cvgw this will take the argument array such as `[ "foo=bar", "EnvVariable" ]` crawl through it and replace any argument without `=` with `"EnvVariable=EnvValue"` **if** a  value was returned by `resolver` function  (`os.Getenv` in prod code, mocked function in tests). If zero length value returned, it will be replaced with `"EnvVariable="`

integration/images.go
cvgw: No changes needed. Was just clarifying for my owner understanding. I think it's reasonable to not log any ARG values.

cmd/executor/cmd/root.go
cvgw: Thanks for explaining, cheers

antechrestos:
@cvgw something went wrong with kaniko. However, things went well in Travis 🤔...

tejal29:
@antechrestos i triggered a rebuilt. Previous failure was a flake.

antechrestos:
@tejal29 thanks

antechrestos:
@tejal29 😢

antechrestos:
@tejal29  I have rebased my branch. I am not sure it might change anything :thinking:

antechrestos:
@cvgw @tejal29 may I have another try with kokoro? :pray:

tejal29:
@antechrestos The travis test passed, so i am going to merge this in!

antechrestos:
@tejal29 at our own risk 😊

sourcec0de:
THANK YOU! ❤️

nhed:
Thanks this is so helpful (albeit undocumented) feature

Issue #992, opened 2020-01-23
Kaniko doesnt support Variant label for images


PhoenixMage:
**Actual behavior**
When using kaniko on an armv7 instance kaniko seems to use the armhf architecture.
If I use docker build then armv7 is correctly used.

**Expected behavior**
Kaniko should use the armv7 architecture.

**To Reproduce**
/kaniko/executor -f Dockerfile --no-push
gives armhf
docker build
gives armv7

**Additional Information**
- Dockerfile
FROM alpine:3.11
RUN apk update --print-arch

 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li>- [ ] </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li>- [Yes ] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [ ] </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [ ] </li></ul>|

PhoenixMage:
Seems runtime.GOARM doesnt exist.
My thoughts were to set a symbol at compile time using the linker but I lack the skills to implement that.

tejal29:
We did merge a #980 which prefers platform you are currently running on. 
I tried adding `Variant` to `goarm`, however i got this error
```
var (
	goarm uint8
)
func currentPlatform() v1.Platform {
	return v1.Platform{
		OS:           runtime.GOOS,
		Architecture: runtime.GOARCH,
+		Variant: string(goarm),
	}
}
```

tejal29:
Forgot to mention, 
I  modified the Makefile to retrieve `GOARM` from the current platform i am building on. 
However it is empty
```
tejaldesai@@kaniko (507)$ make && docker cp /usr/local/google/home/tejaldesai/workspace/kaniko/out/executor c765d295c9cb:/kaniko
GOARCH=amd64 GOARM= GOOS=linux CGO_ENABLED=0 go build -ldflags '-extldflags "-static" -X github.com/GoogleContainerTools/kaniko/pkg/version.version=v0.16.0 -w -s  ' -o out/executor github.com/GoogleContainerTools/kaniko/cmd/executor
tejaldesai@@kaniko (507)$
```

PR #996, opened 2020-01-24
Take a compile time variable from the linker to set arm variant

PhoenixMage:
Fixes issue #992 

**Description**

Takes a compile time variable from the linker to set the Arm variant for pulling from docker hub.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Bug fixes

```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -147,20 +147,20 @@ func SHA256(r io.Reader) (string, error) {
 
 // CurrentPlatform returns the v1.Platform on which the code runs
 func currentPlatform() v1.Platform {
-	var arch_variant string
+	var archVariant string
 	if strings.HasPrefix(runtime.GOARCH, "arm") {
 		switch goarm {
 		case "6":
-			arch_variant="v6"
+			archVariant="v6"
 		case "7":
-			arch_variant="v7"
+			archVariant="v7"
 		case "8":
-			arch_variant="v8"
+			archVariant="v8"
 		}
 	}
 	return v1.Platform{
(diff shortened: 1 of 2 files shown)


pkg/util/util.go
```diff
@@ -144,8 +147,20 @@ func SHA256(r io.Reader) (string, error) {
 
 // CurrentPlatform returns the v1.Platform on which the code runs
 func currentPlatform() v1.Platform {
+	var archVariant string
+	if strings.HasPrefix(runtime.GOARCH, "arm") {
```
tejal29: I am little confused how this works. 
IIUC, we build a Kaniko binary on my local environment. It would contain the `arm` and `archVariant` thats on my machine right?

Does this mean, our Dockerfile should also change?
```
FROM golang:1.12
ARG GOARCH=amd64
```

tejal29:
Thanks you for your contribution @PhoenixMage! Some `gofmt` issues and one question on if we need to change the way image is built.

PhoenixMage:
GOARCH is set at runtime and based on the previous patch that was merged it
resolved the architecture at a high level.

This works by embedding GOARM as a symbol in the code when it is linked
(the variable goarm and the -X in the make file) so the kaniko version is
essentially locked to a variant.

The archVariant is the format docker expects the variant to be in.

That being said I don't believe the docker images need to change as it
should pull the correct base OS, GOARCH will be determined from that and
variant will be locked from compile time.

tejal29:
closed due to inactivity

PR #1000, opened 2020-01-25
whitelist  /tmp/apt-key-gpghome.* directory

### LINKED ISSUE #769: error building image: error building stage: lstat /tmp/apt-key-gpghome.VoPBz66R2g/gnupg_spawn_agent_sentinel.lock: no such file or directory
**Actual behavior**
I am trying to build a Dockerfile which is built correctly with docker daemon.
I have this error : 
``` 
Setting up google-chrome-stable (77.0.3865.75-1) ...
update-alternatives: using /usr/bin/google-chrome-stable to provide /usr/bin/x-www-browser (x-www-browser) in auto mode
update-alternatives: using /usr/bin/google-chrome-stable to provide /usr/bin/gnome-www-browser (gnome-www-browser) in auto mode
update-alternatives: using /usr/bin/google-chrome-stable to provide /usr/bin/google-chrome (google-chrome) in auto mode
Processing triggers for libc-bin (2.24-11+deb9u4) ...
Processing triggers for libgdk-pixbuf2.0-0:amd64 (2.36.5-2+deb9u2) ...
INFO[0278] Taking snapshot of full filesystem...        
INFO[0279] Adding whiteout for /var/lib/apt/lists       
error building image: error building stage: lstat /tmp/apt-key-gpghome.7n7iliD9iR/gnupg_spawn_agent_sentinel.lock: no such file or directory
````

**Expected behavior**
As it works with docker daemon, I expect that the build is also OK with kaniko without changing the dockerfile.

**To Reproduce**
Use this Dockerfile in a builder-node directory
```
FROM google/cloud-sdk:262.0.0-slim

RUN apt-get update && apt-get install --yes curl && \
    curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
    apt-get install -y nodejs
 
RUN echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/chrome.list && \
    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg |  apt-key add - && \
    echo "deb https://dl.yarnpkg.com/debian/ stable main" |  tee /etc/apt/sources.list.d/yarn.list && \
    apt-get update && \
    apt-get install --no-install-recommends -y software-properties-common && \
    # installation des paquets via apt
    apt-get install --no-install-recommends -y --allow-unauthenticated unzip google-chrome-stable yarn build-essential && \
    apt-get clean && \
    rm -rf /etc/apt/sources.list.d/chrome.list /var/lib/apt/lists
```

Run it : 
```
run_in_docker.sh Dockerfile $(pwd)/builder-node gcr.io/myproject/builder-node:kaniko false
```
**Additional Information**
 - no other file in build context than the Dockerfile
```
ls -l builder-node 
total 4
-rw-r--r-- 1 olivier olivier 2354 15 sept. 23:26 Dockerfile
```
 - Kaniko Image built with `make image` from commit `d67a8225031f2809f0d6c8c4f1825f233d8574a1` : 
```
gcr.io/kaniko-project/executor                  latest              af14b1090d2e        2 hours ago         49.1MB
```

### PR DESCRIPTION
**Pull request by tejal29**
Fixes #769 

**Description**
In this PR
- Whitelist can now accept patterns like `/tmp/tdd.*/`
- Add `/tmp/apt-key-gpghome.*` to whitelist since `apt-key add` adds temporary files in this directory. 

**Submitter Checklist**

review them:

- [X] Adds integration tests if needed.

**Reviewer Notes**

**Release Notes**

```
Directories matching `/tmp/apt-key-gpghome.*` pattern are not whitelisted. Dockerfiles which run `apt-key add` will now work.

```

### DIFF
--- a/pkg/snapshot/layered_map.go
+++ b/pkg/snapshot/layered_map.go
@@ -20,11 +20,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/GoogleContainerTools/kaniko/pkg/timing"
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
+	"github.com/sirupsen/logrus"
 )
 
 type LayeredMap struct {
@@ -113,13 +115,18 @@ func (l *LayeredMap) Add(s string) error {
 // from the current layered map by its hashing function.
 // Returns true if the file is changed.
 func (l *LayeredMap) CheckFileChange(s string) (bool, error) {
-	oldV, ok := l.Get(s)
 	t := timing.Start("Hashing files")
 	defer timing.DefaultRun.Stop(t)
 	newV, err := l.hasher(s)
 	if err != nil {
+		// if this file does not exist in the new layer return.
+		if os.IsNotExist(err) {
+			logrus.Tracef("%s detected as changed but does not exist", s)
+			return false, nil
+		}
 		return false, err
 	}
+	oldV, ok := l.Get(s)
 	if ok && newV == oldV {
 		return false, nil
 	}
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -26,7 +26,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"syscall"
 	"time"
@@ -696,7 +695,7 @@ func HasFilepathPrefix(path, prefix string, prefixMatchOnly bool) bool {
 	}
 
 	for index := range prefixArray {
-		m, err := regexp.MatchString(prefixArray[index], pathArray[index])
+		m, err := filepath.Match(prefixArray[index], pathArray[index])
 		if err != nil {
 			return false
 		}
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"syscall"
 	"time"
@@ -60,6 +61,12 @@ var initialWhitelist = []WhitelistEntry{
 		// from the base image
 		Path:            "/etc/mtab",
 		PrefixMatchOnly: false,
+	},
+	{
+		// we whitelist /tmp/apt-key-gpghome, since the apt keys are added temporarily in this directory.
+		// from the base image
+		Path:            "/tmp/apt-key-gpghome",
+		PrefixMatchOnly: true,
 	},
 }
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by tejal29** [comment]
I don't see the need to make this configurable since users running `apt-key add` adds the keys in `tmp`. They get deleted when user runs 
```
apt-get autoclean
```
  or 
```
apt-get clean
```

It is expected these files get cleaned up.

**Comment by cvgw** [comment]
This seems like a reasonable and relatively minor change, but I'm having some trouble understanding it.

My impression is that perhaps `apt-get` will sometimes delete a file in `/tmp` which corresponds to a key of sorts? Perhaps the scenario is similar to the follow example;
```
// layer 0 includes the temporary key for `apt` which is located in /tmp
FROM some-base-image // layer 0
// layer 1 deletes the temp key from layer 0 as part of using `apt`
RUN apt-get update && apt-get install foo // layer 1
```
Because the temp key from layer 0 is removed (opposed to be marked as deleted with a `.wh.` file) the `LayerMap` hits an unexpected error as the file no longer exists.

Is that correct?

**Comment by tejal29** [comment]
> This seems like a reasonable and relatively minor change, but I'm having some trouble understanding it.
> 
> My impression is that perhaps `apt-get` will sometimes delete a file in `/tmp` which corresponds to a key of sorts? Perhaps the scenario is similar to the follow example;
> 
> ```
> // layer 0 includes the temporary key for `apt` which is located in /tmp
> FROM some-base-image // layer 0
> // layer 1 deletes the temp key from layer 0 as part of using `apt`
> RUN apt-get update && apt-get install foo // layer 1
> ```
> 
> Because the temp key from layer 0 is removed (opposed to be marked as deleted with a `.wh.` file) the `LayerMap` hits an unexpected error as the file no longer exists.
> 
> Is that correct?

Yes. usually folks have `apt-key`, `apt-get` and `apt-get clean` all in one run command.
` apt-get update && apt-key add key-from-base-image && apt-get install foo && apt-get clean`

We need to debug more on why 
1. kaniko first detects files add in `/tmp/apt-get-temp-workdir` and  
2. Are the commands still running in background and maybe kaniko detected them as finished
    Hence, files got deleted in between "Taking snapshot" and "Actually adding to layer"

Does that make sense?

**Comment by cvgw** [comment]
I guess I'm still unclear about 
1) Where the "key" is coming from (base image, RUN command, etc)
2) What causes they key to be removed
3) When the key is removed (during `RUN`, after layer has finished, etc.)

My impression from your last comment is that perhaps some of these are currently unknown?

**Comment by tejal29** [comment]
> I guess I'm still unclear about
> 
> 1. Where the "key" is coming from (base image, RUN command, etc)

The keys can come from anywhere. e.g
- in the command below it was from internet.
- it can also be a k8 secret. 
```
RUN echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/chrome.list && \
    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg |  apt-key add - && \
```

> 2. What causes they key to be removed

Usually, `apt-get clean` will cause the key to be removed.
> 3. When the key is removed (during `RUN`, after layer has finished, etc.)
> 
> My impression from your last comment is that perhaps some of these are currently unknown?

yes. we need to debug more on why was it detected as added.

PR #1014, opened 2020-01-30
Allow contributors to launch integration tests against local registry

### LINKED ISSUE #1012: Add documentation about local integration tests
Looking at the [travis file](https://github.com/GoogleContainerTools/kaniko/blob/master/.travis.yml), we can see that there is a way to run integration tests locally. This great feature was brought by [the following pull request](https://github.com/GoogleContainerTools/kaniko/pull/979)

I think this feature quite great and should be documented as it is quite usefull. 

However, looking at how it is done, I found the script that updates the integration docker files and copies the base image from gcr.io to local registry.

If any developer did it on its laptop, it would end with git modified file which is annoying.

It can be solved in several way

1. The light way

keep a save of each updated dockerfiles and use the `trap` command to rollback the changes at the end of the `integration-test.sh` script.

2. The hard way

- add a variable to `Makefile` `INTEGRATION_REGISTRY` valued by default to `gcr.io`
- add the same variable to `integration-test.sh` valued by default to `gcr.io`
- launch the copy from `gcr.io` to target if differents
- either the light way or by using a build argument variable `INTEGRATION_REGISTRY` in each `FROM` command and use it in each `docker` command launched by tests.

I would be glad to implement any chosen solution, even a third one 😄

antechrestos:
Fixes #1012 

**Description**

This change allows contributors to launch integration tests with a local registry.

The idea is to check whether `IMAGE_REPO` parameter and 

- if it starts with `gcr.io`, keep the previous thing
- if it starts with anything else, the program will crawl through dockerfiles, save them as temporary files, import the based `gcr.io` public images and push them to the target repo then launch the test. The rollback of savec files is made as a defer function

Please also note that I isolated the code of test launching in a dedicated function as `os.Exit` and `defer` cannot be mixed (as `os.Exit` exists without calling defered functions).

**Submitter Checklist**

review them:

- [x] Includes [unit tests] **Not needed**
- [x] Adds integration tests  **Not needed**

**Reviewer Notes**

- [x] Unit tests and or integration tests added.


antechrestos:
@tejal29 can I have the kokoro result?
Isn't any way to make him post the result in the discussions? Since it is able to remove a label, it should be able to post a message with an attachment I guess

samos123:
Nice pull request. Significant improvement over how it was before. We're planning to get rid of Kokoro since Travis CI seems to be doing the job well enough and has public build logs. 

I don't know why Kokoro is failing. @tejal29 can you take a look at why Kokoro is failing? this is the short error log: 
```
    --- FAIL: TestRun/test_Dockerfile_test_hardlink (53.80s)
        cmd.go:41: [container-diff diff --no-cache daemon://gcr.io/kaniko-test/docker-dockerfile_test_hardlink gcr.io/kaniko-test/kaniko-dockerfile_test_hardlink -q --type=file --type=metadata --json]
        cmd.go:42: fatal error: concurrent map iteration and map write
```

integration/config.go
```diff
@@ -0,0 +1,34 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import "strings"
+
+type gcpConfig struct {
```
samos123: We could probably change this to testConfig instead of gcpConfig, but no need to do this as part of PR unless you feel like it. I should have done this in my previous PR.

antechrestos: @samos123 I changed it to `integrationTestConfig`

antechrestos:
@samos123 The issue you points must be the one @cvgw [posted](https://github.com/GoogleContainerTools/container-diff/issues/323)

antechrestos:
@samos123 I've submitted [a fix](https://github.com/GoogleContainerTools/container-diff/pull/324)

samos123:
Ah so we should wait for a new container-diff binary that includes that fix right? Either way, I've triggered kokoro just now to run again. I think it's safe to ignore kokoro in this case but let's wait for @cvgw or @tejal29 feedback.

antechrestos:
@samos123 Luckily kokoro run successfully :smile:

integration/integration_test.go
```diff
@@ -80,37 +82,66 @@ func getDockerMajorVersion() int {
...
+		fmt.Println("Missing required tools")
+		os.Exit(1)
+	}
+
+	if dockerfiles, err := FindDockerFiles(dockerfilesPath); err != nil {
+		fmt.Println("Coudn't create map of dockerfiles", err)
+		os.Exit(1)
+	} else {
+		config = initGCPConfig()
+		if exitCode, err := launchTests(m, dockerfiles); err != nil {
```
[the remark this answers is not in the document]
antechrestos: Yes it is simpler like that

antechrestos:
@tejal29 done.

Other point, the integration test *TestRun/test_Dockerfile_test_copy_symlink* has random behaviour. I don't know why but it passed the first time, then I had to repush cause my code was not well formatted (one excessive space), then it failed twice and succeeded the third time.

The two failure were due to 

```
--- FAIL: TestRun/test_Dockerfile_test_copy_symlink (5.85s)

        integration_test.go:208: diff = [

              {

                "Image1": "localhost:5000/docker-dockerfile_test_copy_symlink",

                "Image2": "localhost:5000/kaniko-dockerfile_test_copy_symlink",

                "DiffType": "File",

                "Diff": {

                  "Adds": [

                    {

                      "Name": "/tmp/link",

                      "Size": -1

                    }

                  ],

                  "Dels": null,

                  "Mods": [

                    {

                      "Name": "/tmp",

                      "Size1": 6,

                      "Size2": 11

                    }

                  ]

                }

              },

              {

                "Image1": "localhost:5000/docker-dockerfile_test_copy_symlink",

                "Image2": "localhost:5000/kaniko-dockerfile_test_copy_symlink",

                "DiffType": "Metadata",

                "Diff": {

                  "Adds": [],

                  "Dels": []

                }

              }

            ]

        integration_test.go:211: []integration.diffOutput differ (-got, +want):   []integration.diffOutput{

              	{

              		Image1:   "localhost:5000/docker-dockerfile_test_copy_symlink",

              		Image2:   "localhost:5000/kaniko-dockerfile_test_copy_symlink",

              		DiffType: "File",

              		Diff: &integration.fileDiffResult{

            - 			Adds: []integration.fileDiff{{Name: "/tmp/link", Size: -1}},

            + 			Adds: nil,

              			Dels: nil,

              		},

              	},

              	{Image1: "localhost:5000/docker-dockerfile_test_copy_symlink", Image2: "localhost:5000/kaniko-dockerfile_test_copy_symlink", DiffType: "Metadata", Diff: &integration.metaDiffResult{Adds: []string{}, Dels: []string{}}},

              }
```

antechrestos:
@tejal29 looks like you [fixed the issue](https://github.com/GoogleContainerTools/kaniko/pull/1030). I will rebase my work.

tejal29:
Thank you much!

PR #1016, opened 2020-01-31
fix test flake

tejal29:
Fix unit test flake.


tejal29:
Since in go there is no generic sort function, i don't think its possible.

PR #1022, opened 2020-02-04
[Draft] Do not apply the dockerignore patterns between intermediate stages

### LINKED ISSUE #595: The .dockerignore file also matches between stages
**Actual behavior**
The patterns in `.dockerignore` match not only in workdir but also when copying between stages.

**Expected behavior**
I expected kaniko to have the same semantics as in `docker build`.
The patterns should match only on initial import relative to `/workdir/` not between stages relative to `/`.

**To Reproduce**

```
$ cat Dockerfile 
from busybox
run mkdir /foo/
run touch /foo/bar
run ls -l / /foo/

from busybox
copy --from=0 /foo/ /foo/
run ls -l / /foo/
$ cat .dockerignore 
/foo/
$ docker run --rm -v $PWD:/workspace gcr.io/kaniko-project/executor:v0.9.0 --no-push
[...]
INFO[0007] copy --from=0 /foo/ /foo/                    
INFO[0007] Skipping paths under /kaniko, as it is a whitelisted directory 
INFO[0007] Skipping paths under /dev, as it is a whitelisted directory 
INFO[0007] Skipping paths under /proc, as it is a whitelisted directory 
INFO[0007] Skipping paths under /sys, as it is a whitelisted directory 
INFO[0007] Skipping paths under /workspace, as it is a whitelisted directory 
INFO[0008] run ls -l / /foo/                            
INFO[0008] cmd: /bin/sh                                 
INFO[0008] args: [-c ls -l / /foo/]                     
ls: /foo/: No such file or directory
```
Trying the same with v0.7.0 or `docker build` works fine.

tejal29:
Fixes #595

Only apply `.dockerignore` file for resolving sources
1. For `Add` command
2. For `Copy` command if there is no `from`

Also some performance cleanup.

TODO: cleanup copyDir and copyFile

### DIFF
--- a/pkg/commands/add.go
+++ b/pkg/commands/add.go
@@ -46,7 +46,7 @@ type AddCommand struct {
 func (a *AddCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.BuildArgs) error {
 	replacementEnvs := buildArgs.ReplacementEnvs(config.Env)
 
-	srcs, dest, err := util.ResolveEnvAndWildcards(a.cmd.SourcesAndDest, a.buildcontext, replacementEnvs)
+	srcs, dest, err := util.ResolveEnvAndWildcardsWithIgnore(a.cmd.SourcesAndDest, a.buildcontext, replacementEnvs)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


tejal29:
this was replaced by #1234

PR #1027, opened 2020-02-05
refactor copy.chown code and add more tests

tejal29:
Refactoring code in #962  and adding more tests

### DIFF
--- a/pkg/commands/copy.go
+++ b/pkg/commands/copy.go
@@ -32,6 +32,11 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
+// for testing
+var (
+	getUIDAndGID = util.GetUIDAndGIDFromString
+)
+
 type CopyCommand struct {
 	BaseCommand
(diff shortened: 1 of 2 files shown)


pkg/commands/copy.go
```diff
@@ -126,6 +120,21 @@ func (c *CopyCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bu
 	return nil
 }
 
+func getUserGroup(chownStr string, env []string) (int64, int64, error) {
+	if chownStr == "" {
+		return util.DoNotChangeUID, util.DoNotChangeGID, nil
+	}
+	chown, err := util.ResolveEnvironmentReplacement(chownStr, env, false)
+	if err != nil {
+		return -1, -1, err
```
cvgw: Why -1 and not `util.DoNotChangeUID` ?

tejal29: The return value here should be ignored since `err` is not nil. 
I did not use the `util.DoNotChangeUID` since the return value does not matter.

PR #1042, opened 2020-02-09
Extended test suite

cvgw:
Do not merge

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_issue_506
+++ b/integration/dockerfiles/Dockerfile_test_issue_506
@@ -1,0 +1,2 @@
+FROM ubuntu:18.04
+RUN mkdir -p /var/run/test-foobar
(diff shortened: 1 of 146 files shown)


cvgw:
@samos123 I copied in all of the docker files from jessfraz/dockerfiles that didn't require any context. It looks like the free tier of Travis CI has a maximum job time of 50 minutes https://github.com/travis-ci/travis-ci/issues/6232.

Maybe we could split it up into multiple jobs some how? I'm not sure how that would work with the current integration test code.

It looks like the Kokoro job failed due to lack of disk space which is the same error I hit when trying to run this integration suite on my local machine.

WDYT?

cc @tejal29

tejal29:
> Maybe we could split it up into multiple jobs some how? I'm not sure how that would work with the current integration test code.

+1 on trying to run in this parallel jobs on travis. 
1. unit tests
2. functional tests - (current integration tests) - the ones we have. I feel the ones we have just test functionality.
3. these actual dockerfiles

> It looks like the Kokoro job failed due to lack of disk space which is the same error I hit when trying to run this integration suite on my local machine.

Not sure how we can fix this.  I would like these tests to be run locally too if possible

Issue #495, opened 2018-12-18
Kaniko build failing on Kubernetes with simple node project calling npm install


johnmcollier:
I'm trying to build a basic node application (https://github.com/johnmcollier/nodetest) using Kaniko. Using an NFS volume mount, I've mounted in its Dockerfile and build context under `/workspace`. The arguments I pass the Kaniko container are: 
```
          args: [ "--dockerfile=/workspace/nodetest/Dockerfile", 
                  "--context=/workspace/nodetest", 
                  "--skip-tls-verify",
                  "--cache=true",
                  "--destination=myregistry/nodetest" ]
```

However, the build fails at the `npm install` step, saying:
```
Unhandled rejection Error: EROFS: read-only file system, mkdir '/root/.npm'

npm ERR! cb() never called!

npm ERR! This is an error with npm itself. Please report this error at:
npm ERR!     <https://npm.community>
error building image: error building stage: waiting for process to exit: exit status 1
```

**To Reproduce**
1. Mount the Dockerfile and build context into Kaniko
2. Build the image using `/kaniko/executor`

**Additional Information**
 - Dockerfile:
   ```
   FROM ibmcom/ibmnode

   WORKDIR "/app"

   # Install app dependencies
   COPY package.json /app/
   RUN apt-get update \
   && apt-get dist-upgrade -y \
   && apt-get clean \
   && echo 'Finished installing dependencies'
   RUN cd /app; npm install --production
    
   COPY . /app
    
   ENV NODE_ENV production
   ENV PORT 3000
    
   EXPOSE 3000
   CMD ["npm", "start"]
   ```
 - Build Context:
    See https://github.com/johnmcollier/nodetest
    Can be built with docker with a simple `docker build -t nodetest .`
 - Kaniko Image (fully qualified with digest):
    ```
    Image:         gcr.io/kaniko-project/executor:debug
    Image ID:      docker-pullable://gcr.io/kaniko-project/executor@sha256:23820df3cf2812da0f4c03a25e6bb7d65ad3af11386f49306eec22b0bf7bdbd0

Edit: I've attached the full build logs from the Kaniko pod: 
[kaniko_logs.txt](https://github.com/GoogleContainerTools/kaniko/files/2692312/kaniko_logs.txt)

chickenbeef:
[...]

PR #1043, opened 2020-02-10
Fix example pod.yml to not mount to root

jsravn:
As mounting to root makes `/root` a read only filesystem, which breaks a lot of builds that try to add files to `/root` (like `/root/.npm`).

Fixes https://github.com/GoogleContainerTools/kaniko/issues/495.

**Description**

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Bug fixes

```

### DIFF
--- a/examples/pod.yaml
+++ b/examples/pod.yaml
@@ -11,7 +11,7 @@ spec:
             "--destination=<user-name>/<repo>"] # replace with your dockerhub account
     volumeMounts:
       - name: kaniko-secret
-        mountPath: /root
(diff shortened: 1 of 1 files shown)


jsravn:
It's not clear why `/kaniko/.docker` works but it does :). I suppose `/root/.docker` should work as well.

tejal29:
> `/kaniko/.docker` the


@jsravn `/root/.docker` would work if your build is not adding files to `/root` which is the case in your build.

PR #1046, opened 2020-02-12
add design proposal template

tejal29:
Add design proposal template to continue design discussion on a PR. 
This document will also serve as a record of design change

### DIFF
--- a/docs/design_proposals/design-proposal-template.md
+++ b/docs/design_proposals/design-proposal-template.md
@@ -1,0 +1,72 @@
+# Title
+
+* Author(s): \<your name\>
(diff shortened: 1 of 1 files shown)


cvgw:
@tejal29 Did we want this to show up in the issue templates? I think it is in the wrong directory for that and I think you have to add issue templates via the UI.

This template does not show up when you create an issue

tejal29:
> @tejal29 Did we want this to show up in the issue templates? I think it is in the wrong directory for that and I think you have to add issue templates via the UI.
> 
> This template does not show up when you create an issue

no. i don't want this to be an issue template. This template will be used by us or contributors who propose a new feature.

PR #1048, opened 2020-02-13
design proposal 01: filesystem resolution

### LINKED ISSUE #1047: Design Proposal 01: Filesystem Resolution
Write up and discuss a proposal on how to resolve filesystem paths to address issues such at #1038

cvgw:
Fixes #1047 

**Description**

Design proposal for how to handle resolving the filesystem

**Submitter Checklist**

**Reviewer Notes**

**Release Notes**

### DIFF
--- a/docs/design_proposals/filesystem-resolution-proposal-01.md
+++ b/docs/design_proposals/filesystem-resolution-proposal-01.md
@@ -46,7 +46,7 @@ functionality is out of scope for this proposal.
 
 This search space is currently managed in an inconsistent and somewhat ad-hoc
 way; code that manages the various search dimensions is spread out and
-duplicated. There are also a number of poorly handled edge cases which continue
+duplicated. There are also a number of edge cases which continue
 to cause bugs.
 
 The search space dimensions cannot be reduced or substituted.
@@ -56,8 +56,32 @@ whitelists not respected, and unchanged files added to layers.
 
 ## Design
 
-Resolution of the filesystem and the objects it contains should be handled
-consistently and with a single API.
+During snapshotting resolution of the filesystem and the objects it contains should be handled
+consistently and with a pair of APIs.
+
+### First API
+The first API will generate a list of filepaths to resolve
+
+The API should take a limited set of arguments
+* A filesystem root
+
+The API should return only two arguments
+* A set of filepaths to resolve
+* error
+
+The API will walk the filesystem similar to how Snapshotting of the full
+filesystem is currently handled.
+
+It will optimize the walk by skipping any whitelisted directories.
+
+It will not resolve symlinks.
+
+### Second API
+The second API will resolve filepaths to be added to the layer
+
+Callers of the first API can pass the output to the second API
+
+Some callers of the second API will not use the first API (e.g. Copy commands)
 
 * Callers of this API should not be concerned with the type of object at a given filepath (e.g. symlink or not).
 * Callers of this API should not be concerned with whether a given path is whitelisted.
@@ -65,7 +89,7 @@ consistently and with a single API.
   without further processing.
 
 The API should take a limited set of arguments
-* The root path in the filesystem
+* A list of absolute filepaths to scan
 * The whitelist
 
 The API should return only two arguments
@@ -73,9 +97,18 @@ The API should return only two arguments
   processing
 * error
 
+The API will iterate over the set of filepaths and for each item
+* check whether it is whitelisted; if it is, skip it
+* check whether it is a symlink; if it is, resolve the link ancestor and the
+  target, add the link ancestor to the output, check whether the target is whitelisted and if
+  not add the target to the output
+
 ### Open Issues/Questions
 
 \<Ignore symlinks targeting whitelisted paths?\>
+
+Given some link `/foo/link/bar` whose target is a whitelisted path such as
+`/var/run`, should `/foo/link/bar` be added to the layer?
 
 Resolution: Not Yet Resolved
 
--- a/docs/design_proposals/filesystem-resolution-proposal-01.md
+++ b/docs/design_proposals/filesystem-resolution-proposal-01.md
@@ -1,0 +1,110 @@
+# Filesystem Resolution 01
+
+* Author(s): cgwippern@google.com
+* Reviewers:
+* Date: 2020-02-12
+* Status: [Reviewed/Cancelled/Under implementation/Complete]
+
+Here is a brief explanation of the Statuses
+
+1. Reviewed: The proposal PR has been accepted, merged and ready for
+   implementation.
+2. Under implementation: An accepted proposal is being implemented by actual work.
+   Note: The design might change in this phase based on issues during
+   implementation.
+3. Cancelled: During or before implementation the proposal was cancelled.
+   It could be due to:
+   * other features added which made the current design proposal obsolete.
+   * No longer a priority.
+4. Complete: This feature/change is implemented.
+
+## Background
+
+Kaniko builds Docker image layers as overlay filesystem layers; specifically it
+creates a tar file which contains the entire content of a given layer in the
+overlay filesystem. Each overlay layer corresponds to one image layer.
+
+Overlay filesystems should only contain the objects changed in each layer;
+meaning that if only one file changes between some layer A and some B, layer B
+would only contain a single file (the one that changed).
+
+To accomplish this, Kaniko walks the entire filesystem to discover every object.
+Some of these objects may actually be a symlink to another object in the
+filesystem; in these cases we must consider both the link and the target object.
+
+Kaniko also maintains a set of whitelisted (aka ignored) filepaths. Any object
+which matches one of these filepaths should be ignored by kaniko.
+
+This results in a 3 dimensional search space
+
+* changed relative to previous layer
+* symlink
+* whitelisted
+
+Kaniko must also track which objects are referred to by multiple stages; this
+functionality is out of scope for this proposal.
+
+This search space is currently managed in an inconsistent and somewhat ad-hoc
+way; code that manages the various search dimensions is spread out and
+duplicated. There are also a number of poorly handled edge cases which continue
+to cause bugs.
+
+The search space dimensions cannot be reduced or substituted.
+
+Currently there are a number of bugs around symlinks incorrectly resolved,
+whitelists not respected, and unchanged files added to layers.
+
+## Design
+
+Resolution of the filesystem and the objects it contains should be handled
+consistently and with a single API.
+
+* Callers of this API should not be concerned with the type of object at a given filepath (e.g. symlink or not).
+* Callers of this API should not be concerned with whether a given path is whitelisted.
+* This API should return a set of filepaths which should be added to the layer
+  without further processing.
+
+The API should take a limited set of arguments
+* The root path in the filesystem
+* The whitelist
+
+The API should return only two arguments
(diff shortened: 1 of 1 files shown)


docs/design_proposals/filesystem-resolution-proposal-01.md
```diff
@@ -0,0 +1,110 @@
...
+In a linux shell this raises an error
+```
+$ ls /usr/lib/bar
+=> /usr/lib/bar/foo.txt
+$ ln -s /usr/lib/bar barlink
+$ ln -s /usr/lib/bar/foo.txt barlink/foo.txt
+=> ERROR
+```
+
+Given some path `/usr/foo/bar` which is a link to `/dev/null`, and `/dev` is
```
tejal29: What happens when a docker files has a command
`Copy some_bin /var/run/some_bin`

`/var/run` is whitelisted.
In this scenario, `Docker` would create `/var/run/some_bin` in the layer.

docs/design_proposals/filesystem-resolution-proposal-01.md
```diff
@@ -0,0 +1,110 @@
...
+consistently and with a single API.
+
+* Callers of this API should not be concerned with the type of object at a given filepath (e.g. symlink or not).
+* Callers of this API should not be concerned with whether a given path is whitelisted.
+* This API should return a set of filepaths which should be added to the layer
+  without further processing.
+
+The API should take a limited set of arguments
+* The root path in the filesystem
+* The whitelist
```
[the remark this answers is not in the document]
cvgw: I believe you are referring to this block of code
https://github.com/GoogleContainerTools/kaniko/blob/ae4e7d91aa6e825f6ac42d9dc085fda8d109cfd0/pkg/executor/build.go#L381

which as far as I can tell is currently only triggered by an uncached `RUN` command
https://github.com/GoogleContainerTools/kaniko/blob/ae4e7d91aa6e825f6ac42d9dc085fda8d109cfd0/pkg/commands/run.go#L137

So it seems the far more common case is passing a list of files.

Seems like we should actually separate this into two APIs. One that generates a list of files, one that filters/transforms the list of files.

docs/design_proposals/filesystem-resolution-proposal-01.md
cvgw: I'm not sure, what do we currently do when the destination of a copy command is a whitelisted path?

docs/design_proposals/filesystem-resolution-proposal-01.md
```diff
@@ -0,0 +1,110 @@
...
+
+The API should return only two arguments
+* A set of filepaths which should be added to the layer without further
+  processing
+* error
+
+### Open Issues/Questions
+
+\<Ignore symlinks targeting whitelisted paths?\>
```
cvgw: I guess this is related to copy question as well https://github.com/GoogleContainerTools/kaniko/pull/1048/files#r378600747

If a symlink targets a whitelisted path, should we ignore that symlink?

If the destination of a copy command is a whitelisted path, should we ignore that command?

docs/design_proposals/filesystem-resolution-proposal-01.md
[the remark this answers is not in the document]
cvgw: I don't know the full context on why some commands use the files api and why some scan from the root of the filesystem. I would guess it's an optimization for commands where we know for a fact which files were involved in the command. In the case of something like a RUN command its impossible to know which file were involved without scanning the filesystem.

This does present a challenge as the logic each command uses to generate the list of files it passes to the API is slightly different. Trying to unify, or handle any variation would be of great benefit.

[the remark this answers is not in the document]
cvgw: Any example of the comment above; in `CopyCommand` is `dest` is a symlink we only return the target in the set of files to snapshot; should we also be returning the link?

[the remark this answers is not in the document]
tejal29: > I would guess it's an optimization for commands where we know for a fact which files were involved in the command. 

+1. its for efficiency. There is no explanation in the design document (internal only) go/kaniko-design-doc.

docs/design_proposals/filesystem-resolution-proposal-01.md
tejal29: Right now, for Copy and ADD command we do the following
- Resolve source and dest by substituting any env varaibles
- Resolve any wild cards in sources
- Get Relative source file path respect to root path. In this step, 
   - we apply [`CheckWhitelist(path) && !HasFilepathPrefix(path, root, false)`](https://github.com/GoogleContainerTools/kaniko/blob/f3b2c4064be52706f892799d762c446541dcc378/pkg/util/fs_util.go#L439)
   - if `/var/run` is ignored, then this condition should return true for `/var/run/some_bin`
   - if `/var/run` is ignores then this condition should return false for `var/run/another_dir/some_bin`

We should simplify this.

For symlink dir in whitelisted path, there is another flow.

docs/design_proposals/filesystem-resolution-proposal-01.md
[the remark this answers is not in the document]
tejal29: Lets for now keep this proposal simple. We are trying to tackle the problem which is an API for a given path, which will give you a list of files which should not need any further processing. 

We should look into `Copy` and `ADD` and see if parts of these functionalities could be reduced.

docs/design_proposals/filesystem-resolution-proposal-01.md
tejal29: Lets leave looking into `Copy` and `Add` from this proposal and only focus on API for snapshotting entire file snapshot. 

For the snapshot API, we *should not* ignore whitelisted paths.
e.g. An image 'gcr.io/google/toolprovider" provides binary `/usr/bin/tool_binary` which is symlinked to whitelisted path `var/run/tool_binary`, the right thing to do is create a file `/usr/bin/tool_binary` with content of `/var/run/tool_binary`.
We should ignore `/var/run/tool_binary` from the snapshot.
Does that make sense to you?

docs/design_proposals/filesystem-resolution-proposal-01.md
```diff
@@ -0,0 +1,110 @@
...
+
+Resolution of the filesystem and the objects it contains should be handled
+consistently and with a single API.
+
+* Callers of this API should not be concerned with the type of object at a given filepath (e.g. symlink or not).
+* Callers of this API should not be concerned with whether a given path is whitelisted.
+* This API should return a set of filepaths which should be added to the layer
+  without further processing.
+
+The API should take a limited set of arguments
```
tejal29: Can you write this a function signature and lock down the API?

docs/design_proposals/filesystem-resolution-proposal-01.md
```diff
@@ -0,0 +1,110 @@
...
+to cause bugs.
+
+The search space dimensions cannot be reduced or substituted.
+
+Currently there are a number of bugs around symlinks incorrectly resolved,
+whitelists not respected, and unchanged files added to layers.
+
+## Design
+
+Resolution of the filesystem and the objects it contains should be handled
```
tejal29:
```suggestion
Resolution of the filesystem and the objects it contains when taking file system snapshot should be handled
```

docs/design_proposals/filesystem-resolution-proposal-01.md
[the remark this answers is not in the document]
cvgw: >Lets for now keep this proposal simple. We are trying to tackle the problem which is an API for a given path, which will give you a list of files which should not need any further processing.

The API also needs to support a list of files as we mentioned. The generation of that list of files is also important. E.G what happens if a copy command only returns the targets of symlinks and not the link paths themselves?

docs/design_proposals/filesystem-resolution-proposal-01.md
cvgw: It sounds like you are proposing a new behavior, which is to not add symlinks to the layer tar, but instead to create regular files with the content of the symlink target (basically what you mentioned the other day). Is that correct?

docs/design_proposals/filesystem-resolution-proposal-01.md
cvgw: So what is the end result? Do we copy the file to the whitelist destination?

docs/design_proposals/filesystem-resolution-proposal-01.md
cvgw: Once we come to agreement about the semantics of the API I can do that

docs/design_proposals/filesystem-resolution-proposal-01.md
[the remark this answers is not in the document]
cvgw: What are the dimensions the filtering logic need to handle:
* symlink/not-symlink
* whitelisted, not whitelisted

For the case of a symlink; what should be returned in the final output
There is the link, the target, and possibly the "real" link
IE
```
ln -s /foo/bar /baz/boom

touch /foo/bar/zoom.txt
```

link is `/baz/boom/zoom.txt`
target is `/foo/bar/zoom.txt`
the "real" link is `/baz/boom`

If my recent investigation is correct adding `/baz/boom/zoom.txt` to the layer tar will break the filesystem because the existing link `/bar/boom` makes `/bar/boom/zoom.txt` invalid.

docs/design_proposals/filesystem-resolution-proposal-01.md
cvgw: I did a quick test
```
FROM ubuntu:16.04
COPY meow.txt /var/run
```
docker
```
ls /var/run
=> initctl  lock  log  meow.txt  mount  sendsigs.omit.d  shm  systemd  user  utmp
```
kaniko
```
ls /var/run
/var/run
```

cvgw: Interestingly

I did a quick test
```
FROM ubuntu:16.04
COPY meow.txt /var/run/meow.txt
```
docker
```
ls /var/run
=> initctl  lock  log  meow.txt  mount  sendsigs.omit.d  shm  systemd  user  utmp
```
kaniko
```
ls /var/run
=> /var/run/meow.txt
```

docs/design_proposals/filesystem-resolution-proposal-01.md
tejal29: yes. Docker does the same. It does not create the symlink.

cvgw: that seems reasonable, but also perhaps out of scope for this proposal (specifically how a file is added to the tar).

As far as this proposal specifically, we need to consider whether to ignore paths which are symlinks that target whitelisted objects.

It sounds like you are suggesting, no we should not ignore them.

docs/design_proposals/filesystem-resolution-proposal-01.md
```diff
@@ -0,0 +1,167 @@
...
+API?
+
+Resolution: Not Yet Resolved
+
+\<Should the API handle diff'ing files?\>
+
+The proposal currently states that the list of files returned from the API
+should be immediately added to the layer, but this would imply that diff'ing
+existing files, finding newly created files, and handling deleted files would
+have already been done. It may be advantageous to handle these outside of the
```
tejal29: +1

Issue #1038, opened 2020-02-07
Kaniko-Built Image Cannot Be Pulled from Repo Due To Symlink Error


vguaglione:
**Actual behavior**
kaniko-built image that has been pushed to a repository is not extracted correctly upon download, but the same docker-built image that has been pushed to a repository is extracted correctly upon download. 

**Expected behavior**
Image built with kaniko should be extracted correctly when pulled from a repository.

**To Reproduce**
1. Build the container using the dockerfile described below and push to a container registry (can be any)
2. Pull the image using a "docker pull" command
3. Error generated on pull

Output provided below is from 2 runs of the pull command, first, to pull the docker-built version, then to pull the kaniko-built version:

RESULTS OF PULL WHEN DOCKER-BUILT:

➜  docker pull myrepo/dihi-health-intelligence/mlaas/mlaas-ui:363168a0
363168a0: Pulling from dihi-health-intelligence/mlaas/mlaas-ui
d318c91bf2a8: Pull complete 
e78b80ba2df3: Pull complete 
57e393e12bdc: Pull complete 
4cd4d3e2db54: Pull complete 
4d2e313894a7: Pull complete 
Digest: sha256:5a8c044582b7551bdedf91b62472523a84d364b169e3136619ca8f991bd7004d
Status: Downloaded newer image for myrepo/dihi-health-intelligence/mlaas/mlaas-ui:363168a0
myrepo/dihi-health-intelligence/mlaas/mlaas-ui:363168a0

RESULTS OF PULL WHEN KANIKO-BUILT:

➜  docker pull myrepo/dihi-health-intelligence/mlaas/mlaas-ui:0ae48f4b
0ae48f4b: Pulling from dihi-health-intelligence/mlaas/mlaas-ui
d318c91bf2a8: Pull complete
4ac345e3547d: Extracting [==================================================>]  353.7MB/353.7MB
b2a1d0b62606: Download complete 
db78db57c5e7: Download complete 
4c4383ef3686: Download complete 
failed to register layer: Error processing tar file(exit status 1): symlink libBrokenLocale-2.30.so /lib64/libBrokenLocale.so.1: no such file or directory

Two users are currently reporting this problem.  One of the users who is currently affected has provided us with the following information:

"[12:22 PM] SB
i get the same error. i'm looking into whether it has something to do with the part of the dockerfile that the error seems to be pointing at (making symlinks to /proc/self/fd/1 and 2)

[12:24 PM] SB
[...]

Issue #1024, opened 2020-02-04
Our images now fail to run with OCI error


webmutation:
**Actual behavior**
The new images return  Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"/usr/bin/java\": stat /usr/bin/java: no such file or directory": unknown

**Expected behavior**
Previously it run without issue. No OCI runtime

**To Reproduce**
1. Build image with Kaniko
2. Try to run image docker run
**Additional Information**
FROM openjdk:8-jre-slim
 
# Expose ports to enable running the service
# Ports should be standardized to make it easier to debug
# Exposing two services in the same port can create conflicts
 
ENV PORT 8080
EXPOSE 8080
 
# List of ARGS input from Kaniko Build
ARG IMAGE_DATE
ARG VCS_REVISION
ARG VCS_SEMVER
ARG PKG_WORKDIR
 
# Labeling based on https://github.com/opencontainers/image-spec/blob/master/annotations.md
LABEL org.opencontainers.image.created="${IMAGE_DATE}"              \
      org.opencontainers.image.revision="${VCS_REVISION}"           \
      org.opencontainers.image.version="${VCS_SEMVER}"              \
      org.opencontainers.image.title="mytitle"                      \
      org.opencontainers.image.description="mydescription"          \
      org.opencontainers.image.authors="myauthors"                  \
      org.opencontainers.image.vendor="myvendor"                    \
      org.opencontainers.image.url="myurl"                          \
      org.opencontainers.image.documentation="mydocumentationlink"  \
      org.opencontainers.image.source="mygitrepourl"
 
# Copy of distribution/target folder artifacts
# In case additional Artifacts are required
 
# All containers should run in least privileged mode, meaning not ROOT.
# NOTE: On OpenShift there is a warning when you try to run as ROOT
RUN addgroup -g 1001 -S cc && \
    adduser -u 1001 -S -G cc cc && \
    chown -R 1001:0 /home/cc && \
    chmod -R g=u /home/cc
     
COPY --chown=1001:0 ${PKG_WORKDIR}/target/*.jar /home/cc/service.jar
 
USER 1001
# Command to initialize the service
CMD ["/usr/bin/java", "-jar", "home/cc/service.jar"]

RemcodM:
I have similar problems, but they already occur when performing a `docker pull` on a freshly build image with kaniko debug-v0.17.0.

[...]

PR #1069, opened 2020-02-20
Resolve filepaths before scanning for changes

cvgw:
**Description**

Implement the new filepath resolver described in https://github.com/GoogleContainerTools/kaniko/blob/d31d2bc74c884ff5d7932b1f6e7209c6fc63cc5e/docs/design_proposals/filesystem-resolution-proposal-01.md

One change to note is that ancestor directories will no longer be added regardless of whether they are changed. The ancestor directory logic now occurs before the change checking, so ancestor directories will only be added if they have changed.

**Submitter Checklist** 

review them:

**Reviewer Notes**

### DIFF
--- a/pkg/filesystem/resolve.go
+++ b/pkg/filesystem/resolve.go
@@ -125,8 +125,10 @@ func resolveSymlinkAncestor(path string) (string, error) {
 	if !filepath.IsAbs(path) {
 		return "", errors.New("dest path must be abs")
 	}
+
 	last := ""
 	newPath := path
+
 loop:
 	for newPath != "/" {
 		fi, err := os.Lstat(newPath)
@@ -134,11 +136,15 @@ loop:
 			return "", errors.Wrap(err, "failed to lstat")
 		}
 
-		switch mode := fi.Mode(); {
-		case mode&os.ModeSymlink != 0:
+		if util.IsSymlink(fi) {
 			last = filepath.Base(newPath)
 			newPath = filepath.Dir(newPath)
-		default:
+		} else {
+			// Even if the filenode pointed to by newPath is a regular file,
+			// one of its ancestors could be a symlink. We call filepath.EvalSymlinks
+			// to test whether there are any links in the path. If the output of
+			// EvalSymlinks is different than the input we know one of the nodes in the
+			// the path is a link.
 			target, err := filepath.EvalSymlinks(newPath)
 			if err != nil {
 				return "", err
--- a/pkg/filesystem/resolve.go
+++ b/pkg/filesystem/resolve.go
@@ -1,0 +1,158 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package filesystem
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/GoogleContainerTools/kaniko/pkg/util"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// ResolvePaths takes a slice of file paths and a slice of whitelist entries. It resolve each
+// file path according to a set of rules and then returns a slice of resolved paths or error.
+// File paths are resolved according to the following rules:
+// * If path is whitelisted, skip it.
+// * If path is a symlink, resolve it's ancestor link and add it to the output set.
+// * If path is a symlink, resolve it's target. If the target is not whitelisted add it to the
+// output set.
+// * Add all ancestors of each path to the output set.
+func ResolvePaths(paths []string, wl []util.WhitelistEntry) (pathsToAdd []string, err error) {
+	logrus.Info("Resolving paths")
+
+	fileSet := make(map[string]bool)
+
+	for _, f := range paths {
+		// If the given path is part of the whitelist ignore it
+		if util.IsInProvidedWhitelist(f, wl) {
+			logrus.Debugf("path %s is whitelisted, ignoring it", f)
+			continue
+		}
+
+		link, e := resolveSymlinkAncestor(f)
+		if e != nil {
+			err = e
+			return
+		}
+
+		if f != link {
+			logrus.Tracef("updated link %s to %s", f, link)
+		}
+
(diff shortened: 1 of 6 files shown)


pkg/snapshot/snapshot_test.go
```diff
@@ -63,19 +63,24 @@ func TestSnapshotFSFileChange(t *testing.T) {
...
 	}
-	for _, dir := range util.ParentDirectoriesWithoutLeadingSlash(fooPath) {
-		snapshotFiles[dir] = ""
-	}
-	for _, dir := range util.ParentDirectoriesWithoutLeadingSlash(batPath) {
-		snapshotFiles[dir] = ""
-	}
-	numFiles := 0
+
+	// Their parents didn't change so they shouldn't be included
```
samos123: I would prefer to remove the code, we can always recover it from git.

pkg/snapshot/snapshot_test.go
```diff
@@ -155,17 +160,20 @@ func TestSnapshotFSChangePermissions(t *testing.T) {
 	snapshotFiles := map[string]string{
 		batPathWithoutLeadingSlash: "baz2",
 	}
-	for _, dir := range util.ParentDirectoriesWithoutLeadingSlash(batPath) {
-		snapshotFiles[dir] = ""
-	}
-	numFiles := 0
+
+	// The parents haven't changed so they shouldn't be in the tar
+	//for _, dir := range util.ParentDirectoriesWithoutLeadingSlash(batPath) {
```
samos123: same here, remove the code instead of commenting it out

pkg/filesystem/resolve.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// filesWithParentDirs returns every ancestor path for each provided file path.
+// I.E. /foo/bar/baz/boom.txt => [/, /foo, /foo/bar, /foo/bar/baz, /foo/bar/baz/boom.txt]
+func filesWithParentDirs(files []string) []string {
+	filesSet := map[string]bool{}
+
+	for _, file := range files {
+		file = filepath.Clean(file)
+		filesSet[file] = true
```
tejal29: you can skip traversing if fileSet[file] = true
```suggestion
		if _, ok := fileSet[file] ; !ok {
		   filesSet[file] = true
		   for _, dir := range util.ParentDirectories(file) {
			dir = filepath.Clean(dir)
			filesSet[dir] = true
		   }
		 }
```

pkg/filesystem/resolve.go
```diff
@@ -0,0 +1,158 @@
...
+
+		for _, dir := range util.ParentDirectories(file) {
+			dir = filepath.Clean(dir)
+			filesSet[dir] = true
+		}
+	}
+
+	newFiles := []string{}
```
tejal29: We can optimize this by allocation memory of expected size
```
newFiles := make([]string, 0, len(filesSet))
```

pkg/snapshot/snapshot.go
```diff
@@ -149,19 +155,42 @@ func (s *Snapshotter) scanFullFilesystem() ([]string, []string, error) {
 	)
 	timing.DefaultRun.Stop(timer)
 
+	filesToResolve := make([]string, 0, len(memFs))
+	for file := range memFs {
+		if strings.HasPrefix(file, "/tmp/dir") {
```
tejal29: is this intentional change?

pkg/filesystem/resolve.go
```diff
@@ -0,0 +1,158 @@
...
+	if !filepath.IsAbs(path) {
+		return "", errors.New("dest path must be abs")
+	}
+	last := ""
+	newPath := path
+loop:
+	for newPath != "/" {
+		fi, err := os.Lstat(newPath)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to lstat")
```
tejal29: does the err contain the file name? if not, please add the file name

pkg/filesystem/resolve.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		switch mode := fi.Mode(); {
+		case mode&os.ModeSymlink != 0:
+			last = filepath.Base(newPath)
```
tejal29: you can use `util.IsSymlink` here instead.

pkg/filesystem/resolve.go
```diff
@@ -0,0 +1,158 @@
...
+			newPath = filepath.Dir(newPath)
+		default:
+			target, err := filepath.EvalSymlinks(newPath)
+			if err != nil {
+				return "", err
+			}
+
+			if target != newPath {
+				last = filepath.Base(newPath)
```
tejal29: shd this be target?

pkg/filesystem/resolve.go
```diff
@@ -0,0 +1,158 @@
...
+		fi, err := os.Lstat(newPath)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to lstat")
+		}
+
+		switch mode := fi.Mode(); {
+		case mode&os.ModeSymlink != 0:
+			last = filepath.Base(newPath)
+			newPath = filepath.Dir(newPath)
+		default:
```
tejal29: I am a little bit lost here and probably we need to add tests here. 
say a/b/foo.txt is a link pointing to c/d/bar.txt.

In this case, we need to follow the symlink for `a/b/foo.txt` and in your case the `default` block follows the symlink for `a/b/`

Is my analysis correct?

pkg/filesystem/resolve.go
cvgw: yes, lstat errors always contain the file name

pkg/filesystem/resolve.go
cvgw: I don't believe so. We want to return the link, not the target.

pkg/snapshot/snapshot.go
cvgw: Nope, left over cruft from debugging! Thanks for catching

pkg/snapshot/snapshot_test.go
cvgw: Yes, plan is to remove after testing it finished

pkg/filesystem/resolve.go
cvgw: The function `filesWithParentDirs` was copy/pasted from the util pkg. If you think it won't clutter the change I'm happy to take this opportunity to make these updates.

pkg/filesystem/resolve.go
cvgw: In your example the default block would not be hit as `a/b/foo.txt` is a symlink (as opposed to one of it's ancestors) so the return of this function would be `a/b/foo.txt`

If, for example `a/b/` was a link to `c/d/` then `a/b/bar.txt` would return `a/b/`

Adding tests

tejal29: ah got it! Thanks. the `if else` makes it more clear.

cvgw:
related issues #1039 #1038 #1024

Issue #764, opened 2019-09-09
ARG Variables are substituted with empty string when using cache


BenHizak:
**Actual behavior**
Kaniko substitutes ARG with `(empty string)` when using cache. When not using cache, the behavior is normal.

**Expected behavior**
Kaniko should substitute arguments.

**To Reproduce**
See reproducing steps here: https://gist.github.com/BenHizak/4632b3df40892a46712f29b4cf4a6057

Using the following Kaniko image
`gcr.io/kaniko-project/executor@sha256:9000801241318d6c9b8c2b505974f22fd74c57e0adfdb09c98870aef4923bb65`

donmccasland:
Probably just log output issue.  Need to make sure we filter credential based ENV vars.

tejal29:
Unfortunately, we don't evaluate the `ARG` and hence the arg substitution is empty.

cvgw:
@tejal29 I'm assuming we will never log the value of ARG because they may contain sensitive data like SSH keys. WDYT?

BenHizak:
Why not let users decide? it makes it difficult to see important elements
of the CI pipeline. Is there a way to denote specific ARGs to show?

tejal29:
> @tejal29 I'm assuming we will never log the value of ARG because they may contain sensitive data like SSH keys. WDYT?

No, we never added the ARG value to `confif.ENV` which ends up replacing all the places which use the arg. 

I added a PR to fix this. 

You can now see the arg in log
```
 # /kaniko/executor -f dockerfiles/Dockerfile1 --context=dir://workspace --destination=gcr.io/tejal-test/test --tarPath=image.tar
INFO[0000] Resolved base name python:${VERSION} to python: 
INFO[0000] Using dockerignore file: /workspace/.dockerignore 
INFO[0000] Resolved base name python:${VERSION} to python: 
INFO[0000] Retrieving image manifest python:3.7.2       
INFO[0001] Retrieving image manifest python:3.7.2       
INFO[0002] Built cross stage deps: map[]                
INFO[0002] Retrieving image manifest python:3.7.2       
INFO[0002] Retrieving image manifest python:3.7.2   
```

tejal29:
> @tejal29 I'm assuming we will never log the value of ARG because they may contain sensitive data like SSH keys. WDYT?

[...]

PR #1084, opened 2020-02-28
add args to config.Env so following RUN commands can use it

### LINKED ISSUE #713: Kaniko's --build-arg accepts only literal values

tejal29:
Fixes #713, #764 
Add args in docker command `ARG`  to `config.Env` so they get evaluated later in the docker file.

### DIFF
--- a/pkg/commands/arg.go
+++ b/pkg/commands/arg.go
@@ -36,7 +36,8 @@ func (r *ArgCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 	}
 
 	buildArgs.AddArg(key, val)
-	return nil
+	newEnvs := []instructions.KeyValuePair{{Key: key, Value: *val}}
+	return util.UpdateConfigEnv(newEnvs, config, config.Env)
 }
 
 func ParseArg(key string, val *string, env []string, ba *dockerfile.BuildArgs) (string, *string, error) {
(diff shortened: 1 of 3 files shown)


tejal29:
This is not required. All command use `buildArgs.ReplacementEnv` and then replace.

PR #1085, opened 2020-02-28
remove build args from composite key and replace all build args

### LINKED ISSUE #1008: Is there a way not to include build-args in the cache key?
We are trying to use the kaniko executor in cloud build. We pass in a build timestamp and build commit sha into our dockerfile as build-args. Since these change with every build, the cache keys change every build even though the contents of the layers are the same. This causes a 100% cache miss rate. Is there any way around this? We would like to be able to pass in build args that change every build but still use kaniko cache.

tejal29:
Fixes #1008

**Description**

In this change, we remove all the `build-args` being added to composite key. 

With `--cache` flag true, kaniko will try to optimize the builds. It will execute all the metadata command like "ENV", "ARG" and update the `config.Env` to find  dependencies between stage. 
It discards the updated config changes after the optimize phase. 

In this PR, I added a change
1. remove builds args seeded to composite cache key.
2. Replace the command string with any ARG/ENV command which got executed before. 

This will ensure, only layers which actually use the `ARG` will change if `ARG` value changes. 

Added tests for the same.
**Submitter Checklist**

review them:

- [X ] Adds integration tests if needed.

**Reviewer Notes**

**Release Notes**

```
`--build-args` will now be added to layer cache key only if it is used in the dockerfile command for that layer.
```

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -72,7 +72,7 @@ type stageBuilder struct {
 	stageIdxToDigest map[string]string
 	snapshotter      snapShotter
 	layerCache       cache.LayerCache
-	pushCache        cachePusher
+	pushLayerToCache cachePusher
 }
 
 // newStageBuilder returns a new type stageBuilder which contains all the information required to build the stage
@@ -115,7 +115,7 @@ func newStageBuilder(opts *config.KanikoOptions, stage config.KanikoStage, cross
 		layerCache: &cache.RegistryCache{
 			Opts: opts,
 		},
-		pushCache: pushLayerToCache,
+		pushLayerToCache: pushLayerToCache,
 	}
 
 	for _, cmd := range s.stage.Commands {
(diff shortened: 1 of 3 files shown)


pkg/executor/build.go
```diff
@@ -201,7 +207,7 @@ func (s *stageBuilder) optimize(compositeKey CompositeCache, cfg v1.Config) erro
 			return errors.Wrap(err, "failed to get files used from context")
 		}
 
-		compositeKey, err = s.populateCompositeKey(command, files, compositeKey)
+		compositeKey, err = s.populateCompositeKey(command, files, compositeKey, s.args, cfg.Env)
```
cvgw: It seems like adding `s.args` to the cache key would result in the same problem we have right now where all of the args are in the cache key all the time.

cvgw: I'm confused about the difference between `StageBuilider.args` and `StageBuilder.opts.args`

cvgw: Additionally, do we need to add the args if we've already populated the command with them?

cvgw:
Would you mind expanding the `Test_stageBuilder_build` tests to cover this change by asserting that the cache key received by the mock only contain the correct values. Thanks

pkg/executor/build.go
tejal29: > It seems like adding `s.args` to the cache key would result in the same problem we have right now where all of the args are in the cache key all the time.

In this PR, the args are not added to the cache key. They are used to replace args in the command. The key is constructed using only the evaluated command [here](https://github.com/GoogleContainerTools/kaniko/blob/ae5a7bf1586d10e95e4f7a5df56148bb8e079908/pkg/executor/build.go#L160)

tejal29: > I'm confused about the difference between `StageBuilider.args` and `StageBuilder.opts.args`

`StageBuilder.args` are union of  `KanikoOptions.args` i.e. `StageBuilder.opts.args` and any arg variables added in that stage with `ARG` command.
https://github.com/GoogleContainerTools/kaniko/blob/ae5a7bf1586d10e95e4f7a5df56148bb8e079908/pkg/executor/build.go#L132

tejal29: > Additionally, do we need to add the args if we've already populated the command with them?

We don't add the key. The line you are looking at is a different func `optimize` and not part of `populateCompositeKey`

cvgw: > > It seems like adding `s.args` to the cache key would result in the same problem we have right now where all of the args are in the cache key all the time.
> 
> In this PR, the args are not added to the cache key. They are used to replace args in the command. The key is constructed using only the evaluated command [here](https://github.com/GoogleContainerTools/kaniko/blob/ae5a7bf1586d10e95e4f7a5df56148bb8e079908/pkg/executor/build.go#L160)

Ahh ok, I see why I was confused now. Might be more clear to move the command resolve out of `populateCompositeKey`, but might be easier for maintenance to keep it in there.

tejal29:
@cvgw Please take another look!

PR #1103, opened 2020-03-04
fixing docker run command in README.md

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

**Description**

İf you try to run this command you got an error like this :

> docker: Error response from daemon: source is not directory.
See 'docker run --help'.

so , i fixed this command to working properly.


developer-guy:
@googlebot I signed it!

PR #1113, opened 2020-03-06
 fix resolve link for dirs with trailing /

### LINKED ISSUE #1112: Resolve paths does not resolve correctly for paths with "/"
A GCB user reported this bug, 

```
FROM node:lts-slim
WORKDIR /var/www/ <-- THIS TRAILING SLASH 

```

resulted into this error

```
error building image: error building stage: failed to take snapshot: unable to add file /var/www/build/build to layered map: error creating hash for /var/www/build/build: lstat /var/www/build/build: no such file or directory
```

tejal29:
Fixes #1112

In this PR, 
- wrap errors wherever possible.
- Use `filepath.Clean` to remove trailing '/' from path before recursively finding ancestors. 
   filepath.Dir("/foo/bar/baz/") -> returns "foo/bar/baz"
- Added a unit test with the fix. Without `filepath.Clean(path)`, the test fails
```
Test_resolveSymlinkAncestor$/^dir_ends_with_$/^_is_not_a_symlink$ #gosetup
=== RUN   Test_resolveSymlinkAncestor
=== RUN   Test_resolveSymlinkAncestor/dir_ends_with_/_is_not_a_symlink
--- FAIL: Test_resolveSymlinkAncestor (0.00s)
    --- FAIL: Test_resolveSymlinkAncestor/dir_ends_with_/_is_not_a_symlink (0.00s)
        resolve_test.go:255: expected result to be /tmp/830246385/var/www not /tmp/830246385/var/www/www
FAIL

```

See https://play.golang.org/p/TAjN36H796z
```
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
	fmt.Println("On Unix:")

	fmt.Println(filepath.Dir("/foo/bar/baz"))
	fmt.Println(filepath.Dir("/foo/bar/baz/"))
	fmt.Println(filepath.Dir(filepath.Clean("/foo/bar/baz/")))

}
```

Returns:
```
On Unix:
/foo/bar
/foo/bar/baz
/foo/bar
```

### DIFF
--- a/pkg/filesystem/resolve_test.go
+++ b/pkg/filesystem/resolve_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package filesystem
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


pkg/filesystem/resolve_test.go
```diff
@@ -234,6 +235,27 @@ func Test_resolveSymlinkAncestor(t *testing.T) {
 		}
 	})
 
+	t.Run("dir ends with / is not a symlink", func(t *testing.T) {
```
cvgw: Add a test for "dir ends with / is a symlink" ?

PR #1134, opened 2020-03-14
Add support to --chown flag to ADD command (Issue #57)

### LINKED ISSUE #57: Support for --chown flag in ADD command

dani29:
Fixes #57 
**Description**

Docker supports --chown flag on the ADD command, similarly to the COPY command (which is already implemented by Kaniko). The ADD command supports remote sources (URLs) and unpacking tarballs, whereas the COPY command only copies files from local context. 

For local files, the ADD implementation relies on the COPY implementation, so it's just a matter of passing the Chown config. The PR adds support for remote files, whereas --chown does not affect extracted tar files.

I've moved the `GetUserGroup()` method to `command_util.go` as it's now used by both COPY and ADD, hope it makes more sense. 
**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Add support to --chown flag to ADD command
```

### DIFF
--- a/pkg/commands/add.go
+++ b/pkg/commands/add.go
@@ -47,6 +47,11 @@ type AddCommand struct {
 func (a *AddCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.BuildArgs) error {
 	replacementEnvs := buildArgs.ReplacementEnvs(config.Env)
 
+	uid, gid, err := util.GetUserGroup(a.cmd.Chown, replacementEnvs)
+	if err != nil {
+		return errors.Wrap(err, "getting user group from chowm")
+	}
+
 	srcs, dest, err := util.ResolveEnvAndWildcards(a.cmd.SourcesAndDest, a.buildcontext, replacementEnvs)
 	if err != nil {
 		return err
@@ -66,7 +71,7 @@ func (a *AddCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 				return err
 			}
 			logrus.Infof("Adding remote URL %s to %s", src, urlDest)
-			if err := util.DownloadFileToDest(src, urlDest); err != nil {
+			if err := util.DownloadFileToDest(src, urlDest, uid, gid); err != nil {
 				return errors.Wrap(err, "downloading remote source file")
 			}
 			a.snapshotFiles = append(a.snapshotFiles, urlDest)
@@ -94,6 +99,7 @@ func (a *AddCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 	copyCmd := CopyCommand{
 		cmd: &instructions.CopyCommand{
 			SourcesAndDest: append(unresolvedSrcs, dest),
+			Chown:          a.cmd.Chown,
 		},
 		buildcontext: a.buildcontext,
 	}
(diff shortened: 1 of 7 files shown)


pkg/util/fs_util.go
```diff
@@ -534,14 +534,13 @@ func AddVolumePathToWhitelist(path string) {
 // 	1. If <src> is a remote file URL:
 // 		- destination will have permissions of 0600
 // 		- If remote file has HTTP Last-Modified header, we set the mtime of the file to that timestamp
-func DownloadFileToDest(rawurl, dest string) error {
+func DownloadFileToDest(rawurl, dest string, uid, gid int64) error {
```
greut: Does it make sense to have int64 here (and in `GetUserGroup`), when there are `uint32` everywhere else?

dani29: in `copy.go` the *ExecuteCommand* method passes int64 uid and gid params to the util.CopyDir method, which passes this around to many other places such as `fs_utils.go`...
I'm not sure what should be the actual datatype (it probably doesn't matter), but making it consistent across the code-base would require touching many functions, and I am not sure if it's a good idea. WDYT?

pkg/commands/add.go
```diff
@@ -66,7 +71,7 @@ func (a *AddCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 				return err
 			}
 			logrus.Infof("Adding remote URL %s to %s", src, urlDest)
-			if err := util.DownloadFileToDest(src, urlDest); err != nil {
+			if err := util.DownloadFileToDest(src, urlDest, uid, gid); err != nil {
```
greut: The only place conversions are needed seems to be here then. `int64(uid), int64(gid)`

dani29: I'm sorry, I'm not sure I'm following. Most methods in `fs_util.go` are expecting a int64 u/gid, with the exception of CreateFile (which is a bit weird IMO). Do you suggest to remove the casting from `util.GetUserGroup()` and perform it here?  I am not sure what'd be the advantage of that.

greut: two casts instead of four would be introduced.

dani29: I did not add the method `GetUserGroup()`, it was already there - I just moved it to the util file.
If it would return uint32 instead of int64, then in `copy.go` you'd need to perform casting twice, when passing these methods to util.CopyDir() & util.CopyFile().

pkg/util/fs_util.go
```diff
@@ -534,14 +534,13 @@ func AddVolumePathToWhitelist(path string) {
...
 	resp, err := http.Get(rawurl)
 	if err != nil {
 		return err
 	}
 	defer resp.Body.Close()
-	// TODO: set uid and gid according to current user
-	if err := CreateFile(dest, resp.Body, 0600, 0, 0); err != nil {
+	if err := CreateFile(dest, resp.Body, 0600, uint32(uid), uint32(gid)); err != nil {
```
tejal29: Thanks for addressing this TODO!

pkg/commands/add.go
greut: I didn't take the time to read this in a bigger context. Good job!

dani29:
I have started writing the unit-tests but then realized we'll have a problem to run them in a predictable way, because we need to pass a valid UID and GID.

Creating them on setup would be difficult (if impossible), but even if we tear down nicely, I'm not sure that's a flow we'd like to introduce to users running unit tests. And relying on some existing UID:GID might make these tests coupled to a specific platform.

Sticking to integration test for now - I'd imagine this problem had/would come up in the future.

PR #1135, opened 2020-03-15
Make cloudbuild.yaml re-usable for anyone

samos123:
Cloud build is also really useful for devs that have limited resources
to build Kaniko locally. This commit makes adds parameters to cloudbuild
to make it easy to adjust to your own project.

**Submitter Checklist**

review them:

**Reviewer Notes**


deploy/cloudbuild.yaml
```diff
@@ -2,18 +2,21 @@ steps:
   # First, build kaniko
   - name: "gcr.io/cloud-builders/docker"
     args: ["build", "-f", "deploy/Dockerfile",
-           "-t", "gcr.io/kaniko-project/executor:${COMMIT_SHA}", "."]
```
tejal29: We would need to change the trigger in kaniko Project to set `PROJECT_ID` to "kaniko-project" along with this change?

Can you please test the trigger which used this file after adding this substitution value 
https://pantheon.corp.google.com/cloud-build/triggers/edit/0539d73e-5ccf-46e7-af5a-6f857f99c871?project=kaniko-project

samos123: $PROJECT_ID will automatically be the current project that the cloud build is executing in. Does cloud build trigger run in the project `kaniko-project`?

tejal29: yes!. Great thanks for confirming.

PR #1140, opened 2020-03-16
Add ability to use public GCR repos without being authenticated

### LINKED ISSUE #1122: Kaniko can't pull public images from gcr.io
**Actual behavior**
Using kaniko to pull and build a kaniko image fails (kaniko is a public image) because it requires credentials.

**Expected behavior**
I expect kaniko to pull a public image w/out requiring credentials.

**To Reproduce**
1. Dockerfile is
```
FROM gcr.io/kaniko-project/executor:debug

# Over-ride the default entrypoint to avoid always calling /kaniko/executor
# so we can have additional build steps.
ENTRYPOINT [""]
```

2. Build image
```
% docker run -v `pwd`:/workspace gcr.io/kaniko-project/executor:debug --context=/workspace --no-push
INFO[0000] Resolved base name gcr.io/kaniko-project/executor:debug to gcr.io/kaniko-project/executor:debug 
INFO[0000] Resolved base name gcr.io/kaniko-project/executor:debug to gcr.io/kaniko-project/executor:debug 
INFO[0000] Retrieving image manifest gcr.io/kaniko-project/executor:debug 
ERRO[0000] Error while retrieving image from cache: gcr.io/kaniko-project/executor:debug error getting credentials - err: exit status 1, out: `docker-credential-gcr/helper: could not retrieve GCR's access token: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.` 
INFO[0000] Retrieving image manifest gcr.io/kaniko-project/executor:debug 
error building image: error getting credentials - err: exit status 1, out: `docker-credential-gcr/helper: could not retrieve GCR's access token: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.`
%
```

 | Please check if this a new feature you are proposing        | <ul>No</ul>|
 | Please check if the build works in docker but not in kaniko | <ul>Yes</ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul>Yes</ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul>No</ul>|

samos123:
Apply lazy configuration of docker credential helper for GCR at run-time only when GCR is used as the destination.

Fixes #1122 

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

Add ability to use public GCR images without being authenticated to GCR


pkg/executor/push.go
```diff
@@ -114,7 +115,15 @@ func CheckPushPermissions(opts *config.KanikoOptions) error {
 		if checked[destRef.Context().RepositoryStr()] {
 			continue
 		}
+		if strings.Contains(destRef.RegistryStr(), "gcr.io") {
```
filesnate: Comments explaining why you are eating this would be super-helpful for future devs.

samos123: sorry, not fully following, would a comment like this help?
"Checking if user provided .docker/config.json is present"
the reason I eat the error is because I only care if the file doesn't exist.

filesnate: Not the what, why are you checking for config.json?  I have a pretty good idea why because of the bug, but it will not be obvious why you are doing that for future devs.
```
// Checking for existence of docker.config as it's normally required for authenticated registries. However, if it doesn't exist, we don't want to error out since it's not required for public (non-authenticated) registries.
```

samos123: ah yeah that makes a lot more sense. I somehow thought you were fixed on the `err` variable lol. I added comments to explain the why part.

samos123:
@filesnate I added you as co-author since you made the original PR and I just improved on it. If that's fine with you, please add a comment here with following message: "@googlebot I consent."

filesnate:
@googlebot I consent.

filesnate:
> @filesnate I added you as co-author since you made the original PR and I just improved on it. If that's fine with you, please add a comment here with following message: "@googlebot I consent."
> 
> Also thank you for the thorough reviews and testing to make sure it works for your use case as well.

Thank YOU for doing the heavy lifting in the go code that I way less familiar with!

pkg/executor/push.go
```diff
@@ -115,6 +116,19 @@ func CheckPushPermissions(opts *config.KanikoOptions) error {
 			continue
 		}
 
+		// Historically kaniko was pre-configured by default with gcr credential helper,
+		// in here we keep the backwards compatibility by enabling the GCR helper only
+		// when gcr.io is in one of the destinations.
+		if strings.Contains(destRef.RegistryStr(), "gcr.io") {
```
tejal29: Can you please test to push to a public GCR repo one needs to be authenticated? My gut is mostly yes in which case this is good. If not, then we might need to add this is documentation. 

It would be great if you can add tests for this function. 
1. You can declare 2 variables 
```
// for testing
var (
execute = execCommad
checkRemotePushPermissions = remote.CheckPushPermission
stat = os.Stat
)
```

In your tests, you could create mock for these and make sure, you 
1. When "/kaniko/.docker/config.json" exists, os.Stat is not called
2. When it does not exist os.Stat is called and mock for "execute" returns nil.

pkg/executor/push.go
```diff
@@ -115,6 +116,19 @@ func CheckPushPermissions(opts *config.KanikoOptions) error {
...
+			// Checking for existence of docker.config as it's normally required for
+			// authenticated registries and prevent overwriting user provided docker conf
+			if _, err := os.Stat("/kaniko/.docker/config.json"); os.IsNotExist(err) {
```
tejal29: please define a constant for this file name in the same package.

pkg/executor/push.go
samos123: I added the testing. I noticed we were already using [afero](https://github.com/spf13/afero) in push.go and push_test.go so there wasn't a need to mock os.Stat. Thank you so much for detailed recommendations on how to test, really helpful.

There is no way to make a GCR repo pushable by the public. You can only push if you're explicitly given permission to push to a GCR repo. For reference: https://cloud.google.com/container-registry/docs/access-control or did you mean something else?

samos123:
`DOCKER_CREDENTIAL_GCR_CONFIG` should be left in the Dockerfiles since we're not setting it in `push.go`. It will only be used when `docker-credential-gcr configure-docker` is run through `push.go`. Setting the environment variable in `push.go` would be less clean imo.

pkg/executor/push.go
samos123: Done.

pkg/executor/push_test.go
```diff
@@ -307,3 +309,67 @@ func Test_makeTransport(t *testing.T) {
...
+	checkRemotePushPermission = fakeCheckPushPermission
+	for _, test := range tests {
+		testName := fmt.Sprintf("%s_ExistingDockerConf_%v", test.Destination, test.ExistingConfig)
+		t.Run(testName, func(t *testing.T) {
+			fs = afero.NewMemMapFs()
+			opts := config.KanikoOptions{
+				Destinations: []string{test.Destination},
+			}
+			if test.ExistingConfig {
+				afero.WriteFile(fs, DockerConfLocation, []byte(""), os.FileMode(0644))
```
tejal29: should this file be deleted after the test execution?

samos123: Ah good catch, I don't think it's necessary since each iteration of the for loop creates a new temporary mem based FS in lin 352. So the original file in memory is already gone by the time the function executed hence defer would be useless. I've updated the PR and added a defer anyway since it doesn't hurt.

tejal29:
I just realized this will remove file`/kaniko/.docker/config.json` in user workspace if it previously existed as a side effect.  - ofc a rare case. 
It would be ideal if for testing, 
1. we could override the variable, 
2. create a config file using `ioutil.TempFile` and then 
3. delete it in defer.

samos123:
`/kaniko/.docker/config.json` won't get deleted since afero creates membased temporary mock FS during unit test. In addition execCommand is stubbed out with a fake exec command during unit tests.

PR #1145, opened 2020-03-18
release v0.19.0

tejal29:
Release v0.19.0


tejal29:
Looks like we need to upgrade kokoro to run on golang 1.14.

samos123:
kokoro should be disabled to run all-together imo since it only works for Googlers and has already been made optional.

PR #1152, opened 2020-03-21
feat: add the Dockerfile Linter hadolint on kaniko

### LINKED ISSUE #902: Adding linting of Docker image to Kaniko container
Feature Request
One of the main issues we have is improper Dockerfiles. So as a first step I run hadolint to avoid failed Kaniko builds. However I have to use a different Docker image for this. It would be more efficient to run it in the same container or for kaniko to have a linting option.

JordanGoasdoue:
Fixes #902 

**Description**

feat: add the Dockerfile Linter `hadolint` directly on kaniko with activating flag, and multiple ignoring rules cli flag

Note: Trusted Registry from hadolint isn't added to kaniko on this PR because after i've done some test with it and it seems not working well on these cases :
- Fail for multi-part repo: https://github.com/hadolint/hadolint/issues/401
- Fail when Registry is for instance localhost:5000

I prefer not adding it yet to avoid having complains on kaniko about this.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

`Kaniko now adds the possibility to activate a Dockerfile Lint which check the Dockerfile before building the image with the possibility to ignore unwanted rules`

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ![kaniko logo](logo/Kaniko-Logo.png)
 
-kaniko is a tool to build container images from a Dockerfile, inside a container or Kubernetes cluster. 
+kaniko is a tool to build container images from a Dockerfile, inside a container or Kubernetes cluster.
 
 kaniko doesn't depend on a Docker daemon and executes each command within a Dockerfile completely in userspace.
 This enables building container images in environments that can't easily or securely run a Docker daemon, such as a standard Kubernetes cluster.
@@ -15,7 +15,7 @@ We'd love to hear from you!  Join us on [#kaniko Kubernetes Slack](https://kuber
 
 :mega: **Please fill out our [quick 5-question survey](https://forms.gle/HhZGEM33x4FUz9Qa6)** so that we can learn how satisfied you are with Kaniko, and what improvements we should make. Thank you! :dancers:
 
-Kaniko is not an officially supported Google project. 
+Kaniko is not an officially supported Google project.
 
 _If you are interested in contributing to kaniko, see [DEVELOPMENT.md](DEVELOPMENT.md) and [CONTRIBUTING.md](CONTRIBUTING.md)._
 
@@ -69,6 +69,9 @@ _If you are interested in contributing to kaniko, see [DEVELOPMENT.md](DEVELOPME
     - [--verbosity](#--verbosity)
     - [--whitelist-var-run](#--whitelist-var-run)
     - [--label](#--label)
+    - [--dockerlint](#--dockerlint)
+    - [--dockerlint-ignore](#--dockerlint-ignore)
+    - [--dockerlint-trust](#--dockerlint-trust)
   - [Debug Image](#debug-image)
 - [Security](#security)
 - [Comparison with Other Tools](#comparison-with-other-tools)
@@ -280,7 +283,7 @@ There is also a utility script [`run_in_docker.sh`](./run_in_docker.sh) that can
 ./run_in_docker.sh <path to Dockerfile> <path to build context> <destination of final image>
 ```
 
-_NOTE: `run_in_docker.sh` expects a path to a 
+_NOTE: `run_in_docker.sh` expects a path to a
 Dockerfile relative to the absolute path of the build context._
 
 An example run, specifying the Dockerfile in the container directory `/workspace`, the build
@@ -529,6 +532,26 @@ Ignore /var/run when taking image snapshot. Set it to false to preserve /var/run
 #### --label
 
 Set this flag as `--label key=value` to set some metadata to the final image. This is equivalent as using the `LABEL` within the Dockerfile.
+
+#### --dockerlint
+
+Set this flag as `--dockerlint` to activate the dockerfile linter hadolint available here: https://github.com/hadolint/hadolint.
+
+If you want to use your own hadolint config to override the kaniko's one, you can provide it as a volume: `-v $HOME/.hadolint.yaml:/kaniko/.hadolint.yaml`
+
+If you want to ignore hadolint rules directly inside the Dockerfile, you can use `inline ignores`: https://github.com/hadolint/hadolint/tree/v1.17.5#inline-ignores
+
+#### --dockerlint-ignore
+
+Set this flag as `--dockerlint-ignore <rule>` to ignore one of the rule mentionned here: https://github.com/hadolint/hadolint/tree/v1.17.5#rules
+You can set it multiple times for multiple ignoring dockerfile lint rules.
+It needs `--dockerlint` activated to be taken into account.
+
+#### --dockerlint-trust
+
+Set this flag as `--dockerlint-trust <registry>` to only build image if it's FROM a trusted registry.
+You can set it multiple times for multiple registries to trust.
+It needs `--dockerlint` activated to be taken into account.
 
 ### Debug Image
 
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"testing"
@@ -32,6 +33,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/daemon"
+	"github.com/sirupsen/logrus"
 
 	"github.com/GoogleContainerTools/kaniko/pkg/timing"
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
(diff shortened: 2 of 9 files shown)


integration/integration_test.go
```diff
@@ -454,6 +456,83 @@ func TestRelativePaths(t *testing.T) {
...
+		t.Errorf("Failed to create dir_where_to_extract: %v", err)
+	}
+
+	lintOKDockerfile := "Dockerfile_hadolint_OK"
+	lintKODockerfile := "Dockerfile_hadolint_KO"
+	lintSpecialDockerfile := "Dockerfile_hadolint_Special"
+
+	files := map[string]string{
+		lintOKDockerfile:      "FROM debian:9.11\nRUN echo \"hey\"",
+		lintKODockerfile:      "FROM debian\nRUN echo \"hey\"",
```
samos123: what does `lintKODockerfile` mean? KnockOut was what the only thing I could think of. Preferably change to `lintErrorDockerFile`  unless it's obvious to everyone.

pkg/dockerfile/dockerfile.go
```diff
@@ -87,6 +105,38 @@ func Stages(opts *config.KanikoOptions) ([]config.KanikoStage, error) {
...
+		}
+		var lintCfg LintConfig
+		lintCfg.IgnoredRules = ignoredRules
+
+		data, err := yaml.Marshal(&lintCfg)
+		if err != nil {
+			return nil, errors.Wrap(err, fmt.Sprintf("marshaling hadolint config %v", lintCfg))
+		}
+
+		cfgPath := "/kaniko/.hadolint.yaml"
```
samos123: what if the user provides .hadolint.yaml config through a volume mount e.g. `docker run -v $HOME/.hadolint.yaml:/kaniko/.hadolint.yaml kaniko-executor ...`? It seems that might be a possible scenario which should be catched by checking if the file exist before creating it and use the already existing file instead.

pkg/dockerfile/dockerfile.go
```diff
@@ -38,6 +42,7 @@ import (
 func Stages(opts *config.KanikoOptions) ([]config.KanikoStage, error) {
 	var err error
 	var d []uint8
+	lintDockerfilePath := constants.DockerfilePath
```
samos123: could you rename to `dockerFilePath`? at first when reading the code I assumed there was a separate lintDockerFile that was different from the DockerFile passed in. It's a bit confusing.

pkg/dockerfile/dockerfile.go
```diff
@@ -46,13 +51,26 @@ func Stages(opts *config.KanikoOptions) ([]config.KanikoStage, error) {
...
 		d, err = ioutil.ReadFile(opts.DockerfilePath)
 	}
 
 	if err != nil {
 		return nil, errors.Wrap(err, fmt.Sprintf("reading dockerfile at path %s", opts.DockerfilePath))
 	}
 
+	if opts.DockerfileLint {
+		if match {
+			if err := ioutil.WriteFile(lintDockerfilePath, d, 0644); err != nil {
```
samos123: alternatively we could write `d` to a tempfile and run lint on the tmpfile. That would remove the need to declare the variable `lintDockerfilePath`. The benefit of this approach would be ensuring that you can always write the contents of the Dockerfile irrelevant of what kind of path is provided e.g. https:// s3:// or local file system, since in future maybe different methods will be added, but making sure lintDockerFilePath is set correctly might become more complicated, writing to a tempfile won't be.

README.md
```diff
@@ -530,6 +530,18 @@ Ignore /var/run when taking image snapshot. Set it to false to preserve /var/run
 
 Set this flag as `--label key=value` to set some metadata to the final image. This is equivalent as using the `LABEL` within the Dockerfile.
 
+#### --dockerlint
```
samos123: Add docs about path to `/kaniko/.hadolint.yaml` which users could override by providing it as a volume.

pkg/dockerfile/dockerfile.go
JordanGoasdoue: Thank you very much for this review.
I will just explain here the reason why I didn't take into account this kind of scenario : I found that trusted registry from hadolint doesn't work perfectly with mentionned bug on description of this PR and then I've decided to exclude it from kaniko to avoid kaniko users to complain about it.
Adding for users the possibility to have their own hadolint config means they can add trusted registries inside. 
Now after your review I'm just in mode "OK let's do this anyway, if they complain it will be hadolint issue and not kaniko's one".
I will add the possibility for client to override hadolint config, but in same time it means I have to add trustedRegstry multiple args in kaniko cli.
Will work on it starting tomorrow.

samos123: Let's wait for @tejal29 thoughts as well before you start doing a major refactor since I'm a new contributor and she might have totally different view on this. I didn't originally consider your note about trustedRegistry when I reviewed the PR, that's a good point. My understanding is trustedRegistry setting won't work with GCR but will work with other registry settings, which is irrelevant of kaniko, so it would still be valuable for users to be able to get full flexibility to provide a `.hadolint.yaml`?

JordanGoasdoue: I think you are right. We shouldn't block users from using their own hadolint config, even if one of hadolint feature isn't totally working right now, because as you said, irrelevant of kaniko

pkg/dockerfile/dockerfile.go
JordanGoasdoue: I guess here, the solution is to write `d` to `constants.DockerfilePath` only if the file doesn't exist yet.
Because in the end, it's exactly what is doing the function `copyDockerfile()` with as source `opts.DockerfilePath` which is not an URL. This way we can cover all the next possible cases without even having a variable `lintDockerfilePath`

integration/integration_test.go
JordanGoasdoue: Yeah you're right, I will rename this as `lint_DL3006_Dockerfile`, it's even more detailed this way

pkg/dockerfile/dockerfile.go
samos123: +1 that would be even cleaner

samos123:
@JordanGoasdoue just discussed this with @tejal29 briefly, however we feel this might be better done as a separate step in a CI/CD pipeline. For example, run hadolint or whatever linter the user may use as a separate step in their pipeline.

The downside of including hadolint in Kaniko are:
* increased size of kaniko image due to having to include additional binaries
* additional complexity for devs since code base is increased
* additional maintenance effort needed to ensure the included hadolint stays up to date and ensure the code that integrates hadolint keeps working
* additional complexity for end-users because now there is an additional configuration parameter

Taking into consideration the downsides, could you explain more why hadolint should be included into kaniko instead of being a separate step?

JordanGoasdoue:
@samos123 Thank you for asking me my thoughts about this, it's a nice exercise indeed !
With @tejal29, you are ofc right, there are some downsides in adding this feature here.
However, I will explain why Hadolint should be added anyways:

Kaniko's main role is to build and push images from Dockerfile inside a container or a Kubernetes cluster itself since it doesn't depend on Docker daemon.

It opens a world of secure scalability and parallel docker build / push.
Thus, Dockerfile is the raw material of Kaniko, right ?

Why not pushing the best practices of Docker images directly inside your own project then?
It's a great message for users: `Here with Kaniko, we want to emphasize the quality and the trust of your own docker images, without letting you down`.

I've talked just before about scalability and parallel build, and it's exactly what I want to see in the next big major releases of Kaniko, and why not one day the support of docker-compose..
I feel that creating Hadolint stages just for this would be kind of over-killed, especially when the lifecycle would be as following:
- Hadolint[Check Multiples Dockerfiles] -> Kaniko[Build Multiples Dockerfiles -> Push Multiples Dockerfiles]
or ...
- Hadolint[Check Dockerfile] -> Kaniko[Build Dockerfile -> Push Dockerfile] for each Dockerfile

In the end, it would easily become a Spaghetti solution with too much Containers / Pods.
It's a pity to add such complex architecture when it's all about emphasizing quality, best practice and trust of images, don't you think?

Imagine instead Kaniko Container(s) or Kaniko Pod(s) which could do the following: [Check Dockerfile(s) -> Build Dockerfile(s) -> Push Dockerfile(s)].
It would be even possible to allow strong or weak coupling between these steps directly from a parameter.

A scalable solution which is able to quickly build and push multiple images from the best practices.

It's simple, it's sexy, it's... Kaniko

tejal29:
@JordanGoasdoue Thank you for the detailed explanation. All what you said makes sense. 
I am also a big believer in "Shifting left" and pushing best practices to development stage. 
Writing a good valid dockerfile is one of them. 
However, this to me sounds more like a "linter" which should be incorporated during test phase. 
e.g. when a container application developer makes a change to their code, they should
1. run unit tests using make or any other tool
2. run lint and then finally commit their changes. 
This tool can be run locally on developers machine without the need to docker just like other linters right to get quicker feedback? it would be easier for developers to iterate through the tool errors locally then grabbing kaniko container pods logs.
What do you think?

 Kaniko was designed to build docker images securely inside a container and the `hadolint` does not need to run inside a container for any security reasons. Is that accurate?

Another issue i have right now with this PR, we are adding a bunch of command line args to kaniko. One for each config/arg to the lint tool. In future, this approach is going to be hard to maintain. 
( this is ofc a secondary issue.)

I also feel, IDES  could be a great product to integrate `hadolint`. 
Please let me know if all this makes sense.

JordanGoasdoue:
@tejal29 thanks for your detailed message.
About the tool running locally before pushing the code we have same vision. 
But then I had the vision about a Hadolint in the pipeline CI/CD, and with this way of thinking, it's why I saw hadolint and kaniko as a good package to ensure before pushing any image that it's from good practices, and from trustable registry. 
Here it's not mandatory to have for kaniko as it's not directly a way of building safely inside a container. 
About the bunch of commands, we could find a way here and only accept hadolint config from volume, without accepting args from kaniko cli.

In the end I get your point and understand hadolint is good but here not mandatory for kaniko. 

Thanks for taking time explaining. 
Will close the PR now.

PR #1153, opened 2020-03-23
Checkout a specific git commit

why-xn:
**Description**

I have added support for checking out to a specific git commit. For this, the git url pattern should be like below.

git://[repository url][#reference][#commit-id]

**Reviewer Notes**

**Release Notes**

- Support for checking out a specific git commit

### DIFF
--- a/pkg/buildcontext/git.go
+++ b/pkg/buildcontext/git.go
@@ -41,6 +41,28 @@ func (g *Git) UnpackTarFromBuildContext() (string, error) {
 	if len(parts) > 1 {
 		options.ReferenceName = plumbing.ReferenceName(parts[1])
 	}
-	_, err := git.PlainClone(directory, false, &options)
+	r, err := git.PlainClone(directory, false, &options)
+
+	if err == nil && len(parts) > 2 {
+		// ... retrieving the commit being pointed by HEAD
+		_, err := r.Head()
+		if err != nil {
+			return directory, err
+		}
+
+		w, err := r.Worktree()
+		if err != nil {
+			return directory, err
+		}
+
+		// ... checking out to desired commit
+		err = w.Checkout(&git.CheckoutOptions{
+			Hash: plumbing.NewHash(parts[2]),
+		})
+		if err != nil {
+			return directory, err
+		}
+	}
+
 	return directory, err
 }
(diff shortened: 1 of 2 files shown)


why-xn:
@googlebot I signed it!

pkg/buildcontext/git.go
```diff
@@ -41,6 +41,28 @@ func (g *Git) UnpackTarFromBuildContext() (string, error) {
 	if len(parts) > 1 {
 		options.ReferenceName = plumbing.ReferenceName(parts[1])
 	}
-	_, err := git.PlainClone(directory, false, &options)
+	r, err := git.PlainClone(directory, false, &options)
+
+	if err == nil && len(parts) > 2 {
+		// ... retrieving the commit being pointed by HEAD
+		_, err := r.Head()
```
tejal29: Can you please add more explanation around why this is necessary?
It would be great to add some tests around this integration. 
Please see your existing Git Context integration test.

tomprince:
I would find something like this useful in a CI pipeline, where the kaniko task might end being run after another commit was pushed. In that case, it should be possible to run kaniko in such a way that the image built comes from the commit the rest of the pipeline is expecting, rather than whatever commit happend to get pushed later.

tejal29:
> I would find something like this useful in a CI pipeline, where the kaniko task might end being run after another commit was pushed. In that case, it should be possible to run kaniko in such a way that the image built comes from the commit the rest of the pipeline is expecting, rather than whatever commit happend to get pushed later.

Thanks for the explanation @tomprince. Not aware how other pipeline orchestrator works. I am most familiar with [tekton](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun.yaml) and it supports running a pipeline at a git commit which other tasks  in pipeline (in this case kaniko will run on)
Your use case also sounds legit. Just curios do you have your own pipeline orchestrator ?

tomprince:
> Thanks for the explanation @tomprince. Not aware how other pipeline orchestrator works. I am most familiar with [tekton](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun.yaml) and it supports running a pipeline at a git commit which other tasks in pipeline (in this case kaniko will run on)
> Your use case also sounds legit. Just curios do you have your own pipeline orchestrator ?

I will note that I probably won't use this, since we have additional constraints that means we need to wrap kaniko, but might use something like this without those contraints.

I'm not sure of @why-xn setup, but I know the setup we use does not support persistent volumes between different tasks in a pipeline, so would want to pin the commit here.

aca:
Hope this get merged.

aca:
@why-xn Seems like review is requested. Any updates on this?

tejal29:
@why-xn Can you add some unit tests?

gcalmettes:
Confirming that this would be a very nice and useful feature to have. Thanks for considering this !

PR #1164, opened 2020-03-29
Add secondary group impersonation w/ !cgo support 

### LINKED ISSUE #1097: The USER command does not set the correct gids, so extra groups are dropped
**Actual behavior**

When adding a user with an additional group and setting the user with USER, the user isn't a member of the group, according to id

**Expected behavior**

Users should be a member of all the groups listed in /etc/group

**To Reproduce**

> mkdir user-group
> cat user-group/Dockerfile
```
FROM ubuntu:latest

RUN groupadd -g 20000 bar
RUN groupadd -g 10000 foo

RUN useradd -c "Foo user" -u 10000 -g 10000 -G bar -m foo

RUN id foo

USER foo

RUN id
```

> docker run --rm -it -v `pwd`/user-group:/workspace gcr.io/kaniko-project/executor:v0.17.1  --no-push

Observe:
```
INFO[0007] RUN id foo                                   
INFO[0007] cmd: /bin/sh                                 
INFO[0007] args: [-c id foo]                            
uid=10000(foo) gid=10000(foo) groups=10000(foo),20000(bar)
INFO[0007] No files were changed, appending empty layer to config. No layer added to image. 
INFO[0007] USER foo                                     
INFO[0007] cmd: USER                                    
INFO[0007] RUN id                                       
INFO[0007] cmd: /bin/sh                                 
INFO[0007] args: [-c id]                                
uid=10000(foo) gid=10000(foo) groups=10000(foo)
``` 

As you can see the second id command shows that the foo user is not a member of bar, even though the first id command had the right answer.

 | Please check if the build works in docker but not in kaniko | <ul><li>- [V] </li></ul>| 

tstromberg:
TL;DR: PR is a bit funky so as to support !cgo builds, such as what we use in the Kaniko image. 

Without cgo, Go is not able to use getgroups() to resolve group information. This PR falls back to parsing /etc/group with the same logic as what is used in Go's source tree, but adds secondary group information.

Integration tests will require some hand-holding.

Fixes #1097

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -79,11 +80,10 @@ func (r *RunCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 
 	// If specified, run the command as a specific user
 	if userStr != "" {
-		uid, gid, err := util.GetUIDAndGIDFromString(userStr, true)
+		cmd.SysProcAttr.Credential, err = impersonate(userStr)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "impersonate")
 		}
-		cmd.SysProcAttr.Credential = &syscall.Credential{Uid: uid, Gid: gid}
 	}
 
 	env, err := addDefaultHOME(userStr, replacementEnvs)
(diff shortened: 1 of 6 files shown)


pkg/commands/run.go
```diff
@@ -93,6 +92,7 @@ func (r *RunCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.Bui
 
 	cmd.Env = env
 
+	logrus.Infof("Running: %s", cmd.Args)
```
tejal29, line 95: Can we change this to debug?

pkg/util/groupids_cgo.go
```diff
@@ -0,0 +1,13 @@
+// +build linux darwin
```
tejal29, line 1: Boilerplate license header is missing.

pkg/util/groupids_fallback.go
```diff
@@ -0,0 +1,76 @@
+// +build linux darwin
```
tejal29, line 1: Same here biolerplate license header is missing.

pkg/util/groupids_fallback.go
```diff
@@ -0,0 +1,76 @@
...
+
+		// There's no spec for /etc/passwd or /etc/group, but we try to follow
+		// the same rules as the glibc parser, which allows comments and blank
+		// space at the beginning of a line.
+		line = bytes.TrimSpace(line)
+		if len(line) == 0 || line[0] == '#' {
+			continue
+		}
+
+		// wheel:*:0:root
```
tejal29, line 67:
```suggestion
		// wheel:*:0:root,anotherGrp
```

tejal29:
Regarding integration tests, all the integration tests pass. Did you mean, you need to add new intergration test? You could add the dockerfile mentioned in the issue to `integration/dockerfiles/`

PR #1167, opened 2020-03-29
fix previous name checking in 'executor.build.fetchExtraStages'

antechrestos:
**Description**

Fixes minor bug in `fetchExtraStages` function. The `continue`instruction only breaks the inner `for` loop instead of the outer one. (see [example](https://play.golang.org/p/_dZLrbq7RMN) of two for loop and a continue in inner one)

**Submitter Checklist**

review them:

- [ ] Includes [unit tests]

**Reviewer Notes**

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -705,7 +705,7 @@ func fetchExtraStages(stages []config.KanikoStage, opts *config.KanikoOptions) e
 	t := timing.Start("Fetching Extra Stages")
 	defer timing.DefaultRun.Stop(t)
 
-	var names = []string{}
+	var names []string
 
 	for stageIndex, s := range stages {
 		for _, cmd := range s.Commands {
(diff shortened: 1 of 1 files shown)


pkg/executor/build.go
```diff
@@ -701,11 +701,16 @@ func fetchExtraStages(stages []config.KanikoStage, opts *config.KanikoOptions) e
 				continue
 			}
 			// Check if the name is the alias of a previous stage
+			previousName := false
 			for _, name := range names {
 				if name == c.From {
-					continue
```
gilbsgilbs: Definitely not a golang expert here, but pretty sure you could use a [`continue` statement with label](https://golang.org/ref/spec#Continue_statements) instead of a boolean flag.

antechrestos: @gilbsgilbs I am not a golang expert either, however I've been instructed not to ever use a label .. Out of machine language, I guess. A better solution would be to isolate this pied of code in a dedicated function I guess

antechrestos: I isolated this code in a dedicated function. Great profile picture by the way :smile:

antechrestos:
@tejal29 Guess I was right to wait as retrying the build with the current code works. This change is a minor one...Do I need to add some unit test?

tejal29:
@antechrestos this is very minor and small refactor. You don't need to add tests!

antechrestos:
thanks

PR #1196, opened 2020-04-14
Add http support for git repository context

yw-liu:
**Description**

In developing or testing environment, add http support for Git Repository context would be helpful. 

Usage: set the GIT_PULL_METHOD env var to http or https when starting the container.

### DIFF
--- a/pkg/buildcontext/git.go
+++ b/pkg/buildcontext/git.go
@@ -27,12 +27,12 @@ import (
 
 const (
 	gitPullMethodEnvKey = "GIT_PULL_METHOD"
-	gitPullMethodHttps  = "https"
-	gitPullMethodHttp   = "http"
+	gitPullMethodHTTPS  = "https"
+	gitPullMethodHTTP   = "http"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/buildcontext/git.go
```diff
@@ -44,3 +54,11 @@ func (g *Git) UnpackTarFromBuildContext() (string, error) {
 	_, err := git.PlainClone(directory, false, &options)
 	return directory, err
 }
+
+func getGitPullMethod() string {
+	gitPullMethod := os.Getenv(gitPullMethodEnvKey)
+	if ok := supportedGitPullMethods[gitPullMethod]; !ok {
+		gitPullMethod = gitPullMethodHTTPS
```
tejal29, line 61:
```suggestion
		return gitPullMethodHTTPS
```

tejal29:
Thanks @yw-liu for your PR.

pkg/buildcontext/git.go
yw-liu: Thanks @tejal29  for your suggestion, you may commit that.

PR #1199, opened 2020-04-16
Fix line endings in shell script

luba239:
Add escapes to line endings, remove extra quote


luba239:
@googlebot I signed it!

PR #1221, opened 2020-05-04
Remove direct use of DefaultTransport

antechrestos:
**Description**

Using `DefaultTransport` and manipulating its TLS configuration may lead to unexpected behaviour. For example, lets say at one moment, for a given registry that was asked to disable TLS validation we use default transport, then after that call every TLS uses of default transport will be disabled.

The only remaining use is in `Warm` method (file `warm.go`) but I think this use is only used in `warmer` image.

In this change I centralized the previous work done to clone `DefaultTransport` object on every needs and only after apply configuration to it.

To go deeper I think it might be interesting to provide in this new module a function allowing to configure once and for all  the transport creation at the beginning of the program (right after the parameter reading) and remove every references to kaniko options when they are only needed to get transport configuration. I did not do this in this change as I think it should be discussed.
 

**Submitter Checklist**

review them:

- [x] Includes unit tests

**Reviewer Notes**


tejal29:
@antechrestos Thank you for this Pr. This is definitely more secure approach. 
Can you please rebase your PR?

antechrestos:
@tejal29 done

I did not put the `NewRetry` everywhere and kept it in the push section

tejal29:
Thanks!

PR #1234, opened 2020-05-06
Apply dockefile exclude only for first stage

### LINKED ISSUE #595: The .dockerignore file also matches between stages

tejal29:
Fixes #595

**Description**

Only apply dockerignore file for the first stage. 
Add integration test to make sure this test case isnt broken in future.

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_dockerignore
+++ b/integration/dockerfiles/Dockerfile_test_dockerignore
@@ -1,5 +1,11 @@
 # This dockerfile makes sure the .dockerignore is working
 # If so then ignore/foo should copy to /foo
 # If not, then this image won't build because it will attempt to copy three files to /foo, which is a file not a directory
-FROM scratch 
+FROM scratch as base
 COPY ignore/* /foo
+
+From base as first
+COPY --from=base /foo ignore/bar
+
+FROM first
+COPY --from=first ignore/* /fooAnother/
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -596,6 +596,7 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 	}
 	logrus.Infof("Built cross stage deps: %v", crossStageDependencies)
 
+	util.IsFirstStage = true
 	for index, stage := range kanikoStages {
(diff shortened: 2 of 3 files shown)


MajorBreakfast:
FYI I think this PR implements it exactly like Docker: Docker seems to also respect the dockerignore only on the host filesystem. Inside intermediate container file systems it seems to treat the dockerignore like any other file

<details>
<summary>How I check this</summary>
  
Create file `a.txt` file and `.dockerignore` that ignores `a.txt`, then run this dockerfile
  
```dockerfile
FROM alpine:latest AS intermediate

COPY . /
RUN ls -a # shows .dockerignore, no a.txt
RUN touch a.txt
RUN ls -a # shows .dockerignore and a.txt

FROM alpine:latest

COPY --from=intermediate . /
RUN ls -a # shows .dockerignore and a.txt
```
  
</details>

PR #1251, opened 2020-05-07
Add /etc/nsswitch.conf for /etc/hosts name resolution

yw-liu:
**Description**

Add nsswitch.conf to the /etc/ folder of the executor image, this is for situations such as a user configures static host-ip mappings in /etc/hosts.

### DIFF
--- a/files/nsswitch.conf
+++ b/files/nsswitch.conf
@@ -1,0 +1,25 @@
+# /etc/nsswitch.conf
+#
+# As described on the web page https://man7.org/linux/man-pages/man3/gethostbyname.3.html,
+# without the nsswitch.conf file, the gethostbyname() and gethostbyaddr() domain queries
+# will fail to a local name server, thus the /etc/hosts will take no effect.
+#
+# For example, when hostaliases are specified for a kubernetes pod, without proper settings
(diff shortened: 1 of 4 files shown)


tstromberg:
Hi @yw-liu - Do you mind updating the PR description to describe the problem you are trying to solve, and why this is the best way to fix it?

Specifically: what is the situation you've seen where the kaniko container needs access to /etc/hosts? Are you trying to make it so that Kaniko can access a registry that is not in DNS?

I suspect you may be trying to workaround the unusual behavior where Go, and other implementations, try to use DNS first if /etc/hosts is missing, but I can't tell just yet: https://github.com/golang/go/issues/35305

Other than improving the PR description, the file looks fine and I would be OK with merging this.

files/nsswitch.conf
```diff
@@ -0,0 +1,20 @@
+# /etc/nsswitch.conf
+#
+# Example configuration of GNU Name Service Switch functionality.
```
tstromberg, line 3: Please remove this comment, and instead make the comment describe *why* an nsswitch.conf is provided.

yw-liu: Thanks for your advice, the comment is updated.

PR #1263, opened 2020-05-11
Add ability to set git auth token using environment variables

yorinasub17:
**Description**

Currently the only way to set the git auth is by including it in the context git URL. This can be problematic for certain environments such as ECS where the command args are stored in logs. This changeset introduces updates to allow setting the authentication info using the environment variables `GIT_USERNAME` and `GIT_PASSWORD`.

**Submitter Checklist**

review them:

NOTE: I didn't add an integration test for this change given that it requires real git credentials with a private git repo to fully test. If the maintainers have ideas for how to do this, happy to attempt an implementation. FWIW, I ran this manually after creating a new container and it worked perfectly:

```
%~> docker run \                                                                                                                                  
  -e GIT_PULL_METHOD=https -e GITHUB_OAUTH_TOKEN -e GIT_USERNAME \
   kaniko:git-auth-token-env \
   --context git://github.com/gruntwork-io/module-ci.git#refs/heads/master \
   --context-sub-path modules/ecs-deploy-runner/docker --no-push \
   --build-arg GITHUB_OAUTH_TOKEN
Enumerating objects: 291, done.
Counting objects: 100% (291/291), done.
Compressing objects: 100% (170/170), done.
Total 5678 (delta 178), reused 196 (delta 112), pack-reused 5387
INFO[0004] Resolved base name python:3.8-alpine to with-secrets
INFO[0004] Resolved base name python:3.8-alpine to without-secrets
```

**Reviewer Notes**

**Release Notes**

```
- For git contexts, you can now set authentication credentials for basic auth using environment variables `GIT_USERNAME` and `GIT_PASSWORD`.
```

### DIFF
--- a/pkg/buildcontext/git.go
+++ b/pkg/buildcontext/git.go
@@ -23,12 +23,17 @@ import (
 	"github.com/GoogleContainerTools/kaniko/pkg/constants"
 	git "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/transport"
+	"gopkg.in/src-d/go-git.v4/plumbing/transport/http"
 )
 
 const (
 	gitPullMethodEnvKey = "GIT_PULL_METHOD"
 	gitPullMethodHTTPS  = "https"
 	gitPullMethodHTTP   = "http"
+
+	gitAuthUsernameEnvKey = "GIT_USERNAME"
+	gitAuthPasswordEnvKey = "GIT_PASSWORD"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


yorinasub17:
@googlebot I signed it!

pkg/buildcontext/git.go
```diff
@@ -55,6 +61,18 @@ func (g *Git) UnpackTarFromBuildContext() (string, error) {
 	return directory, err
 }
 
+func getGitAuth() transport.AuthMethod {
+	username := os.Getenv(gitAuthUsernameEnvKey)
+	password := os.Getenv(gitAuthPasswordEnvKey)
+	if username != "" || password != "" {
```
tejal29, line 67: Thanks for adding this new feature. Would this work, if user only specified username but no password?
Shd it be 
1. both are present or 
2. None?

yorinasub17: Thanks for the review!

I set it as `||` because github handles the case where the username is the personal access token and the password is blank. In certain use cases, this is desirable to avoid the PAT being cached in the credentials helpers. FWIW, I do exactly that in my test example above in the PR description.

tejal29: Thanks for the explanation.

PR #1319, opened 2020-06-15
feat: upgrade go-git

caarlos0:
**Description**

https://github.com/src-d/go-git was deprecated, the new repo is https://github.com/go-git/go-git.

This PR upgrades from the older to the new.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- upgrades go-git from v4 to v5.
```


caarlos0:
I would also say maybe delete the `vendor` folder and use just `go.mod` and `go.sum`, so the diffs are nicer? 🤔

tejal29:
> I would also say maybe delete the `vendor` folder and use just `go.mod` and `go.sum`, so the diffs are nicer? 🤔

The reason we use `vendor` mode is make sure travis jobs don't download all the dependencies and have faster run time.

PR #1320, opened 2020-06-15
feat: recurse submodules

caarlos0:
**Description**

I think it's safe to assume that most people will want their submodules to be cloned as well. This adds that option to our go-git cloning.

Something I'm not sure about is how to test it. Should we create a phony git repo somewhere with submodules so we can add tests here?

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- kaniko now clone git repositories recursing submodules by default
```

### DIFF
--- a/pkg/buildcontext/git.go
+++ b/pkg/buildcontext/git.go
@@ -50,9 +50,10 @@ func (g *Git) UnpackTarFromBuildContext() (string, error) {
 	directory := constants.BuildContextDir
 	parts := strings.Split(g.context, "#")
 	options := git.CloneOptions{
-		URL:      getGitPullMethod() + "://" + parts[0],
-		Auth:     getGitAuth(),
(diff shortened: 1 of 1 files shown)


aca:
Faced the same issue.

Tried it but having authentication issue.
When kaniko context is set like this It fails to clone submodules
```git://{token}@github.mycompany.com/org/repo.git#refs/heads/master```

Do you have any clue for this?

aca:
> Faced the same issue.
> 
> Tried it but having authentication issue.
> When kaniko context is set like this It fails to clone submodules
> `git://{token}@github.mycompany.com/org/repo.git#refs/heads/master`
> 
> Do you have any clue for this?

If anyone face the same issue.
Try using GIT_USERNAME instead of adding token in context.
Clone options in kaniko should be fixed...

Still, only submodules with http protocol would work, https://

PR #1344, opened 2020-07-06
Add support for squashing.

tomprince:

**Description**

Add support for squashing created images. This depends on https://github.com/google/go-containerregistry/pull/735

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- kaniko adds a new flag `--squash` to build images with a single layer
```

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -632,6 +632,11 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 			sourceImage, err = mutate.CreatedAt(sourceImage, v1.Time{Time: time.Now()})
 			if err != nil {
 				return nil, err
+			if opts.Squash {
+				sourceImage, err = mutate.Squash(sourceImage)
+				if err != nil {
+					return nil, err
+				}
 			}
 			if opts.Reproducible {
 				sourceImage, err = mutate.Canonical(sourceImage)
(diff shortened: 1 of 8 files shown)


pkg/executor/build.go
```diff
@@ -632,6 +632,11 @@ func DoBuild(opts *config.KanikoOptions) (v1.Image, error) {
 			sourceImage, err = mutate.CreatedAt(sourceImage, v1.Time{Time: time.Now()})
 			if err != nil {
 				return nil, err
+			if opts.Squash {
```
tejal29, line 635: Looks like there is a `}` missing.

vendor/github.com/google/go-containerregistry/pkg/v1/mutate/mutate.go
```diff
@@ -342,6 +342,42 @@ func layerTime(layer v1.Layer, t time.Time) (v1.Layer, error) {
...
 }
 
+// Squash squashes the image to a single layer
+func Squash(img v1.Image) (v1.Image, error) {
+	newImage := empty.Image
+
+	tar := Extract(img)
+	layer, err := tarball.LayerFromReader(tar)
+	if err != nil {
+		return nil, fmt.Errorf("error squashing layers: %v", err)
```
greut, line 352: since Go 1.14 is used, I guess it would make sense to use the wrapping helper.

```suggestion
		return nil, fmt.Errorf("error squashing layers: %w", err)
```

tejal29:
@tomprince Have you tried the [`--single-snapshot`](https://github.com/GoogleContainerTools/kaniko#--single-snapshot) flag?

tomprince:
> Have you tried the --single-snapshot flag?

In the original context where I was going to use this[1], we were downloading images via a docker-archive export, and so gained no benefit from sharing base layers (at least in network bandwidth, which was the larger consideration). We were using `--single-snapshot`, but `--squash` should give better image size.


[1] I'm no longer working there.

tejal29:
sounds good. Will you be actively working on this PR?

PR #1359, opened 2020-07-26
Move snapshotPathPrefix into a method

jkutner:
This allows the value to be determined on the fly, which supports consumers that use Kaniko snaphot as a library and may need to change the value of `config.KanikoDir`

**Description**

The Snapshotter path prefix when invoking `TakeSnapshotFS()` is uses a local variable that can be overridden in tests. But this prevents the invoker of the `TakeSnapshotFS()` func from redefining the `config.KanikoDir`. By moving the `snapshotPathPrefix` into a `getSnapshotPathPrefix()` func, the value is resolved just-in-time, which allows the value to be configured again.

**Submitter Checklist**

review them:

**Reviewer Notes**

### DIFF
--- a/pkg/snapshot/snapshot.go
+++ b/pkg/snapshot/snapshot.go
@@ -33,7 +33,7 @@ import (
 )
 
 // For testing
-var snapshotPathPrefix = config.KanikoDir
+var snapshotPathPrefix = ""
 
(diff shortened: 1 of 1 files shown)


jkutner:
@googlebot I signed it!

PR #1453, opened 2020-10-09
feat: Added in docker cred helper for Azure Container Registry sourcing auth tokens directly from environment

chrismellard:
**Description**

The existing ACR helper does not support sourcing tokens from OAuth2 endpoints instead relying on proxying other docker helpers in to which the `az` command line places creds. To support CI scenarios a bit more easily I've created a simple docker helper for ACR that sources OAuth2 tokens from the environment, either client credentials grant or Azure MSI.


chrismellard:
@googlebot I signed it!

tejal29 (approved):

PR #1445, opened 2020-10-02
Fix permissions on cache when --no-push is set

ejose19:
Fixes #1265 

**Description**

Since pre-configuration of gcr credential helper was removed on v0.20.0, when using --no-push along with --cache, the credentials were never configured for the cache repo hence making both retrieving and pushing layers fails.

Proposing solution is: If --no-push is set, instead of not checking push permissions, we check if the --cache-repo has permissions.

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [ ] Adds integration tests if needed.

_See [the contribution guide](../CONTRIBUTING.md) for more details._

**Reviewer Notes**

- [ ] The code flow looks good. 
- [ ] Unit tests and or integration tests added.

**Release Notes**

Describe any changes here so maintainer can include it in the release notes, or delete this block.

```
Examples of user facing changes:
- fix permissions on cache when using --no-push flag

```

Issue #1265, opened 2020-05-13
GCR auth broken when --no-push is set since v0.20.0


firaxis:
**Actual behavior**
When running the docker kaniko with --no-push, cache is not available and pushing new cache layers fails.

**Expected behavior**
Docker image should be built using cache.

**To Reproduce**
```
/usr/bin/docker run --rm -v ~/.config/gcloud/:/root/.config/gcloud -v ~/workspace:/workspace 
gcr.io/kaniko-project/executor:v0.22.0 
--destination us.gcr.io/<your project>/kaniko/test 
--context dir:///workspace/ 
--dockerfile Dockerfile 
--cache=true 
--no-push 
--cache-repo us.gcr.io/<your project>/kaniko/test/cache
```

**Additional Information**
 - Dockerfile 
```
FROM debian:9
RUN apt update
RUN apt -y install vim-tiny
```

 
 **Triage Notes for the Maintainers**
The latest version it works is v0.19.0
```
/usr/bin/docker run --rm -v ~/.config/gcloud/:/root/.config/gcloud -v ~/workspace:/workspace 
gcr.io/kaniko-project/executor:v0.19.0
--destination us.gcr.io/<your project>/kaniko/test 
--context dir:///workspace/ 
--dockerfile Dockerfile 
--cache=true 
--no-push 
--cache-repo us.gcr.io/<your project>/kaniko/test/cache
```
 
 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li> NO </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li> NO </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li> NO </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li> NO </li></ul>|

sfynx:
This also breaks any FROM from a GCR repository that depends on authentication.

sameersbn:
i have the same issue. I worked around it by removing `--no-push` and pushing the image with the tag `_` so that the cache is updated

ejose19:
Neither reading the existing cache nor pushing to it works, as you mentioned last working version is 0.19.0

Msg when reading cache
```
INFO[0002] No cached layer found for ...
```

Msg when trying to push to cache
```
INFO[0222] Pushing layer .... to cache now
[...]

PR #1471, opened 2020-10-30
fix: CheckPushPermissions not being called when using --no-push and --cache-repo

ejose19:
**Description**

Issue with GCR and --no-push was solved in #1445, however this follow up [commit](https://github.com/GoogleContainerTools/kaniko/commit/a310cc6d1cd449f95cedd23393de766fdc649651#diff-dafa636d22231b15fa811fdcb798043e640b83c6a968fee91263d8305a406c93) broke it again, as the function is not being called if --no-push is set.


ejose19:
@tejal29 should #1265 be re-opened until this is merged?

tejal29 (approved):

PR #1530, opened 2020-12-21
Profiling for Snapshotting errors.

tejal29:
In this PR, 

1. Update GowalkDir dependency to pick up fixes. 
2. Added default 90 min timeout for walking FS which can be overridden by environment variable. 
3. Add slowjam to kaniko pod

Integrating slowjam with Kaniko was a quick fix. 
1. There is no change in the size of kaniko binary.   
2. The only dependency vendored in is "vendor/github.com/google/slowjam/pkg/stacklog/stacklog.go"
I don't expect the kaniko image size to increase too.   - pending verification.

TODO: 
1. Review the instruction in ReadMe.


medyagh (approved):
looks good to me

tejal29:
Generated slowjam analysis 
image

PR #1534, opened 2020-12-26
reject tarball writes with no destinations

jchorl:
**Description**

Kaniko accepts a `--tarPath` arg to write out to a tarball. I mistakenly passed this without a `--destination` and ended up with a tarball with only a `manifest.json` containing `null`.

I opened https://github.com/google/go-containerregistry/pull/887 to fix the issue in go-containerregistry (calculating a manifest without refs doesn't make much sense), but we should also check for this on the kaniko side with kaniko specific details to help users do the right thing.

**Submitter Checklist**

review them:

There are no tests for the tarball write path as far as I can tell. Adding unit/integration tests to test this one check would be a substantial lift.

**Reviewer Notes**

**Release Notes**

```
- kaniko will error if `--tarPath` is specified without `--destination` instead of building an invalid image
```

### DIFF
--- a/pkg/snapshot/snapshot.go
+++ b/pkg/snapshot/snapshot.go
@@ -169,7 +169,7 @@ func (s *Snapshotter) scanFullFilesystem() ([]string, []string, error) {
 	}
 	for _, path := range resolvedFiles {
 		if util.CheckIgnoreList(path) {
-			logrus.Tracef("Not adding %s to layer, as it's whitelisted", path)
+			logrus.Tracef("Not adding %s to layer, as it's allowlisted", path)
 			continue
 		}
 		filesToAdd = append(filesToAdd, path)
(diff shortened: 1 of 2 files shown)


tejal29 (approved):
Thanks  lot!! Only 1 nit.

pkg/snapshot/snapshot.go
```diff
@@ -169,7 +169,7 @@ func (s *Snapshotter) scanFullFilesystem() ([]string, []string, error) {
 	}
 	for _, path := range resolvedFiles {
 		if util.CheckIgnoreList(path) {
-			logrus.Tracef("Not adding %s to layer, as it's whitelisted", path)
+			logrus.Tracef("Not adding %s to layer, as it's allowlisted", path)
```
tejal29, line 172:
```suggestion
			logrus.Tracef("Not adding %s to layer, as it's ignored", path)
```

PR #1555, opened 2021-01-25
Optimize access to image manifests to reduce propability on DockerHub 429s

SaschaSchwarze0:
**Description**

Kaniko is quite chatty when it comes to retrieving image manifests from container registries. While these calls are all small and usually quick, there are still good reasons to keep the number at a minimum. The most recent prominent reason is the [pull request limit introduced by DockerHub](https://docs.docker.com/docker-hub/download-rate-limit), but we are also seeing from time to time what DockerHub in its [documentation](https://docs.docker.com/docker-hub/download-rate-limit/#other-limits) calls the general rate limit.

I analyzed how often Kaniko retrieves an image for [a simple Dockerfile](https://github.com/SaschaSchwarze0/npm-simple/blob/master/normal/Dockerfile) which at the end happens in remote.go's `RetrieveRemoteImage` and it happened four times:

1. Within build.go's `CalculateDependencies` it calls `RetrieveSourceImage`, this goes for me into the `cachedImage` because the `cache-dir` argument has defined `"/cache"` as default value (and `opts.Cache` is not yet checked there) and retrieves the image to know the digest for the local cache access
2. As there is nothing in my local cache, `image_util`'s `RetrieveSourceImage` then calls `remote`'s `RetrieveRemoteImage` directly
3. The build.go's `newStageBuilder` function also calls `image_util`'s `RetrieveSourceImage` which - as in (1) - ends up in the `cachedImage` code path
4. It then goes the same path as in (2)

Two changes I am making:

1. I am changing the condition in image_util.go to check `opts.Cache` in addition to `opts.CacheDir`. This eliminates (1) and (3).
2. I am introducing a very simple map-based cache in remote.go. That's imo good enough for our use case given. This eliminates (4). We do not need to take care of cache TTL or cleanup because the Kaniko process has a limited lifetime anyway.

If there are multi-stage builds with stages using the same `FROM` image, then even more manifest requests are saved.

**Submitter Checklist**

review them:

**Reviewer Notes**

### DIFF
--- a/pkg/image/image_util.go
+++ b/pkg/image/image_util.go
@@ -69,7 +69,7 @@ func RetrieveSourceImage(stage config.KanikoStage, opts *config.KanikoOptions) (
 
 	// Finally, check if local caching is enabled
 	// If so, look in the local cache before trying the remote registry
-	if opts.CacheDir != "" {
+	if opts.Cache && opts.CacheDir != "" {
 		cachedImage, err := cachedImage(opts, currentBaseName)
 		if err != nil {
 			switch {
--- a/pkg/image/remote/remote.go
+++ b/pkg/image/remote/remote.go
@@ -31,9 +31,20 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var (
+	manifestCache = make(map[string]v1.Image)
+)
+
 // RetrieveRemoteImage retrieves the manifest for the specified image from the specified registry
 func RetrieveRemoteImage(image string, opts config.RegistryOptions, customPlatform string) (v1.Image, error) {
 	logrus.Infof("Retrieving image manifest %s", image)
+
+	cachedRemoteImage := manifestCache[image]
+	if cachedRemoteImage != nil {
+		logrus.Infof("Returning cached image manifest")
+		return cachedRemoteImage, nil
+	}
+
 	ref, err := name.ParseReference(image, name.WeakValidation)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


SaschaSchwarze0:
The Travis job failed when starting minikube:

```bash
+sudo minikube start --vm-driver=none
* minikube v1.17.0 on Ubuntu 18.04 (kvm/amd64)
* Using the none driver based on user configuration

X Exiting due to RSRC_INSUFFICIENT_CORES: Requested cpu count 1 is less than the minimum allowed of 2

Makefile:62: recipe for target 'minikube-setup' failed
make: *** [minikube-setup] Error 29
The command "make minikube-setup" failed and exited with 2 during .

Your build has been stopped.
```

I do not see any relationship related to my changes.

tejal29:
> The Travis job failed when starting minikube:
> 
> ```shell
> +sudo minikube start --vm-driver=none
> * minikube v1.17.0 on Ubuntu 18.04 (kvm/amd64)
> * Using the none driver based on user configuration
> 
> X Exiting due to RSRC_INSUFFICIENT_CORES: Requested cpu count 1 is less than the minimum allowed of 2
> 
> Makefile:62: recipe for target 'minikube-setup' failed
> make: *** [minikube-setup] Error 29
> The command "make minikube-setup" failed and exited with 2 during .
> 
> Your build has been stopped.
> ```
> 
> I do not see any relationship related to my changes.

Thanks is correct. Thanks to @medyagh & @afbjorklund  the issue is fixed in minikube.
https://github.com/kubernetes/minikube/issues/10260

tejal29 (approved):

PR #801, opened 2019-10-03
Support Dockerfile.dockerignore

victornoel:
This fixes #776

**Description**

- Use `Dockerfile.dockerignore` in priority over the context's `.dockerignore` file.
- Actual name checked is based on the docker file passed to the command line
- Introduces unit tests and integration test to validate the behaviour.

~~I haven't added an unit test because it wasn't clear what was to correct way to test this because excludes seems to be managed globally via a static field. I would be interested by guidance from knowledgeable people on this.~~

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [x] Adds integration tests if needed.

_See [the contribution guide](../CONTRIBUTING.md) for more details._

**Reviewer Notes**

- [x] The code flow looks good. 
- [x] Unit tests and or integration tests added.

**Release Notes**

- Support `Dockerfile.dockerignore` file in the same way as `docker build`

PR #1607, opened 2021-03-23
Fix resolution of Dockerfile relative dockerignore

### LINKED ISSUE #1451: dockerfile.dockerignore not working
**Actual behavior**
https://github.com/GoogleContainerTools/kaniko/pull/801 presumably added the ability to use kaniko with a dockerignore file following the format `${DOCKERFILE}.dockerignore` where `$DOCKERFILE` is the path specified in the `--dockerfile` arg to the executor. However, in my tests it's not being picked up and used.

**Expected behavior**
Match what `docker build` does when `DOCKER_BUILDKIT=1` and what #801 claims to do.

I expect if I run the executor with `--dockerfile foo` then `foo.dockerignore` will be used as the ignore file applied to the build context if it is found.

**To Reproduce**

1. Setup a simple environment with a file that we want to fail to copy into the container.

```shell
$ mkdir -p foo
$ touch foo/bar
$ echo "FROM ubuntu\nCOPY foo foo\nRUN ls foo" > Dockerfile
$ echo '*' > Dockerfile.dockerignore
```

2. Build an image.

```shell
$ docker run --rm -it -v $(pwd):/build gcr.io/kaniko-project/executor --context /build --dockerfile /build/Dockerfile --no-push
INFO[0000] Retrieving image manifest ubuntu
INFO[0000] Retrieving image ubuntu
INFO[0000] Retrieving image manifest ubuntu
INFO[0000] Retrieving image ubuntu
INFO[0002] Built cross stage deps: map[]
INFO[0002] Retrieving image manifest ubuntu
INFO[0002] Retrieving image ubuntu
INFO[0002] Retrieving image manifest ubuntu
INFO[0002] Retrieving image ubuntu
INFO[0003] Executing 0 build triggers
INFO[0003] Unpacking rootfs as cmd COPY foo foo requires it.
INFO[0008] COPY foo foo
INFO[0008] Taking snapshot of files...
INFO[0008] RUN ls foo
INFO[0008] Taking snapshot of full filesystem...
INFO[0008] cmd: /bin/sh
INFO[0008] args: [-c ls foo]
INFO[0008] Running: [/bin/sh -c ls foo]
bar
INFO[0008] Taking snapshot of full filesystem...
INFO[0008] Skipping push to container registry due to --no-push flag
```

3. Note that it worked and copied the file instead of ignoring it. **unexpected**
4. Move the file to just `.dockerignore` at the root of the build context and run it again.

```shell
$ mv Dockerfile.dockerignore .dockerignore$ docker run --rm -it -v $(pwd):/build gcr.io/kaniko-project/executor --context /build --dockerfile /build/Dockerfile --no-push
INFO[0000] Using dockerignore file: /build/.dockerignore
INFO[0000] Retrieving image manifest ubuntu
INFO[0000] Retrieving image ubuntu
INFO[0001] Retrieving image manifest ubuntu
INFO[0001] Retrieving image ubuntu
INFO[0002] Built cross stage deps: map[]
INFO[0002] Retrieving image manifest ubuntu
INFO[0002] Retrieving image ubuntu
INFO[0003] Retrieving image manifest ubuntu
INFO[0003] Retrieving image ubuntu
INFO[0004] Executing 0 build triggers
INFO[0004] Unpacking rootfs as cmd COPY foo foo requires it.
INFO[0008] COPY foo foo
INFO[0008] Taking snapshot of full filesystem...
INFO[0010] RUN ls foo
INFO[0010] Taking snapshot of full filesystem...
INFO[0010] cmd: /bin/sh
INFO[0010] args: [-c ls foo]
INFO[0010] Running: [/bin/sh -c ls foo]
ls: cannot access 'foo': No such file or directory
error building image: error building stage: failed to execute command: waiting for process to exit: exit status 2
```

5. Note that it failed to copy the file, as expected, and that an INFO log message noted at the top that it was using the dockerignore file.
 
 | Please check if the build works in docker but not in kaniko | <ul><li>- [X] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [X] </li></ul>|

cehoffman:
Fixes #1451

**Description**

Docker ships with buildkit enabled by default now enabling Dockerfile
specific dockerignore resolution, e.g. Dockerfile.dockerignore.

This functionality was added in #801, but not enabled in integration tests
because the standard docker build did not support this style of ignore.

I'm not sure where to add the integration test because the existing dockerfile
does an adequate job for validating the fix. Also I'm completely open to
changing the name of the new opts value from `OriginalDockerfilePath`.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- kaniko honors .dockerignore files specific to a Dockerfile, e.g.
  Dockerfile.dockerignore
```

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -254,6 +254,12 @@ func copyDockerfile() error {
 	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, util.FileContext{}, util.DoNotChangeUID, util.DoNotChangeGID); err != nil {
 		return errors.Wrap(err, "copying dockerfile")
 	}
+	dockerignorePath := opts.DockerfilePath + ".dockerignore"
+	if util.FilepathExists(dockerignorePath) {
+		if _, err := util.CopyFile(dockerignorePath, constants.DockerfilePath+".dockerignore", util.FileContext{}, util.DoNotChangeUID, util.DoNotChangeGID); err != nil {
+			return errors.Wrap(err, "copying Dockerfile.dockerignore")
+		}
+	}
 	opts.DockerfilePath = constants.DockerfilePath
 	return nil
 }


cmd/executor/cmd/root.go
```diff
@@ -223,6 +223,7 @@ func resolveDockerfilePath() error {
 			return errors.Wrap(err, "getting absolute path for dockerfile")
 		}
 		opts.DockerfilePath = abs
+		opts.OriginalDockerfilePath = abs
```
tejal29, line 226: Is there a difference between `OriginalDockerfilePath` and `DockerfilePath`

cehoffman: Yes because the line immediately following in copyDockerFile

https://github.com/GoogleContainerTools/kaniko/blob/0477900febfacb98b88abdbfaf8bbaeb05f51060/cmd/executor/cmd/root.go#L257

tejal29: Sorry @cehoffman for the delayed response. The L257 code makes sure, we fall back to `Dockerfile` if the file passed in command line arg `--dockerfile` does not exist. 

If we fallback to `Dockerfile` then the dockerfile ignore should also be `Dockerfile.dockerignore` right?

cehoffman: L257 I believe is to make the Dockerfile used for build always ends up at `/kaniko/Dockerfile`. I see now I didn’t modify the the context local Dockerfile lookup on L235

I think perhaps the right solution is to extend `copyDockerfile()` to detect a relative dockerignore and copy it to `/kaniko/Dockerfile.dockerignore`. That would then let the existing lookup resolve and not require the `OriginalDockerfilePath` struct item.

PR #1622, opened 2021-04-07
adds ignore-path command arguments to executor

### LINKED ISSUE #1621: Allow configuration of extra ignore-list paths
### Suggestion

This refers to the [ignore-list](https://github.com/GoogleContainerTools/kaniko/blob/86e361c20ef7063c1fd80111269d68da6e859c56/pkg/util/fs_util.go#L380-L382) that is used when cleaning up the filesystem between different build steps. Currently the ignore list is [initialized with a few default values](https://github.com/GoogleContainerTools/kaniko/blob/86e361c20ef7063c1fd80111269d68da6e859c56/pkg/util/fs_util.go#L58-L75), and then [all filesystem mounts are added to the list](https://github.com/GoogleContainerTools/kaniko/blob/86e361c20ef7063c1fd80111269d68da6e859c56/pkg/util/fs_util.go#L410-L416).

It would be great if we could pass a CLI flag to add paths to the list, for example `/kaniko/executor --ignore-path /first/path --ignore-path /second/path`. 

### Use-case

In our case, we extend the Kaniko image with a few functionalities that our builds require, so we place all the shell scripts and tools in one directory (say `/.tools`). These scripts manage the build lifecycle and run the `kaniko` executor binary. We base our `Dockerfile` on the official Kaniko executor image and, among other things, add a `VOLUME /.tools` directive so that it's picked up by Kaniko and added to the ignore list.

However we run our builds on Kubernetes, and the problem with this approach is that we can't enforce disk storage limits on `/.tools` because it's not defined as a volume in the Kubernetes container spec. We also can't do that because it would then mount `/.tools` as an empty directory not containing the files that are in the actual image.

### ---
I realize that running the kaniko executor on a different image is neither officially supported nor recommended, but this option would make things a lot easier for us. If this idea is approved we'd be happy to give it a shot and submit a PR :)

jonfriesen:
Fixes #1621

**Description**

Allows the command line setting of custom ignore paths via the executor. This follows the style of `/kaniko/executor --ignore-path /first/path --ignore-path /second/path`.

This change renames `initialIgnoreList` with `defaultIgnoreList` and creates a `baseIgnoreList` which is assigned to `ignoreList`. This allows ignore paths to be added via the command line `--ignore-path` argument that persists across `DetectFilesystemIgnoreList(...)` calls while still allowing the `ignoreList` to be reset from changes made in flight. 

**Reviewer Notes**

**Release Notes**

- adds `--ignore-path` executor argument to allow custom ignore-path's to be set. This command can be used repeatedly for multiple paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ _If you are interested in contributing to kaniko, see [DEVELOPMENT.md](DEVELOPME
     - [--use-new-run](#--use-new-run)
     - [--verbosity](#--verbosity)
     - [--whitelist-var-run](#--whitelist-var-run)
+    - [--ignore-path](#--ignore-path)
   - [Debug Image](#debug-image)
 - [Security](#security)
   - [Verifying Signed Kaniko Images](#verifying-signed-kaniko-images)
@@ -747,6 +748,10 @@ Set this flag as `--verbosity=<panic|fatal|error|warn|info|debug|trace>` to set 
 #### --whitelist-var-run
 
 Ignore /var/run when taking image snapshot. Set it to false to preserve /var/run/* in destination image. (Default true).
+
+#### --ignore-path
+
+Set this flag as `--ignore-path=<path>` to ignore path when taking an image snapshot. Set it multiple times for multiple ignore paths.
 
 ### Debug Image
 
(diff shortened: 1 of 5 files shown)


SaschaSchwarze0:
Is it worth listing the new argument in the project's README?

jonfriesen:
@SaschaSchwarze0 good catch! I'll add it to the README shortly. Thanks :)

jonfriesen:
@SaschaSchwarze0  updated :pray:

tejal29 (approved):

RufusJWB:
It would be great, if you could sort the command argument description in the README file at the alphabetically correct place. I was searching quite some time for it...

PR #1676, opened 2021-06-18
Add support for CPU variants

### LINKED ISSUE #1591: No support for CPU variant (like v7 for ARM) on customPlatform builds
**Actual behavior**

The moment that `--customPlatform=linux/arm` gets specified, the variant `v5` gets selected for those multi-arch images supporting `linux/arm/v5` and `linux/arm/v7`. **There's no possibility to specify architecture `arm`, variant `v7` on `customPlatform` builds** (I've also tried the specification that works for Docker `linux/arm/v7`). As a consequence, for example for Debian images only the older `armel` architecture can be selected, not the newer `armhf`.

**Expected behavior**
An option to specify the architecture variant (e.g. as mentioned `linux/arm/v7`) when more than one supported (like is the case for ARM32).

**To Reproduce**
1. Get a container host with `binfmt` support for arm/v5 and arm/v7.
2. Execute a Kaniko build using the below provided Dockerfile with the option `--customPlatform=linux/arm`.
3. Observe output `armel` (instead of the expected `armhf`).

**Additional Information**
 - Dockerfile
```
FROM debian:stable

RUN dpkg --print-architecture
```
 - Build Context
   No files needed
 
 | Please check if the build works in docker but not in kaniko | works in docker with the option `--platform linux/arm/v7` | 
 | Please check if this error is seen when you use `--cache` flag | Yes, the error is still being seen |
 | Please check if your dockerfile is a multistage dockerfile | No multistage |

Silvanoc:
Signed-off-by: Silvano Cirujano Cuesta <silvano.cirujano-cuesta@siemens.com>
Inspired-by: mickkael 19755421+mickkael@users.noreply.github.com

Fixes #1591

**Description**

Adds the option to build an image with a CPU variant for the customplatform option. Look for `variant` here for the meaning of CPU variant: https://github.com/opencontainers/image-spec/blob/master/image-index.md#image-index-property-descriptions.

The CPU variant specification cannot be included in the resulting image due to a limitation of the OCI-image specification. The effect of this PR is that in the case of multi-arch base images the right architecture and CPU variant gets used.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- kaniko supports cpu variants in custom platform, like `--customPlatform=linux/arm/v5`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,7 +588,12 @@ and you want to build one of its subfolders instead of the root folder.
 #### --customPlatform
 
 Allows to build with another default platform than the host, similarly to docker build --platform xxx
-the value has to be on the form `--customPlatform=linux/arm` , with acceptable values listed here: [GOOS/GOARCH](https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63)
+the value has to be on the form `--customPlatform=linux/arm`, with acceptable values listed here: [GOOS/GOARCH](https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63).
+
+It's also possible specifying CPU variants adding it as a third parameter (like  `--customPlatform=linux/arm/v5`).
+Currently CPU variants are only known to be used for the ARM architecture as listed here: [GOARM](https://github.com/golang/go/wiki/GoArm#supported-architectures)
+
+_The resulting images cannot provide any metadata about CPU variant due to a limitation of the OCI-image specification._
 
 _This is not virtualization and cannot help to build an architecture not natively supported by the build host. This is used to build i386 on an amd64 Host for example, or arm32 on an arm64 host._
 
--- a/pkg/image/remote/remote.go
+++ b/pkg/image/remote/remote.go
@@ -138,10 +138,16 @@ func remoteOptions(registryName string, opts config.RegistryOptions, customPlatf
 // CurrentPlatform returns the v1.Platform on which the code runs
 func currentPlatform(customPlatform string) v1.Platform {
 	if customPlatform != "" {
-		return v1.Platform{
-			OS:           strings.Split(customPlatform, "/")[0],
-			Architecture: strings.Split(customPlatform, "/")[1],
+		customPlatformArray := strings.Split(customPlatform, "/")
+		imagePlatform := v1.Platform{}
+		imagePlatform.OS = customPlatformArray[0]
+		if len(customPlatformArray) > 1 {
+			imagePlatform.Architecture = customPlatformArray[1]
+			if len(customPlatformArray) > 2 {
+				imagePlatform.Variant = customPlatformArray[2]
+			}
 		}
+		return imagePlatform
 	}
 	return v1.Platform{
 		OS:           runtime.GOOS,


mickkael:
I had tried the same, but with the build included, which fails because of the OCI limitation.
https://github.com/mickkael/kaniko/commit/ac6fc1f387136be4580e51ac4176616ba00f87ed

remote only is a good compromise for now. Else qemu is likely a solution for armv6, but armv5's support is limited. (https://hub.docker.com/r/tonistiigi/binfmt)

Silvanoc:
> remote only is a good compromise for now. Else qemu is likely a solution for armv6, but armv5's support is limited. (https://hub.docker.com/r/tonistiigi/binfmt)

As mentioned in [this issue comment](https://github.com/GoogleContainerTools/kaniko/issues/1591#issuecomment-864017827), I mostly focus on the base image (sub)issue, because it's the one we're stumbling upon now.

In fact I'd propose to split #1591 in two:

1. #1591 keeps focused on ensuring that the right base image gets used
2. A new issue focused on the missing support for CPU variant in the OCI-image manifest, this one mostly keeping track of https://github.com/opencontainers/image-spec/pull/777. This would be a long-running issue that will probably take long until being supported at all levels, some projects decide not to track them as issues, I don't know how Kaniko is tracking them.

I consider that either a subset of your patch or mine should fix #1591 (after the split).

- If you prefer to have contributions from different companies, you just have to wait for me to get the signing of the CLA clarified.
- Otherwise taking only the part of your patch affecting `remote` would also do it.

I don't see in the link provided by you any information about qemu-user having issues with ARMv5. An issue with Qemu v5 is being reported, but it doesn't have anything to do with ARMv5. Although don't see ARMv5 reported as supported in that repo, but "binfmt_misc" doesn't require it for working (in fact using a container to configure the host can trigger ugly race conditions that have costed us hours, I would discourage from doing so on systems that need to use `binfmt_misc` concurrently). And as long as I don't see a Qemu reporting about it, I consider it a limitation of that concrete set-up.

mickkael:
To install qemu at built time, I use the image tonistiigi/binfmt , this is also the default in the github action set-qemu, so it's widely used.
This image supports only arm/v7 and arm/v6

Silvanoc:
> To install qemu at built time, I use the image tonistiigi/binfmt , this is also the default in the github action set-qemu, so it's widely used.
> This image supports only arm/v7 and arm/v6

Kaniko shouldn't be anyway concerned about how the user gets the needed host support to build the image. It should only support getting the right base image.

The Docker Hub official Debian image is available for ARMv5 and ARMv7 and they rely on completely different packages (being "armel" and "armhf" the respective names). If Kaniko isn't capable of getting the right one, then no matter how much I do on my host, it won't work.

QEMU is capable of handling both ARMv5 and ARMv7 (I haven't tested if the resulting artifacts also run properly on real hardware though), therefore I still see the limitation on Kaniko not pulling the right base image. And that's what I'd like first to overcome.

Silvanoc:
@googlebot I signed it!

Silvanoc:
@mickkael the build pipeline fails due to some credential issues that I don't know how to fix, could you help me on that?

mickkael:
all the other PR are failing the same test... it isn't your code causing the issue.

Silvanoc:
> all the other PR are failing the same test... it isn't your code causing the issue.

So what does it mean for this PR? It will be reviewed somewhere in time? I shouldn't expect any merging until those CI issues are resolved?

tejal29:
Sorry folks, due to some internal project organization, tests accessing a particular image are failing. 
Will fix it.

tejal29 (approved):

Silvanoc:
@tejal29 do you have any plans to do a release in the upcoming future? Just to estimate when this patch will flow into a release. Thanks

Silvanoc:
> @tejal29 do you have any plans to do a release in the upcoming future? Just to estimate when this patch will flow into a release. Thanks

I'd love to see a new release as demanded in https://github.com/GoogleContainerTools/kaniko/issues/1740 including this patch.

PR #1681, opened 2021-06-23
Handle multiple www-authenticate headers

seba-ban:
Function `pickFromMultipleChallenges` was added to be able to pick from multiple www-authenticate headers the one kaniko can handle.

**Description**

Sometimes registry can send back multiple www-authenticate headers, e.g. `Negotiate` and `Basic`. Kaniko is handling only `basic` and `bearer` types ([here](https://github.com/GoogleContainerTools/kaniko/blob/094fe52b3746b6cd59c4c3669cf2f640647321cd/vendor/github.com/google/go-containerregistry/pkg/v1/remote/transport/transport.go#L73)). Current behavior in case of multiple www-authenticate headers is to pick the first one. However, in cases as described above, `Negotiate` header is picked which results in an error; `Basic` header had been picked, the whole build and push process would run just fine. This PR is supposed to handle such cases.

**Submitter Checklist**

review them:

**Reviewer Notes**


tejal29:
@seba-ban can you please contribute this code to upstream package https://github.com/google/go-containerregistry

tejal29:
Closing this Pr here. Thank you!!

PR #1686, opened 2021-07-03
refactor: adjust bpfd container runtime detection

### LINKED ISSUE #1592: Container detection fails on cgroup v2 devices
**Actual behavior**
Kaniko fails to detect that it's running in a Docker container when not run on a system using cgroup v2.

Error message:
```
kaniko should only be run inside of a container, run with the --force flag if you are sure you want to continue
```

This issue can be reproduced on a Debian Testing/Bullseye (default using cgroup v2) installation and on an ArchLinux installation using cgroup v2.

This issue originates from the way [genuinetools/bpfd/proc](https://github.com/genuinetools/bpfd/blob/master/proc/proc.go) detects if running in a Docker container or not. On Debian Stable/Buster (default using cgroup v1) and Archlinux using cgroup v1

**Expected behavior**
Kaniko detects the container runtime properly and doesn't report any issue.

**To Reproduce**
A system with cgroup v2 is needed to reproduce the issue.

Steps to reproduce the behavior (lines starting with `$ ` are to be run on the host, lines starting with `# ` are to be run in the container):
1. `$ docker run --rm -ti --entrypoint "" gcr.io/kaniko-project/executor:v1.5.1-debug sh`
2. `# echo "FROM alpine:latest" > Dockerfile`
3. `# /kaniko/executor --dockerfile Dockerfile --destination /dev/null`

**Additional Information**
 - Dockerfile 
   Independent of Dockerfile, tested with different ones.
 - Build Context
   Independent of context (tested on contexts without copying any file)
 
 | Please check if the build works in docker but not in kaniko | Cannot be tested in Docker | 
 | Please check if this error is seen when you use `--cache` flag | No difference |
 | Please check if your dockerfile is a multistage dockerfile | Independent from Dockerfile |

ejose19:
Fixes #1592

**Description**

Applies https://github.com/genuinetools/bpfd/pull/19, which improves detection of container runtime based on mechanism used by [systemd](https://github.com/systemd/systemd/blob/6604fb0207ee10e8dc05d67f6fe45de0b193b5c4/src/basic/virt.c#L551-L694).

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- improve container detection
```

### DIFF
--- a/vendor/github.com/genuinetools/bpfd/proc/proc.go
+++ b/vendor/github.com/genuinetools/bpfd/proc/proc.go
@@ -150,6 +150,12 @@ func GetContainerRuntime(tgid, pid int) ContainerRuntime {
 		return runtime
 	}
 
+	// Check for container specific files
+	runtime = detectContainerFiles()
+	if runtime != RuntimeNotFound {
+		return runtime
+	}
+
 	return RuntimeNotFound
 }
 
@@ -161,6 +167,28 @@ func getContainerRuntime(input string) ContainerRuntime {
 	for _, runtime := range ContainerRuntimes {
 		if strings.Contains(input, string(runtime)) {
 			return runtime
+		}
+	}
+
+	return RuntimeNotFound
+}
+
+// Related implementation: https://github.com/systemd/systemd/blob/6604fb0207ee10e8dc05d67f6fe45de0b193b5c4/src/basic/virt.c#L523-L549
+func detectContainerFiles() ContainerRuntime {
+	files := []struct {
+		runtime  ContainerRuntime
+		location string
+	}{
+		// https://github.com/containers/podman/issues/6192
+		// https://github.com/containers/podman/issues/3586#issuecomment-661918679
+		{RuntimePodman, "/run/.containerenv"},
+		// https://github.com/moby/moby/issues/18355
+		{RuntimeDocker, "/.dockerenv"},
+	}
+
+	for i := range files {
+		if fileExists(files[i].location) {
+			return files[i].runtime
 		}
 	}
 


tejal29 (approved):

Silvanoc:
I hope you are aware of the fact that this approach won't work if running from a container image not generated with Docker...

The file being used to detect the container runtime in reality only identifies the tool used to generate the container image. If I use a Docker engine to run a container image generated with buildah, that file will be missing. Still my workload is running in a Docker container.

ejose19:
@Silvanoc Yes, and I agree with all your points on https://github.com/GoogleContainerTools/kaniko/issues/1592#issuecomment-861254141, would be better if docker applied the podman approach so detection wouldn't be so flaky. However, as long the changes doesn't require that much effort, we can apply the upstream workarounds so users doesn't encounter these errors.

> What's even worse: if I use a container image generated with Docker, extract the rootfs and use it with chroot, Kaniko will believe to be running from a Docker container!

That doesn't seem like an usual case at all, and if you intend to run kaniko here, then it wouldn't be much different than having to use `--force`

MyIgel:
After the latest version bump this change got reverted as the upstream mr did not got merged -> would it be possible to reapply it to release a new kaniko version @tejal29?

imjasonh:
> After the latest version bump this change got reverted as the upstream mr did not got merged -> would it be possible to reapply it to release a new kaniko version @tejal29?

Oh nice. Could you send a PR to add a `replace` in the go.mod so we get this change?

skorpy2009:
Hey @imjasonh @priyawadhwa @dlorenc 

I have the same problem and looked at the code. The use of this package seams to be only for this one very simple use case. Most of the package functionality seems not being used and we get an unmaintained (12/2018 last release(v0.0.1)) 3rd party hell which could be a few lines to detect the residency in a container and implementing it in kaniko should be less trouble in long term.
Any opinion before a PR is being opened?

Best
skorpy

imjasonh:
+1 to using what we need, and dropping a seemingly-unmaintained dependency. Thanks for looking into it!

MyIgel:
@imjasonh I added the code in https://github.com/GoogleContainerTools/kaniko/pull/1996 as a PR which seems to be the better maintainable option, would that be a possible solution?

PR #1727, opened 2021-08-22
Feature: debug shell

kvaps:
**Description**

This PR adds two new options: `--debug` and `--debug-shell` which allow to run interactive shell in case of RUN command failed.
This option might be very useful during the development the Dockerfile. I'm going to implement it's support into [`kubectl-build`](https://github.com/kvaps/kubectl-build) plugin.

Live demo:
[asciicast](https://asciinema.org/a/432020)

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [ ] Adds integration tests if needed.

_See [the contribution guide](../CONTRIBUTING.md) for more details._

**Reviewer Notes**

- [ ] The code flow looks good. 
- [ ] Unit tests and or integration tests added.

**Release Notes**

Describe any changes here so maintainer can include it in the release notes, or delete this block.

```
Examples of user facing changes:
- kaniko adds a new flag `--debug` and `--debug-cmd` to spawn interactive debug shell in case of RUN command failed
```

PR #1728, opened 2021-08-22
tar: read directly from stdin

kvaps:
**Description**

When using `--context=tar://stdin` option kaniko first writing tar.gz file on the drive, then unpacking it.
This PR changes this behavior to allow unpacking tar archive by reading it directly from stdin.

This also enables opportunity to reuse stdin to run debug shell #1727

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- reading context from 'tar://stdin' improved by unpacking directly from the piped stream
```

### DIFF
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -261,12 +261,8 @@ func childDirInIgnoreList(path string) bool {
 	return false
 }
 
+// UnTar returns a list of files that have been extracted from the tar archive at r to the path at dest
 func UnTar(r io.Reader, dest string) ([]string, error) {
-	return unTar(r, dest)
-}
-
-// unTar returns a list of files that have been extracted from the tar archive at r to the path at dest
-func unTar(r io.Reader, dest string) ([]string, error) {
 	var extractedFiles []string
 	tr := tar.NewReader(r)
 	for {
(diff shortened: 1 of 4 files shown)


kvaps:
PR rebased to `v1.7.0`, new docker images with this change prepared:

```
ghcr.io/kvaps/kaniko-executor:v1.7.0
ghcr.io/kvaps/kaniko-executor:v1.7.0-debug
ghcr.io/kvaps/kaniko-warmer:v1.7.0
```

source code of fork in `master` branch:
https://github.com/kvaps/kaniko/

pkg/util/fs_util.go
```diff
@@ -261,6 +261,10 @@ func childDirInIgnoreList(path string) bool {
 	return false
 }
 
+func UnTar(r io.Reader, dest string) ([]string, error) {
+	return unTar(r, dest)
```
imjasonh, line 265: Should we just export `UnTar` instead of exporting this wrapper with the same signature?

imjasonh (commented):
This change seems like something that should be accompanied by an integration test, wdyt?

pkg/util/fs_util.go
kvaps: Yeah, let's do that!

imjasonh (approved):

PR #1785, opened 2021-10-19
chore: add release workflow on pr

tejal29:
Should have set it up before.


.github/workflows/pr_release.yaml
```diff
@@ -0,0 +1,139 @@
...
+        version: latest
+       
+
+    - uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./deploy/Dockerfile_slim
+        platforms: linux/amd64,linux/arm64,linux/s390x,linux/ppc64le
+        tags: |
+          gcr.io/kaniko-project/executor:${{ env.GITHUB_SHA }}-slim
```
mattmoor, line 50: We definitely shouldn't be pushing to this project on PRs 🤔 

cc @dlorenc

.github/workflows/pr_release.yaml
```diff
@@ -0,0 +1,139 @@
+name: Build images on pull requests
+
+on: [pull_request]
```
mattmoor, line 3: secrets are not available to `pull_request` but they are available to `pull_request_target` (the script is executed from the base of the PR though)

.github/workflows/pr_release.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+      with:
+        version: latest
+       
+
+    - uses: docker/build-push-action@v2
+      with:
```
mattmoor, line 45: You need to run the docker login action prior to pushing

mattmoor: I guess the other file does this with:
```
    - name: Setup gcloud CLI
      uses: google-github-actions/setup-gcloud@master
      with:
        service_account_key: ${{ secrets.GCR_DEVOPS_SERVICE_ACCOUNT_KEY }}
        project_id: kaniko-project
        export_default_credentials: true
    # Configure docker to use the gcloud command-line tool as a credential helper
    - run: |
        # Set up docker to authenticate
        # via gcloud command-line tool.
        gcloud auth configure-docker
```

mattmoor: However, as I mention above, you cannot access `secrets.GCR...` in `pull_request` workflows, nor do I think we should be pushing things from PRs (especially to the prod project).

.github/workflows/pr_release.yaml
tejal29: yeah push as well as signing is turned off now for project PRs.

.github/workflows/pr_release.yaml
tejal29: yep. I read the fine print 4 hours late :(

PR #1677, opened 2021-06-21
Fix panic in kaniko/pkg/util.mkdirAllWithPermissions

zhangguanzhang:
Signed-off-by: zhangguanzhang <zhangguanzhang@qq.com>

Fixes https://github.com/GoogleContainerTools/kaniko/issues/1657

**Description**

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [ ] Adds integration tests if needed.

_See [the contribution guide](../CONTRIBUTING.md) for more details._

**Reviewer Notes**

- [ ] The code flow looks good. 
- [ ] Unit tests and or integration tests added.

**Release Notes**

Describe any changes here so maintainer can include it in the release notes, or delete this block.

```
Examples of user facing changes:
- kaniko adds a new flag `--registry-repo` to override registry

```

PR #1813, opened 2021-11-06
Fix possible nil pointer derefence in fs_util.go

slrz:
When os.Stat returns an error different from ErrNotExist,
mkdirAllWithPermissions may panic with a nil pointer
derefence due to insufficient error checking.

Avoid the panic by bailing out, returning the error to the
caller.


cmoulliard:
Is this PR the same as #1677 ?

imjasonh (approved):

PR #1842, opened 2021-12-16
Fix the e2e K8s test

mattmoor:
The K8s e2e test has been failing since the move to actions.  This adds a bunch of instrumentation I used to track down the failure.  The biggest change was trying to consolidate test output to go through `t.Log` so that the logs are properly correlated.  I disabled `-v` because I find the resulting interleaving to be incredibly difficult to parse the failures out of.

The crux of the issue was that the kaniko images were in ErrImgPull because the registry wasn't visible to the cluster.  This was fixed by migrating to the minikube registry addon, but the next problem was that hostPath wasn't working properly with the `--driver=docker`.  So I migrated back to what was being successfully used on Travis and 🎉 things pass.

The main "fix" was a few parts:
1. Use the minikube registry addon instead of `registry:2`,
2. Revert to the way minikube was spun up on Travis.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
The e2e test is finally passing 🎉 
```


mattmoor:
Trying a YOLO revert to the old minikube setup, which used `--vm-driver=none` 🤷

mattmoor:
cc @priyawadhwa @imjasonh @dlorenc

integration/images.go
```diff
@@ -218,8 +225,8 @@ func addServiceAccountFlags(flags []string, serviceAccount string) []string {
 	return flags
 }
 
-func (d *DockerFileBuilder) BuildDockerImage(imageRepo, dockerfilesPath, dockerfile, contextDir string) error {
-	fmt.Printf("Building image for Dockerfile %s\n", dockerfile)
+func (d *DockerFileBuilder) BuildDockerImage(t *testing.T, imageRepo, dockerfilesPath, dockerfile, contextDir string) error {
+	t.Logf("Building image for Dockerfile %s\n", dockerfile)
```
imjasonh, line 229: Not `d.logf`?

scripts/minikube-setup.sh
```diff
@@ -15,6 +15,24 @@
...
-minikube start --profile=minikube --driver=docker
\ No newline at end of file
+# conntrack is required for minikube 1.19 and higher for none driver
+if ! conntrack --version &>/dev/null; then
+  echo "WARNING: No contrack is not installed"
+  sudo apt-get update -qq
+  sudo apt-get -qq -y install conntrack
+fi
+
+curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
```
imjasonh, line 25: TODO: Don't depend on latest minikube release?

imjasonh (approved):
👏👏👏

scripts/minikube-setup.sh
mattmoor: This is what it did before and is already strictly more conservative than `minikube-builds` from `master` 😲

integration/images.go
mattmoor: `d.logf` is the outer `t *testing.T`, where what we pass here is from the inner `t.Run`.  I started with `d.logf` and it wasn't partitioning the logs as well as it should have been.  I'd probably rather drop `d.logf` entirely than switch back, since this was the bigger deal IIRC.

scripts/minikube-setup.sh
imjasonh: Oh absolutely, just something to clean up further in the future.

integration/images.go
imjasonh: Ah got it. If there's a way to drop `d.logf` that'd be nice, or at least a comment describing how they're different.

scripts/minikube-setup.sh
mattmoor: Opened: https://github.com/GoogleContainerTools/kaniko/issues/1843

integration/images.go
mattmoor: I looked at where it was still used and it was a place we could switch to a `t.Logf` so I removed most of that stuff in the latest commit and made the type private.

imjasonh (approved):

priyawadhwa:
amazing!! 🎉 🥳

Issue #702, opened 2019-06-20
error checking push permissions when pushing to insecure registry


CeadeS:
**Actual behavior**
When pushing to an insecure registry, process fails with the message:

```
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again
: checking push permission for "adress:port/name:tag": Get https://server:port/v2/: http: server gave HTTP response to HTTPS client
```

**Expected behavior**
Kaniko builds and  pushes Image to registry. Still working at v0.9.0

**To Reproduce**
```
config.yaml
...
args: ["--dockerfile=Dockerfile",    
            "--insecure-registry=YOUR_OWN_INSECURE_REGISTRY",
            "--destination=AN_EXISTING_DESTINATION",            
            "--insecure"]
...
```

vijaySamanuri:
I hit with the same issue when using docker image by tag latest (updated 13 hours ago).
I noticed issue with hostname resolution.

cn-sunletian:
Look at 
Duplicate of #683

priyawadhwa:
Hey @CeadeS could you try again with `gcr.io/kaniko-project/executor: 51734fc3a33e04f113487853d118608ba6ff2b81 `? I'm wondering if #685 will fix your issue.

tejal29:
@CeadeS  and @CeadeS Can you please confirm?

tejal29:
This is now fixed. thank you for your patience.

pvgbabu:
@tejal29 , how did you fixed this issue, could you let us know

just4est:
So??? How to fix this sh... ? There is no answer https://github.com/GoogleContainerTools/kaniko/issues/1179 to

tejal29:
@just4est and @pvgbabu #685 should respect `--insecure`   flag even if the `--insecure-registries` does not contain the destination image registry.
#663 should skip tls verficaition for your insecure registry. 
Can you try `--insecure` and `--skip-tls-verify` ?

tchellomello:
> Hey @CeadeS could you try again with `gcr.io/kaniko-project/executor: 51734fc3a33e04f113487853d118608ba6ff2b81 `? I'm wondering if #685 will fix your issue.

I was hitting the same issue and that image `51734fc3a33e04f113487853d118608ba6ff2b81` fixed the problem!

vitdolphin:
i have the same issue now on the pipeline that worked few days ago.

[...]

PR #1852, opened 2021-12-21
Revert "Run docker_credentials_gcr in warmer (#1780)"

imjasonh:
**Description**

This reverts commit cf4822c31cbfd0a79b9a0c4a59dbc250d258b4c1.

I have a theory in https://github.com/GoogleContainerTools/kaniko/issues/702#issuecomment-949649212 that this change causes the regression in v1.7.0 which also blocks us from cutting new releases.

The original justification in https://github.com/GoogleContainerTools/kaniko/pull/1780 was that the credential helpers aren't available in multi-arch images, but since #1848 we're `go build`ing them ourselves with `GOARCH` set, so that shouldn't be an issue. @tejal29 any more context on that issue might help us come up with a better solution.

In any case, I don't think that change was necessary, so even if it doesn't fix the bug, we should be able to get by without it. With images being pushed for every commit we can test whether this fixes everyone's issues, and if so, we can cut a 1.8.0. If not, we can at least keep digging for the bug.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
Revert change related to GCR credential helpers
```


imjasonh:
Nevermind, I built an image myself at `gcr.io/imjasonh/kaniko@sha256:2fad0b62dcf97853b8afc87c57c5b613b27f91d2911d8e5ab2bec8b8daf40e3a` and it still exhibits the broken behavior, where `:v1.6.0` works for me.

There's a different issue at play here, something between https://github.com/GoogleContainerTools/kaniko/compare/v1.6.0...v1.7.0

PR #1866, opened 2021-12-30
Bump github.com/docker/docker to latest release

imjasonh:
**Description**

```
go get -u github.com/docker/docker
go mod tidy
go mod vendor
```

...fix some build breakages...
...remove `replace` directives...
...add some new `replace` directives...
...fix some build breakages...
...fix some unit test breakages...

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [ ] Adds integration tests if needed.

_See [the contribution guide](../CONTRIBUTING.md) for more details._

**Reviewer Notes**

- [ ] The code flow looks good. 
- [ ] Unit tests and or integration tests added.

**Release Notes**

Describe any changes here so maintainer can include it in the release notes, or delete this block.

```
Update dependency on github.com/docker/docker to v20.10.12
```

PR #1883, opened 2022-01-14
Run GitHub Actions on pushes and PRs to main, not master

imjasonh:
At some point in the last week or so, this repo's default branch was changed from `master` to `main` (by whom, I don't know, there doesn't seem to be any record of it).

Since GitHub Actions workflows were only configured to run on pushes and PRs to `master`, these weren't run, for example, on https://github.com/GoogleContainerTools/kaniko/pull/1866

**Submitter Checklist**

review them:

**Reviewer Notes**

### DIFF
--- a/.github/workflows/images.yaml
+++ b/.github/workflows/images.yaml
@@ -2,9 +2,9 @@ name: Build images
 
 on:
   pull_request:
-    branches: ['master']
+    branches: ['main']
   push:
-    branches: ['master']
+    branches: ['main']
     tags: ['v[0-9]+.[0-9]+.[0-9]+*']
 
 jobs:
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -2,9 +2,9 @@ name: Integration tests
 
 on:
   push:
-    branches: ['master']
+    branches: ['main']
   pull_request:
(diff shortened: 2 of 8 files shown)


priyawadhwa (approved):
Yup this was my bad, sorry I should have opened an issue before I made the change! Thanks for fixing it :)

imjasonh:
> Yup this was my bad, sorry I should have opened an issue before I made the change! Thanks for fixing it :)

No worries, thanks for switching to `main`!

imjasonh:
The remaining integration test failure ([link](https://github.com/GoogleContainerTools/kaniko/runs/4820845499?check_suite_focus=true#step:6:65)) fails because [`docker build`'s behavior](https://docs.docker.com/engine/reference/commandline/build/#git-repositories) when asked to use a Git repo as a build context without a ref specified is to fetch `master` -- and this repo no longer has one of those :( 

Should we:
a. skip the test with a TODO to fix it later, temporarily lose test coverage of this behavior
b. remove the test, ~permanently lose test coverage
c. switch the repo to some other repo that _does_ have a `master`, and take an implicit dependency on this other repo

priyawadhwa:
I would probably go with (a)! (c) is also fine but I'd probably only do it if you already have a repo in mind.

PR #1907, opened 2022-01-31
Remove deploy/cloudbuild-*.yaml files

imjasonh:
As far as I can tell, these are unnecessary, since the equivalent build+push workflows are configured in GitHub Actions. I don't know at what point these became redundant, maybe someone else knows. @priyawadhwa do you know?

In any case, if they are in fact redundant, we should remove them before pushing another release, or else we may get a confusing race condition when pushing if the images aren't fully reproducible for some reason.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
Remove cloudbuild.yaml configs to build images; images are released using GitHub Actions.
```


priyawadhwa:
My guess is that it was just accidentally left behind!

priyawadhwa (approved):

PR #1915, opened 2022-02-02
Restore build args after optimize. Fixes #1910, #1912.

### LINKED ISSUE #1910: Caching does not work with ARG inside the Dockerfile
**Actual behavior**
I am running kaniko with `--cache-copy-layers=true` and it fails to utilize the cache in certain situations.

**Expected behavior**
I expect kaniko to reuse the cache, it seems as if the cache ids are generated wrongly

**To Reproduce**

1. Create the following dockerfile:
```
FROM alpine as build
ARG TEST=123
RUN apk add --no-cache tar
```
2. Run the following command multiple times and observe the cache not being used:
```
docker run -v `pwd`:/workspace gcr.io/kaniko-project/executor:bde904349e9b7797d14f12c7a2657bf79145d18e-debug --dockerfile /workspace/Dockerfile --context dir:///workspace/ --destination 10.0.0.6:5000/test --cache=true --cache-copy-layers=true
```
**Additional Information**
 - Dockerfile: See above
 - Build Context: empty aside from the Dockerfile
 - Kaniko Image (fully qualified with digest) gcr.io/kaniko-project/executor:bde904349e9b7797d14f12c7a2657bf79145d18e-debug (a commit from yesterday)

Looking at the debug output the cache key changes:
```
DEBU[0000] Skip resolving path /cache                   
DEBU[0001] No file found for cache key sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3 stat /cache/sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3: no such file or directory 
DEBU[0001] Image alpine not found in cache              
INFO[0001] Returning cached image manifest              
INFO[0002] Returning cached image manifest              
DEBU[0002] No file found for cache key sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3 stat /cache/sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3: no such file or directory 
DEBU[0002] Image alpine not found in cache              
INFO[0002] Returning cached image manifest              
DEBU[0002] optimize: composite key for command ARG TEST=123 {[sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3 |1 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin ARG TEST=123]} 
DEBU[0002] optimize: cache key for command ARG TEST=123 186ec9fee220d1452a3af884e27fbb20dd3ef21eca7b68aaa316b058359fe73d 
DEBU[0002] optimize: composite key for command RUN apk add --no-cache tar {[sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3 |1 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin ARG TEST=123 |2 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin TEST=123 RUN apk add --no-cache tar]} 
DEBU[0002] optimize: cache key for command RUN apk add --no-cache tar a281ac04c35f422a7956387c1d32787139a79b22a87edd9f7933fbdd3fd65819 
INFO[0002] Checking for cached layer 10.0.0.6:5000/test/cache:a281ac04c35f422a7956387c1d32787139a79b22a87edd9f7933fbdd3fd65819... 
DEBU[0002] Failed to retrieve layer: GET http://10.0.0.6:5000/v2/test/cache/manifests/a281ac04c35f422a7956387c1d32787139a79b22a87edd9f7933fbdd3fd65819: MANIFEST_UNKNOWN: manifest unknown; map[Tag:a281ac04c35f422a7956387c1d32787139a79b22a87edd9f7933fbdd3fd65819] 
INFO[0002] No cached layer found for cmd RUN apk add --no-cache tar 
DEBU[0002] Key missing was: sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3-|1-PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin-ARG TEST=123-|2-PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin-TEST=123-RUN apk add --no-cache tar 
INFO[0002] Unpacking rootfs as cmd RUN apk add --no-cache tar requires it. 
DEBU[0003] build: composite key for command ARG TEST=123 &{[sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3 |2 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin TEST=123 ARG TEST=123]} 
DEBU[0003] build: cache key for command ARG TEST=123 6c4bf2e8db648ae9c9b1db201aff474c5302bddb61b2444debd20ae0a00ebdb1 
INFO[0003] RUN apk add --no-cache tar                   
INFO[0003] args: [-c apk add --no-cache tar]            
INFO[0003] Running: [/bin/sh -c apk add --no-cache tar] 
DEBU[0004] build: composite key for command RUN apk add --no-cache tar &{[sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3 |2 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin TEST=123 ARG TEST=123 |2 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin TEST=123 RUN apk add --no-cache tar]} 
DEBU[0004] build: cache key for command RUN apk add --no-cache tar a4329b9603a45f299011c221612a695286390059448babb87652103053db4209 
INFO[0004] Pushing layer 10.0.0.6:5000/test/cache:a4329b9603a45f299011c221612a695286390059448babb87652103053db4209 to cache now 
INFO[0004] Pushing image to 10.0.0.6:5000/test/cache:a4329b9603a45f299011c221612a695286390059448babb87652103053db4209 
INFO[0004] Pushed 10.0.0.6:5000/test/cache@sha256:ba36e231d062103b990c1106648f707f841039c6ade8cbb435b045af88479548 
DEBU[0004] mapping digest sha256:e53ebcc21ce659d505b6b9bf7fd251af6d118cf420abbc0ce94ec825b8de9a0f to cachekey a281ac04c35f422a7956387c1d32787139a79b22a87edd9f7933fbdd3fd65819 
```
 
### LINKED ISSUE #1912: Kaniko allows to access undefined variables when cache is enabled
**Actual behavior**
Kaniko allows to access ARGS/ENV variables that are not yet defined when `--cache=true` is used

**Expected behavior**
Undefined variables should behave as one would expect from a shell

**To Reproduce**
1. Use this dockerfile:
```
FROM alpine as build
ARG TEST=123
RUN echo $TESTNOTSET
ARG TESTNOTSET=ups
```
2. Run with & without `--cache`

Without cache:
```
INFO[0002] Running: [/bin/sh -c echo $TESTNOTSET]       

INFO[0002] Taking snapshot of full filesystem...        
```

with cache:
```
INFO[0002] Running: [/bin/sh -c echo $TESTNOTSET]       
ups
INFO[0002] Taking snapshot of full filesystem...        
```

**Additional Information**
 - Dockerfile: see above
 - Build Context: empty
 - Kaniko Image (fully qualified with digest) gcr.io/kaniko-project/executor:bde904349e9b7797d14f12c7a2657bf79145d18e-debug
 
Note, this is probably related to #1910 because `optimize()` accumulates *all* ENV/ARG variables before they should be used

apollo13:
Fixes #1910
Fixes #1912

**Description**

After executing metadata commands `optimize()` needs to restore the build args so `build()` does not start with build args that it should not yet know.

I do not have found the correct location for the tests, hints welcome, thank you!

**Submitter Checklist**

**Reviewer Notes**

**Release Notes**

```
- Fix cache key calculation involving ARG and ENV
```

### DIFF
--- a/pkg/dockerfile/buildargs.go
+++ b/pkg/dockerfile/buildargs.go
@@ -17,7 +17,6 @@ limitations under the License.
 package dockerfile
 
 import (
-	"sort"
 	"strings"
 
 	d "github.com/docker/docker/builder/dockerfile"
@@ -56,9 +55,7 @@ func (b *BuildArgs) ReplacementEnvs(envs []string) []string {
 	resultEnv := make([]string, len(envs))
 	copy(resultEnv, envs)
 	filtered := b.FilterAllowed(envs)
-	resultEnv = append(resultEnv, filtered...)
-	sort.Strings(resultEnv)
-	return resultEnv
+	return append(resultEnv, filtered...)
 }
 
 // AddMetaArgs adds the supplied args map to b's allowedMetaArgs
--- a/pkg/dockerfile/buildargs.go
+++ b/pkg/dockerfile/buildargs.go
@@ -17,6 +17,7 @@ limitations under the License.
 package dockerfile
 
 import (
+	"sort"
 	"strings"
 
 	d "github.com/docker/docker/builder/dockerfile"
@@ -51,8 +52,13 @@ func (b *BuildArgs) Clone() *BuildArgs {
 
 // ReplacementEnvs returns a list of filtered environment variables
 func (b *BuildArgs) ReplacementEnvs(envs []string) []string {
+	// Ensure that we operate on a new array and do not modify the underlying array
+	resultEnv := make([]string, len(envs))
+	copy(resultEnv, envs)
 	filtered := b.FilterAllowed(envs)
-	return append(envs, filtered...)
+	resultEnv = append(resultEnv, filtered...)
+	sort.Strings(resultEnv)
+	return resultEnv
 }
 
 // AddMetaArgs adds the supplied args map to b's allowedMetaArgs
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -181,6 +182,9 @@ func initConfig(img partial.WithConfigFile, opts *config.KanikoOptions) (*v1.Con
 func (s *stageBuilder) populateCompositeKey(command fmt.Stringer, files []string, compositeKey CompositeCache, args *dockerfile.BuildArgs, env []string) (CompositeCache, error) {
 	// First replace all the environment variables or args in the command
 	replacementEnvs := args.ReplacementEnvs(env)
+	// The sort order of `replacementEnvs` is basically undefined, sort it
+	// so we can ensure a stable cache key.
+	sort.Strings(replacementEnvs)
 	resolvedCmd, err := util.ResolveEnvironmentReplacement(command.String(), replacementEnvs, false)
 	if err != nil {
 		return compositeKey, err
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -227,6 +227,11 @@ func (s *stageBuilder) optimize(compositeKey CompositeCache, cfg v1.Config) erro
 	if !s.opts.Cache {
 		return nil
 	}
+	var buildArgs = s.args.Clone()
+	// Restore build args back to their original values
+	defer func() {
+		s.args = buildArgs
+	}()
 
 	stopCache := false
 	// Possibly replace commands with their cached implementations.


apollo13:
@imjasonh Any chance you could approve me so I can get the tests running here (I'd also love a review since it fixes a few ugly caching bugs, but being able to run the tests would already be great).

imjasonh:
Absolutely! Thanks for working on this, tests running now, I'll review shortly🙏

apollo13:
Lovely, thanks! If you have any questions on why I did things the way I did please let me know. I am a noob in GO and finding that slice issue in `ReplacementEnvs` cost me quite a bit :) I still haven't figured out how and where to add tests for this, any existing test that I could copy and start from would be great!

pkg/executor/build.go
```diff
@@ -227,6 +227,11 @@ func (s *stageBuilder) optimize(compositeKey CompositeCache, cfg v1.Config) erro
 	if !s.opts.Cache {
 		return nil
 	}
+	var buildArgs = s.args.Clone()
+	// Restore build args back to their original values
```
imjasonh, line 231: Why do we need to restore them?

pkg/dockerfile/buildargs.go
```diff
@@ -51,8 +52,13 @@ func (b *BuildArgs) Clone() *BuildArgs {
 
 // ReplacementEnvs returns a list of filtered environment variables
 func (b *BuildArgs) ReplacementEnvs(envs []string) []string {
+	// Ensure that we operate on a new array and do not modify the underlying array
+	resultEnv := make([]string, len(envs))
+	copy(resultEnv, envs)
 	filtered := b.FilterAllowed(envs)
-	return append(envs, filtered...)
+	resultEnv = append(resultEnv, filtered...)
+	sort.Strings(resultEnv)
```
imjasonh, line 60: In general I'm not totally convinced that it's safe to sort env vars. I understand why we'd want that for cache key generation, but `FOO=x FOO=b echo $FOO` has demonstrably different results if you sort the env vars. The output of those should produce different cached results as well.

Am I misunderstanding the problem that this solves?

imjasonh (commented):
Thanks for making this change! Is there any way we can test this, in such a way that before this change the test would fail?

This would help demonstrate how this change helps fix an issue, and help guard against future regressions.

pkg/executor/build.go
apollo13: I'll try to to illustrate it with an example.

Take 
```
ARG ARG1=val1
ARG ARG2=val2
```
as an example. Going through those two stages `optimize` will generate the following cache keys:
```
'' -- starting empty
ARG ARG1=val1 results in roughly 'ARG ARG1=val1'
ARG ARG2=val2 results now in 'ARG1=val1 ARG ARG2=val2'
```
As you can see the `ARG1` will become an env variable and as such is part of the following cache keys (this is defined in the dockerfile spec that args become envs for the subsequent calls).

Now after `optimize()` the `s.args` really contains `ARG1` & `ARG2` (if you look further down in `optimize` the meta commands are actually executed and now change the state).

Now if we do not restore the build args to the state when optimize started, the actual build steps will start with `ARG1` & `ARG2` prepopulated. This has two issues, taking the above Dockerfiles the cache keys will now roughly look like this:
```
'ARG1=val1 ARG2=val2' -- not starting empty because build args already defined
ARG ARG1=val1 results in roughly 'ARG1=val1 ARG2=val2 ARG ARG1=val1'
ARG ARG2=val2 results now in 'ARG1=val1 ARG2=val2 ARG ARG2=val2'
```
as you can see the cache key is now completely different, so this cached layer will never be found (because optimize looked for a different layer due to the different cache key). -- This can be seen in https://github.com/GoogleContainerTools/kaniko/issues/1910#issuecomment-1027947684 ; during the actual build TEST is always set even if not executed yet -- which is also the root cause of https://github.com/GoogleContainerTools/kaniko/issues/1912

apollo13:
>  Is there any way we can test this, in such a way that before this change the test would fail?

I wish, I was hoping that you could tell me where similar tests can be found so I can dig around there. I have a few minimal Dockerfiles that show the issues; but not sure how to rewrite those to tests.

apollo13:
To provide a few examples; executing kaniko like this (without my patch):
```
docker run -v `pwd`:/workspace gcr.io/kaniko-project/executor:bde904349e9b7797d14f12c7a2657bf79145d18e-debug --dockerfile /workspace/Dockerfile --context dir:///workspace/ --destination 10.0.0.6:5000/test --cache=true --cache-copy-layers=true
```
with the following Dockerfile:
```
FROM alpine as build
ARG TEST=123
RUN echo hi
```
always outputs hi -> no caching of the run layer.

Running it with:
```
FROM alpine as build
ARG TEST=123
RUN echo $TESTNOTSET
ARG TESTNOTSET=ups
```
will echo ups when running with a cache enabled, even though $TESTNOTSET should not be set. 

I'd love to somehow have those two files as test.

pkg/dockerfile/buildargs.go
vinman1: @imjasonh, @apollo13, instead of sorting the result list here which would affect all callers of the `ReplacementEnvs` function (for which there seems to be many), couldn't the sort be done in the `populateCompositeKey` function in `pkg/executor/build.go` just before [line 184](https://github.com/apollo13/kaniko/blob/1dfda2bb812c0d87ae4b2d1372b261f3ef5ff685/pkg/executor/build.go#L184) when the result is used next?

That way, we know that the sorting will only be applied in the use cases of a cache key.

apollo13: I had that before, but I think just sorting `filtered` in here should be fine.

pkg/executor/build.go
apollo13: Oh and we also might need similar handling for `ENV` statements.

apollo13: Nevermind, `ENV` is fine, this issue seems only to affect `ARG`

vinman1: @imjasonh, as I understand it, the `optimize` function actually executes the meta-only commands in order to properly determine the environment state for the purposes of generating a meaningful cache key ([build.go, line 283](https://github.com/apollo13/kaniko/blob/1dfda2bb812c0d87ae4b2d1372b261f3ef5ff685/pkg/executor/build.go#L283).

I assume this was leaving the `s.args` member in a dirty state for the rest of the execution of the `build` method, and therefore was the cause of the first problem fixed by this PR.

pkg/dockerfile/buildargs.go
apollo13: Pushed a new commit moving the sort into `populateCompositeKey`

pkg/executor/build.go
apollo13: Yes that is correct.

imjasonh:
> I wish, I was hoping that you could tell me where similar tests can be found so I can dig around there. I have a few minimal Dockerfiles that show the issues; but not sure how to rewrite those to tests.

I'm afraid it's the blind leading the blind here, I'm not very familiar with all the workings of this project either 😕 

In the absence of new CI tests (and assuming existing tests pass) I'm fine taking your word that this is the more correct behavior. If anybody is able to figure out where to add a test for this though I'll owe you a beer. 🍻

apollo13:
That is great, I promise to add tests once I know how / where :) Beer would be nice if we ever manage to meet in person someday somewhere.

PR #1994, opened 2022-03-16
hasher: hash security.capability attributes

### LINKED ISSUE #1851: Kaniko does not persist setcap side effects
**Actual behavior**
`setcap` command does not persist to the finalized image. Running `getcap` in the finalized image returns an empty result.

**Expected behavior**
`getcap` should return the newly attached capabilities

**To Reproduce**
1. Create a Dockerfile with the following:
```
FROM python:3.9.9-slim-buster AS python
RUN apt-get update && apt-get install --assume-yes libcap2-bin nmap
RUN setcap cap_net_raw,cap_net_admin,cap_net_bind_service+eip $(which nmap)
RUN echo $(getcap $(which nmap))
```
2. See that during the Kaniko build, the echo command indeed produces the desired result.
3. Running the resulted image (from loading the resulting .tar file) and running `getcap` produces empty output

**Additional Information**
Used Kaniko 1.7-debug, and tried all sorts of flags, including single-snapshot and --snapshotMode=full
 
zhouhaibing089:
Fixes #1851 

**Description**

In Dockerfile, if there is something like:

```
RUN setcap cap_net_raw=+ep /path/to/binary
```

kaniko won't detect that there is a change on file `/path/to/binary` and
thus discards this layer. This patch allows the hasher function to
actually look at `security.capability` extended attributes.

The function `Lgetxattr` uses https://github.com/moby/moby/blob/c72c1ca62cd20e74582a01b8a54a58ae7f646d4c/pkg/system/xattrs_linux.go#L8

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
default hahser (--snapshotMode=full) now hashes security.capability attribute 
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/minio/highwayhash"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 // Hasher returns a hash function, used in snapshotting to determine if a file has changed
@@ -56,6 +57,10 @@ func Hasher() func(string) (string, error) {
 		h.Write([]byte(strconv.FormatUint(uint64(fi.Sys().(*syscall.Stat_t).Gid), 36)))
 
 		if fi.Mode().IsRegular() {
+			capability, _ := Lgetxattr(p, "security.capability")
+			if capability != nil {
+				h.Write(capability)
+			}
 			f, err := os.Open(p)
 			if err != nil {
 				return "", err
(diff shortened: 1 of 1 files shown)


zhouhaibing089:
@imjasonh: Could you help take a look when you get a chance? Thanks~

pkg/util/util.go
```diff
@@ -172,3 +177,28 @@ func Retry(operation retryFunc, retryCount int, initialDelayMilliseconds int) er
 
 	return err
 }
+
+func Lgetxattr(path string, attr string) ([]byte, error) {
```
imjasonh, line 181: It doesn't look like this function needs to be exported, can we rename to `lgetxattr` and add a comment?

I'd also love to have a test covering this behavior, since it's pretty inscrutable to me personally.

zhouhaibing089: Sure, I can add a comment there.

Regarding tests, I can take a look at how to write a test(either a unit test or e2e test). But if you have any pointers or suggestions, that'd be great.

imjasonh (approved):

PR #1996, opened 2022-03-20
Move and fix GetContainerRuntime check from bpfd proc

### LINKED ISSUE #1592: Container detection fails on cgroup v2 devices

MyIgel:
Fixes #1592 (again)

**Description**
This adds a slimmed down version of https://github.com/genuinetools/bpfd/blob/a4bfa5e3e9d1bfdbc56268a36a0714911ae9b6bf/proc/proc.go as the original package is unmaintained (~3y no updates of its 0.0.1 release) and kaniko only needs the container runtime detection, see https://github.com/GoogleContainerTools/kaniko/pull/1686#issuecomment-1073269412 . To fix the problem, it has #1686 already applied to it.

The alternative to this change would be to rebuild this functionality, but I'm not totally sure if I could do that by my own due to a lack of go experience but I could try if it would be better to have a "kaniko implementation" of the functionality.

**Submitter Checklist**
- [x] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr) (at least the original ones)
- [x] Adds integration tests if needed. (nope)

**Reviewer Notes**
- [ ] Unit tests and/or integration tests added.


imjasonh:
```
Boilerplate missing in:
././pkg/util/proc_test.go
././pkg/util/proc.go
```

Please add the boilerplate text to the top of these new files.

Also, could we maybe move them into `pkg/util/proc`? Having a single huge `util` package with so much different functionality in it is kind of a smell. I'd like to clean that up in general in the future, and this would be a good step.

MyIgel:
I can definitely move it but wouldn't adding the boilerplate mean that it gets relicensed? I mean I actually don't think that this would be an issue as its MIT to Apache 2 and not an substantial part of the code with a relevantly complex software part but whats your opinion here @imjasonh?

MyIgel:
I added the boilerplate and did some more cleanup, that should be sufficient i guess :D

PR #1997, opened 2022-03-20
feat: kaniko dir config option

### LINKED ISSUE #1945: make KanikoDir a configuration option
At the moment the directory `/kaniko` is hard coded in `constants.go` as
~~~go
KanikoDir = "/kaniko"
~~~

kaniko stores snapshots in this directory. this is a problem, if the volume storage is limited in the build environment. we use gitlab-ci in a k8s environment and connect the default `/builds` directory to a ephemeral volume, so when running a ci job, the build directory has >10GB storage, but the other directories are normal directories in the job-pod and so they use the disk storage of the k8s-nodes. 

when building big images (and also have a lot of parallel jobs running), we sometimes have `no space left on device` because the snapshot in `/kaniko` is to big.

It would be great to have a configuration option, so kaniko stores the snapshots in another directory (`/builds` in our case) so we can use the mounted ephemeral storage. i tried the `--tarPath` option but it looks like that in addition to the tar, the snapshots are also written to `/kaniko`.

in #1881 it was also mentioned to make this directory configurable, but the whole issue has another aspect so i created a new issue.

jdockerty:
Fixes #1945 

**Description**

Moving away from a hard-coded `/kaniko` directory, this was in `constants.go`, to one which is configurable by the user - this can be done either through a flag, `--kaniko-dir`, or an environment variable, `KANIKO_DIR`.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

- `executor` adds a new flag `--kaniko-dir` or environment variable `KANIKO_DIR` to override the default directory that kaniko uses.

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -187,7 +187,7 @@ func addKanikoOptionsFlags() {
 	RootCmd.PersistentFlags().BoolVarP(&opts.SkipTLSVerifyPull, "skip-tls-verify-pull", "", false, "Pull from insecure registry ignoring TLS verify")
 	RootCmd.PersistentFlags().IntVar(&opts.PushRetry, "push-retry", 0, "Number of retries for the push operation")
 	RootCmd.PersistentFlags().IntVar(&opts.ImageFSExtractRetry, "image-fs-extract-retry", 0, "Number of retries for image FS extraction")
-	RootCmd.PersistentFlags().StringVarP(&opts.KanikoDir, "kaniko-dir", "k", "/kaniko", "Path to the kaniko directory")
+	RootCmd.PersistentFlags().StringVarP(&opts.KanikoDir, "kaniko-dir", "", "/kaniko", "Path to the kaniko directory")
 	RootCmd.PersistentFlags().StringVarP(&opts.TarPath, "tarPath", "", "", "Path to save the image in as a tarball instead of pushing")
 	RootCmd.PersistentFlags().BoolVarP(&opts.SingleSnapshot, "single-snapshot", "", false, "Take a single snapshot at the end of the build.")
 	RootCmd.PersistentFlags().BoolVarP(&opts.Reproducible, "reproducible", "", false, "Strip timestamps out of the image to make it reproducible")
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -68,7 +68,7 @@ var RootCmd = &cobra.Command{
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 		if cmd.Use == "executor" {
 
-			if err := checkKanikoDir(os.Getenv("KANIKO_DIR")); err != nil {
+			if err := checkKanikoDir(config.KanikoDir); err != nil {
 				return err
 			}
 
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -182,6 +182,7 @@ func addKanikoOptionsFlags() {
 	RootCmd.PersistentFlags().BoolVarP(&opts.SkipTLSVerifyPull, "skip-tls-verify-pull", "", false, "Pull from insecure registry ignoring TLS verify")
 	RootCmd.PersistentFlags().IntVar(&opts.PushRetry, "push-retry", 0, "Number of retries for the push operation")
 	RootCmd.PersistentFlags().IntVar(&opts.ImageFSExtractRetry, "image-fs-extract-retry", 0, "Number of retries for image FS extraction")
+	RootCmd.PersistentFlags().StringVarP(&opts.KanikoDir, "kaniko-dir", "k", "/kaniko", "Path to the kaniko directory")
 	RootCmd.PersistentFlags().StringVarP(&opts.TarPath, "tarPath", "", "", "Path to save the image in as a tarball instead of pushing")
 	RootCmd.PersistentFlags().BoolVarP(&opts.SingleSnapshot, "single-snapshot", "", false, "Take a single snapshot at the end of the build.")
 	RootCmd.PersistentFlags().BoolVarP(&opts.Reproducible, "reproducible", "", false, "Strip timestamps out of the image to make it reproducible")
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -290,16 +290,16 @@ func resolveEnvironmentBuildArgs(arguments []string, resolver func(string) strin
 // copy Dockerfile to /kaniko/Dockerfile so that if it's specified in the .dockerignore
 // it won't be copied into the image
 func copyDockerfile() error {
-	if _, err := util.CopyFile(opts.DockerfilePath, constants.DockerfilePath, util.FileContext{}, util.DoNotChangeUID, util.DoNotChangeGID); err != nil {
+	if _, err := util.CopyFile(opts.DockerfilePath, config.DockerfilePath, util.FileContext{}, util.DoNotChangeUID, util.DoNotChangeGID); err != nil {
 		return errors.Wrap(err, "copying dockerfile")
 	}
 	dockerignorePath := opts.DockerfilePath + ".dockerignore"
 	if util.FilepathExists(dockerignorePath) {
-		if _, err := util.CopyFile(dockerignorePath, constants.DockerfilePath+".dockerignore", util.FileContext{}, util.DoNotChangeUID, util.DoNotChangeGID); err != nil {
+		if _, err := util.CopyFile(dockerignorePath, config.DockerfilePath+".dockerignore", util.FileContext{}, util.DoNotChangeUID, util.DoNotChangeGID); err != nil {
 			return errors.Wrap(err, "copying Dockerfile.dockerignore")
 		}
 	}
-	opts.DockerfilePath = constants.DockerfilePath
+	opts.DockerfilePath = config.DockerfilePath
 	return nil
 }
 
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -68,8 +68,7 @@ var RootCmd = &cobra.Command{
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 		if cmd.Use == "executor" {
 
-			err := checkKanikoDir(os.Getenv("KANIKO_DIR"))
-			if err != nil {
+			if err := checkKanikoDir(os.Getenv("KANIKO_DIR")); err != nil {
 				return err
 			}
 
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -244,13 +244,11 @@ func addHiddenFlags(cmd *cobra.Command) {
 func checkKanikoDir(dir string) error {
 	if dir != constants.DefaultKanikoPath {
 
-		err := os.MkdirAll(dir, os.ModeDir)
-		if err != nil {
+		if err := os.MkdirAll(dir, os.ModeDir); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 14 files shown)


jdockerty:
@imjasonh I can't set you as a reviewer myself unfortunately 😢 

Could you approve the workflows too, please 😄

imjasonh:
This looks good! We'll see if the tests pass, and I think we can merge this. 👍 

But it's not the whole story either. If a build has `KANIKO_DIR` set, we need to do something to ensure that the contents of `/kaniko` in the image (the executor binary, the cred helpers, the Dockerfile, etc.) get moved into `KANIKO_DIR` so they don't end up in the final image. That shouldn't necessarily block this PR, but I want to make sure we have a plan for making and testing that next change before we proceed.

jdockerty:
Much appreciated @imjasonh 

Looks as though it had an issue with the credentials on the PR itself, `invalid version: Get "https://proxy.golang.org/github.com/....": net/http: TLS handshake timeout`.

Is this what you mean by the cred helpers not being in the correct place? Unsure whether it's a temporary network issue too on the GitHub Actions side, I know they've had a couple of problems recently, with it being a TLS timeout, I'll take a dig into it a little bit later, just to be absolutely sure.

jdockerty:
I believe that makes sense, so that change (I'll raise another PR later on) would be specifically targeted towards the files within `deploy/`.

I had a skim over the `dockerfile` package, but I don't believe this would need to be altered here? I.e. there is no alteration required to any instructions there. Although I might be misunderstanding it.

Do you think it might be best served to have another issue for it as well? Feel free to assign me onto that as well as I'll link back to your comment for reference in the resulting PR, so others can see a full conversation history as it evolved.

imjasonh:
> I believe that makes sense, so that change (I'll raise another PR later on) would be specifically targeted towards the files within `deploy/`.
> 
> I had a skim over the `dockerfile` package, but I don't believe anything would need to be altered in here? I.e. there is no alteration required to any instructions there. Although I might be misunderstanding it.

Files inside `deploy/` wouldn't change. Those files determine how the Kaniko image is produced, and they should remain as-is. The executor binary, cred helpers, etc, should remain in `/kaniko` in the image.

The change we need is for the executor to understand that if it's not in `KANIKO_DIR` (because `KANIKO_DIR != /kaniko`), that it should move there so it doesn't end up in the final image.

That _might_ actually just be as easy as adding `os.MkdirAll(os.Getenv("KANIKO_DIR"))` and `os.Rename("/kaniko", os.Getenv("KANIKO_DIR"))` when the executor starts. The bigger part of that change will be ensuring that this doesn't break any existing behavior.

> Do you think it might be best served to have another issue for it as well? Feel free to assign me onto that as well as I'll link back to your comment for reference in the resulting PR, so others can see a full conversation history as it evolved.

All of this is part of the overarching work to make `/kaniko` configurable, so I think it belongs under the current issue.

jdockerty:
> The change we need is for the executor to understand that if it's not in KANIKO_DIR (because KANIKO_DIR != /kaniko), that it should move there so it doesn't end up in the final image.

That makes a lot of sense, I was getting confused in thinking we'd need to modify the actual image build process itself.

I've added a small check now. The unit tests are passing locally from `make test`, although `make integration-test` is seemingly not able to create the `test_dir` data as expected, even without setting `KANIKO_DIR`. I'd like to check it from within CI too so I can be sure I'm not missing anything that I shouldn't be. 

I believe there's certainly some extra test cases to be done. I'd be hesitant to test the `checkKanikoDir` function itself, since this would practically be testing the `os` package again - although I could be wrong here. I'd probably add some more integration tests, since we'd be setting the environment, reading/writing files based on `KANIKO_DIR` etc., although I could be overcomplicating it.

imjasonh (commented):
This looks promising! I'll try to figure out how to run a test where `KANIKO_DIR` is set to something else, to make sure this works. Thanks for working on this!

jdockerty:
No worries at all Jason, much appreciated for your help too!

imjasonh:
Looking at the integration tests it looks like it should be fairly straightforward to add one that exercises `KANIKO_DIR`. For example, based on this test:

https://github.com/GoogleContainerTools/kaniko/blob/2a8c565dd55c8657bb274a12d993cfb618748687/integration/integration_test.go#L338

You could do something like:

```
func TestKanikoDir(t *testing.T) {
	...

	// Build with kaniko
	kanikoImage := GetKanikoImage(config.imageRepo, "Dockerfile_registry_mirror")
	dockerRunFlags := []string{"run", "--net=host"}
	dockerRunFlags = addServiceAccountFlags(dockerRunFlags, config.serviceAccount)
	dockerRunFlags = append(dockerRunFlags, ExecutorImage,
		"-f", dockerfile,
		"-d", kanikoImage,
		"-e", "KANIKO_DIR=/not-kaniko",
		"-c", fmt.Sprintf("git://%s", repo))
	...
}
```

You can run that with `go test ./integration/... -test.run=TestKanikoDir -repo=gcr.io/your-repo`

If that build works, then it tell us that setting `KANIKO_DIR` doesn't _break_ any builds. We should also try to check that the image it produces doesn't contain anything in `/not/kaniko` or `/kaniko`. That would give us a good idea whether this

jdockerty:
That's super useful, much appreciated Jason. I've fixed the `mkdir` error now, I realised I was passing in `os.Getenv` to the check, rather than the config.KanikoDir option where we retrieve the environment via a function, so that is working nicely now.

If you don't mind, I'd like to quickly check within CI, as I'm seeing a failure, even on the `main` branch with `container-diff`, so I want to rule out that there is some misconfiguration of `docker` or otherwise on my machine.

```sh
    integration_test.go:414: []integration.diffOutput differ (-got, +want):   []integration.diffOutput{
                {
                        Image1:   "localhost:5000/docker-dockerfile_registry_mirror",
                        Image2:   "localhost:5000/kaniko-dockerfile_registry_mirror",
                        DiffType: "File",
                        Diff: &integration.fileDiffResult{
                                Adds: nil,
        -                       Dels: []integration.fileDiff{{Name: "/etc/machine-id", Size: 33}},
        +                       Dels: nil,
                        },
                }
```

For some reason, the `machine-id` is being mounted in, I don't believe it should be though. Especially with `kaniko`, it might have something to do with my setup, but I'd like to 100% confirm as it does seem rather unrelated.

imjasonh:
> If you don't mind, I'd like to quickly sanity check with the CI, as I'm seeing a failure, even on the `main` branch, with `container-diff`, so I want to rule out that there is some misconfiguration of `docker` or otherwise on my machine.
> 
> ```shell
>                         Diff: &integration.fileDiffResult{
>                                 Adds: nil,
>         -                       Dels: []integration.fileDiff{{Name: "/etc/machine-id", Size: 33}},
>         +                       Dels: nil,
>                         },
>                 }
> ```
> 
> For some reason, the `machine-id` is being mounted in, I don't believe it should be though. Especially with `kaniko`, it might have something to do with my setup, but I'd like to 100% confirm as it does seem rather unrelated, I could quite easily be wrong though.

That's definitely weird! It doesn't seem related to your change though, so I wouldn't worry about it for now.

jdockerty:
The tests passed in the CI, so it looks like it was some weird configuration on my end - I'm developing on a VM, so I wonder whether there's any funkiness going on there with `docker`.

cmd/executor/cmd/root.go
```diff
@@ -182,6 +187,7 @@ func addKanikoOptionsFlags() {
 	RootCmd.PersistentFlags().BoolVarP(&opts.SkipTLSVerifyPull, "skip-tls-verify-pull", "", false, "Pull from insecure registry ignoring TLS verify")
 	RootCmd.PersistentFlags().IntVar(&opts.PushRetry, "push-retry", 0, "Number of retries for the push operation")
 	RootCmd.PersistentFlags().IntVar(&opts.ImageFSExtractRetry, "image-fs-extract-retry", 0, "Number of retries for image FS extraction")
+	RootCmd.PersistentFlags().StringVarP(&opts.KanikoDir, "kaniko-dir", "k", "/kaniko", "Path to the kaniko directory")
```
imjasonh, line 190: I don't think we need the shorthand flag. This is hopefully a pretty rare use case, and saving `-k` for something later will probably be useful.

integration/integration_test.go
```diff
@@ -375,6 +375,45 @@ func TestBuildViaRegistryMirrors(t *testing.T) {
...
+	}
+
+	// Build with kaniko
+	kanikoImage := GetKanikoImage(config.imageRepo, "Dockerfile_registry_mirror")
+	dockerRunFlags := []string{"run", "--net=host"}
+	dockerRunFlags = addServiceAccountFlags(dockerRunFlags, config.serviceAccount)
+	dockerRunFlags = append(dockerRunFlags, ExecutorImage,
+		"-f", dockerfile,
+		"-d", kanikoImage,
+		"-k", "/not-kaniko",
```
imjasonh, line 401:
```suggestion
		"--kaniko-dir", "/not-kaniko",
```

integration/integration_test.go
```diff
@@ -375,6 +375,45 @@ func TestBuildViaRegistryMirrors(t *testing.T) {
 	checkContainerDiffOutput(t, diff, expected)
 }
 
+func TestKanikoDir(t *testing.T) {
```
imjasonh, line 378:
```suggestion
// TestKanikoDir tests that a build that sets --kaniko-dir produces the same output as the equivalent docker build.
func TestKanikoDir(t *testing.T) {
```

imjasonh (commented):
This looks really good! The integration test especially makes me pretty confident this isn't going to break any existing behavior. 👍

cmd/executor/cmd/root.go
jdockerty: That makes sense, my thinking was that it's quite an impactful (maybe not?) change, so having a shorthand for it is quite nice. I agree though that there may well be something else that comes later which is best reserved for `-k`.

Shall swap this out!

jdockerty:
Yeah absolutely, the extra test makes me less anxious about it 😆 Much appreciated for pointing me in the right direction on the example test to follow!

I've added your suggested changes now too, no worries on the styling side. I think I'll actually start implementing what you said in my own stuff, I quite like it 😄

imjasonh (approved):

shpprfb:
@jdockerty I'm very excited about this new functionality but I'm seeing issues with this PR as others are mentioning in #1945.

1. --kaniko-dir seems to not do anything at all, I can pass the flag and see /kaniko is still filling up with snapshots and such.
2. Setting KANIKO_DIR as an env var seems to trigger the codepath in checkKanikoDir (https://github.com/GoogleContainerTools/kaniko/blob/76a54a031d5f5c1745c7124f7864516c3c15a21f/cmd/executor/cmd/root.go#L247-L255) for doing the rename, but it always fails with "file exists". This happens whether the destination previously exists (like mounting a volume in kubernetes), or when passing a dirname that I expect kaniko to create for me(although Im not sure what the usefulness of that would be).

I made a quick playground repro of this issue here https://play.golang.com/p/ev1FjrASYAp

Any ideas on how to handle this? Maybe if the destination doesn't exist, the rename is easy and will succeed as long as we don't create it. If the destination does exist, I think we may need to recursively copy the contents of /kaniko there.

jdockerty:
@shpprfb  I do believe I know where the issue is of it not picking it up, it was an oversight by me.

[Here](https://github.com/GoogleContainerTools/kaniko/blob/76a54a031d5f5c1745c7124f7864516c3c15a21f/cmd/executor/cmd/root.go#L71) it is passing the `config` options value. I've raised #2067, which I'll use to investigate further during some free evenings. Then I'll probably further investigate into some other things with Jason, as I believe I've missed some things, even though the tests previously passed for this feature to be merged.

Thanks for raising this 😄

PR #2002, opened 2022-03-21
Add SVG logos

sisp:
**Description**

I've added SVG images for both logos and re-exported the corresponding PNG images based on the SVGs for them to be in sync. The SVGs are not pixel-identical with the original PNGs:

- The "K" in the rounded square was not perfectly vertically centered, I've fixed that.
- The curved lines in the "K" in the rounded square had varying line width, now the line width is fixed.
- The curved lines in the "K" didn't seem to follow a simple/obvious shape, they seemed somewhat arbitrarily curved. Assuming the image size of the monochrome PNG (800x258), I've derived them from the segment of a circle with radius 100 around (126,29) (top circle) and (126,229) (bottom circle).
- The overall logo (rounded square + text) were not perfectly centered horizontally, I've moved the text a bit to the right, so it's centered now.
- The font appears to be "Akrobat SemiBold" with negative letter spacing. I've used this font to re-create the text. There are some minor differences between the original text and the new one, in particular the dot of the "i" is now slightly higher.

I hope you like them despite the minor differences from the original logos.


sisp:
Awesome, I'm glad you like them. :tada: Thanks for merging the PR so promptly!

PR #2004, opened 2022-03-21
Fix minor glitch in the SVG logos

sisp:
**Description**

Fixing a minor glitch in the SVG logos that I oversaw earlier. Apologies for my negligence.

SVG errors

Those 2 path points don't belong there (also in the color logo), now they're gone.

Follow-up of #2002.


imjasonh:
lol, thanks for the fix! I don't think anybody would have noticed, but I do appreciate your attention to detail. 👍

PR #2009, opened 2022-03-22
Fix the /kaniko directory permissions in container

### LINKED ISSUE #1363: Error: error resolving dockerfile path: copying dockerfile: creating file: open /kaniko/Dockerfile: permission denied
Hi I am still running into the same issue. Is there a resolution for this @tejal29

Error: error resolving dockerfile path: copying dockerfile: creating file: open /kaniko/Dockerfile: permission denied

claudex:
Fixes #1363

**Description**

Create /kaniko directory with world permission to allow the creation of
sub directories by any user when the executor is run as non root. This
can lower the security but shouldn't have any impact in a container.

The tar unpack is the only way I found to have a directory with specific
permission as the image is created from "scratch" which doesn't have any
tool to change the permission otherwise.

**Submitter Checklist**

review them:

The container build will fail if there is an issue with the change
The container build will fail if there is an issue with the change

**Reviewer Notes**

**Release Notes**

```
- kaniko executor container can run as non root

```


Carus11:
For what its worth, I built an image with these changes and can confirm it fixed the specific issue about /kaniko permissions so that one wont see that error message when trying to use a different user when running the container.

imjasonh (approved):

hown3d:
I'm not really happy about this change since it broke so many builds:
- integration builds are broken, because files/kaniko.tar is not created in the integration test kaniko builds
- ci builds are broken, because those are run with docker buildx directly instead of the Makefile -> files/kaniko.tar is not available

PR #2067, opened 2022-04-29
fix: kaniko dir env unused

jdockerty:
**Description**

Relating to the [comment](https://github.com/GoogleContainerTools/kaniko/pull/1997#issuecomment-1113525285) made in #1997.

Previously, I made an oversight on what to be passed into the `checkKanikoDir` func, it was only the configuration option. Instead, it should have been a choice between the environment variable and the option. I have decided that the flag will take precedence, but this can easily be altered depending on any further discussion.

There was also an underlying problem of seeing `file already exists` and `invalid cross-device link` errors after the above was fixed, this seems to have been addressed by utilising the `CopyDir` func as part of the `utils` package. This enables the moving of the `KanikoDir` contents, regardless of whether it is destined for another partition or not.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

Fixes the `kaniko-dir` flag, this now takes precedence over the `KANIKO_DIR` environment variable.

### DIFF
--- a/cmd/executor/cmd/root.go
+++ b/cmd/executor/cmd/root.go
@@ -195,7 +195,7 @@ func addKanikoOptionsFlags() {
 	RootCmd.PersistentFlags().BoolVarP(&opts.SkipTLSVerifyPull, "skip-tls-verify-pull", "", false, "Pull from insecure registry ignoring TLS verify")
 	RootCmd.PersistentFlags().IntVar(&opts.PushRetry, "push-retry", 0, "Number of retries for the push operation")
 	RootCmd.PersistentFlags().IntVar(&opts.ImageFSExtractRetry, "image-fs-extract-retry", 0, "Number of retries for image FS extraction")
-	RootCmd.PersistentFlags().StringVarP(&opts.KanikoDir, "kaniko-dir", "", "/kaniko", "Path to the kaniko directory, this takes precedence over the KANIKO_DIR environment variable.")
+	RootCmd.PersistentFlags().StringVarP(&opts.KanikoDir, "kaniko-dir", "", constants.DefaultKanikoPath, "Path to the kaniko directory, this takes precedence over the KANIKO_DIR environment variable.")
 	RootCmd.PersistentFlags().StringVarP(&opts.TarPath, "tarPath", "", "", "Path to save the image in as a tarball instead of pushing")
 	RootCmd.PersistentFlags().BoolVarP(&opts.SingleSnapshot, "single-snapshot", "", false, "Take a single snapshot at the end of the build.")
 	RootCmd.PersistentFlags().BoolVarP(&opts.Reproducible, "reproducible", "", false, "Strip timestamps out of the image to make it reproducible")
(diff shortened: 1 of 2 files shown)


cmd/executor/cmd/root.go
```diff
@@ -68,7 +68,15 @@ var RootCmd = &cobra.Command{
...
 		if cmd.Use == "executor" {
 
-			if err := checkKanikoDir(config.KanikoDir); err != nil {
+			// Command line flag takes precedence over the KANIKO_DIR environment variable.
+			dir := func() string {
+				if opts.KanikoDir != constants.DefaultKanikoPath {
+					return opts.KanikoDir
+				}
+				return config.KanikoDir
+			}()
```
imjasonh, line 77:
```suggestion
			// Command line flag takes precedence over the KANIKO_DIR environment variable.
			dir := config.KanikoDir
			if opts.KanikoDir != constants.DefaultKanikoPath {
				dir = opts.KanikoDir
			}
```

imjasonh (commented):
Otherwise LGTM! Is there a test that's reasonably easy to add for this? It's okay if not.

jdockerty:
Will investigate the failing test, I believe the error on passing the `kaniko-dir` flag resulting in `exit status 1 Error: rename /kaniko /not-kaniko: file exists` is something other people were running into.

I'll dig into this a bit further when I have some spare time.

jdockerty:
I've updated the description to reflect another underlying issue which surfaced after the change made in c04538d, which should be fixed from this PR.

jdockerty:
> Is there a test that's reasonably easy to add for this? It's okay if not.

@imjasonh There's nothing that really springs to mind for a test in this case, the previous one that we created from the other PR passes and the `checkKanikoDir` is now using another function from the `util` package, which is already heavily tested.

Is there any sort of edge case which you can think of which we should cover? I may well be missing something here, but can't think of anything 😄

tejal29 (approved):

tspearconquest:
This seems not to work with non-root containers. Copying the kaniko directory won't work in this case, so we need a more comprehensive solution for the issue.

tspearconquest:
I was able to partially work around the issue by specifying the alternate directory in both the environment variable and the command line flag, and by using a dev release of the image, however I now get `Error: unlinkat //kaniko: permission denied`

I assume this is due to the `os.RemoveAll` call at https://github.com/GoogleContainerTools/kaniko/blob/323e616a679d9186ee270be0b6ca49efa2238847/cmd/executor/cmd/root.go#L259 which won't work with non-root builds:

```
$ docker run --rm -it --entrypoint "" registry.gitlab.com/armed-atk/development/devsecops/images/kaniko/executor:test /bin/sh
/workspace $ ls -ld /kaniko
drwxr-xr-x    2 kaniko   kaniko        4096 Aug  2 06:39 /kaniko
/workspace $ id
uid=65532(kaniko) gid=65532(kaniko)
/workspace $ rm -rf /kaniko
rm: can't remove '/kaniko': Permission denied
/workspace $
```

tspearconquest:
Good morning, today I updated my dockerfile to run as root, and I was able to build the image without issue using Kaniko to build it because the /kaniko path is able to be deleted properly by the above code.

I ran into other issues with the built image which appear to be unrelated to this pull request, so I'll open a new issue for that.

PR #2084, opened 2022-05-10
refactor: Make CLI argument names consistent

gabyx:
- Rename `--snapshotMode` and `--customPlatform` and `--tarPath`

**Description**

Small refactor to make CLI arguments consistent.

**Reviewer Notes**

**Release Notes**

```
- Flag `--snapshotMode` is now called `--snapshot-mode`.
- Flag `--customPlatform` is now called `--custom-platform`.
- Flag `--tarPath` is now called `--custom-path`.
```


gabyx:
> Ah did not know about this, in cobra.

gabyx:
@imjasonh: I guess now the PR should be good to go.

gabyx:
@imjasonh Lets see if the tests run through. They should... Didnt change anything...

glennakamura:
Is there a reason `--tarPath` wasn't changed `--tar-path`?

imjasonh:
> Is there a reason the `--tarPath` option wasn't changed to `--tar-path`?

Probably not.

It'd be worth aligning this as well, if anyone's interested in doing it.

gabyx:
I ll do it! :) Sorry probably a mistake.
BR

Von meinem iPhone gesendet

gabyx:
@imjasonh: Its fixed now.

Also: I formatted the markdown with prettier. Which is now 80 lines and consistent.
Also: I made the argument section a bit nicer.

gabyx:
@imjasonh: This can be merged now, thanks :)

I formatted all markdowns, to better have diffs when people add documentation.

cmd/executor/cmd/root.go
```diff
@@ -272,6 +282,21 @@ func checkContained() bool {
 	return proc.GetContainerRuntime(0, 0) != proc.RuntimeNotFound
 }
 
+// checkNoDeprecatedFlags return an error if deprecated flags are used.
```
imjasonh, line 285: This would break users of the existing flags, and wouldn't be effectively different than just removing the old flag immediately, albeit with a nicer error message.

I think we should allow the deprecated flag and just log a warning instead of failing.

gabyx:
@imjasonh Jeah totally right, we just map the values.

cmd/executor/cmd/root.go
gabyx: 👍

cmd/executor/cmd/root.go
```diff
@@ -84,8 +84,12 @@ var RootCmd = &cobra.Command{
 				return err
 			}
 
+			if err := checkNoDeprecatedFlags(); err != nil {
+				return errors.Wrap(err, "deprecated flags used")
+			}
```
imjasonh, line 89: This still looks like it will fail the command if any deprecated flags are used... 🤔

gabyx:
@imjasonh: I changed the bahv. to warning.

Made a note that we should make it deprecated in v2.0.0. Is there a notes file for v2.0.0 to list all things which might be noeted? Maybe there is also a global GO file to make such a variabel, such that we can add checks, and once we are at 2.0.0 we can set it to `true` and check all places and refactor them.

imjasonh (approved):

imjasonh:
> Made a note that we should make it deprecated in v2.0.0. Is there a notes file for v2.0.0 to list all things which might be noeted? Maybe there is also a global GO file to make such a variabel, such that we can add checks, and once we are at 2.0.0 we can set it to `true` and check all places and refactor them.

I don't think we need to wait for a full v2, so long as there's ample time to switch, and clear instructions how. I think that's satisfied here.

If you want to file an issue to remind us all to complete the migration after the next release (v1.10), that sounds fine. Otherwise I think there's enough context in the code and history to come back and remove it later.

Thanks for doing this!

acohenOT:
README still lists the deprecated parameter. Can this be fixed? 
https://github.com/GoogleContainerTools/kaniko#flag---snapshotmode

PR #2106, opened 2022-05-23
fix: getUIDandGID is able to resolve non-existing users and groups

### LINKED ISSUE #1456: COPY with --chown command should not require user or group IDs to exist
**Actual behavior**
This looks to be similar to #477, but given that was closed a year ago with unable to reproduce I thought I'd start afresh.

Building:
```
FROM alpine:latest AS helper
COPY --chown=1000:1000 scripts /scripts
```
Gives:
`error building image: error building stage: failed to execute command: getting user group from chown: user: unknown user 1000`

**Expected behavior**
This Dockerfile works with docker - I'd expect that specifying the --chown would create uid/gid 1000 without me creating them as a build step.

**To Reproduce**
Use `gcr.io/kaniko-project/executor:latest` (`v1.2.0`) to build the following image:
```
FROM alpine:latest AS helper
COPY --chown=1000:1000 scripts /scripts
```
(you'll obviously need a `scripts/` dir in the build context)

hown3d:
Signed-off-by: Höhl, Lukas <lukas.hoehl@accso.de>

Fixes #1456

**Description**
A common pattern in dockerfiles is to provide a plain uid and gid number, which doesn't neccesarily exist inside the os.

Currently, an error accurs, specifying, that the user and group doesn't exist.
This PR changes the logic to lookup the users and groups only if the chown command can't be parsed into a valid uint32.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

chown flag in Dockerfile COPY statement can specify non-existing user/group


pkg/commands/user.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // for testing
 var (
-	Lookup = util.Lookup
```
imjasonh, line 33: This seems to be unused by tests? `user_test.go` calls `util.LookupUser` instead, unless I'm misreading.

If this is the case, we can just have code in this package call `user.LookupUser` directly instead of using this var, or maybe it should be using `util.LookupUser`.

imjasonh (commented):
In this case I'd love to have an integration test Dockerfile that chowns to a non-existent user, then checks later that it's owned by that user, even though that user doesn't exist. Do you think that's something we could add to this PR?

Otherwise, this looks great, and thank you so much for working on this! 👍

pkg/commands/user.go
imjasonh: Excellent! Thanks for reading this code closer than I did. 😅

pkg/util/command_util.go
```diff
@@ -426,10 +423,32 @@ func GetUserFromUsername(userStr string, groupStr string, fallbackToUID bool) (s
...
+		group, err = user.LookupGroupId(groupStr)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return group, nil
 }
 
-func Lookup(userStr string) (*user.User, error) {
+func LookupUser(userStr string) (*user.User, error) {
```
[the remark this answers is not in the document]
hown3d: This function does indeed require the user to be present, which needs, if I think about it, to be fixed too if it's just a UID

hown3d:
> In this case I'd love to have an integration test Dockerfile that chowns to a non-existent user, then checks later that it's owned by that user, even though that user doesn't exist. Do you think that's something we could add to this PR?
> 
> Otherwise, this looks great, and thank you so much for working on this! 👍

I added a Dockerfile, is there something I need to configure inside the integration tests themself?

pkg/util/command_util.go
```diff
@@ -372,48 +373,45 @@ func GetUIDAndGIDFromString(userGroupString string, fallbackToUID bool) (uint32,
...
+		return getIdsFromUsernameAndGroup(userStr, groupStr, fallbackToUID)
 	}
 
-	gid64, err := strconv.ParseUint(gidStr, 10, 32)
+	gid, err := strconv.ParseUint(groupStr, 10, 32)
 	if err != nil {
-		return 0, 0, err
+		// check if the group is specified as a string
+		group, err := getGroupFromStr(groupStr)
+		if err != nil && !fallbackToUID {
```
hown3d, line 386: this conditional is too broad. If the error is not nil, but fallbackToUID is true, group will be nil and the program will panic: (see https://github.com/GoogleContainerTools/kaniko/runs/6558115558?check_suite_focus=true#step:6:157)

pkg/util/command_util.go
```diff
@@ -38,8 +38,7 @@ import (
 
 // for testing
 var (
-	getIdsFromUsernameAndGroup = getIdsFromUsernameAndGroupImpl
-	getGroupFromStr            = getGroup
+	GetUIDAndGID = getUIDAndGID
```
imjasonh, line 41:
```suggestion
	getUIDAndGIDFunc = getUIDAndGID
```

I'd rather not export this method if we can avoid it.

hown3d:
@imjasonh This should be ready to go now, added some more tests for other functions involving the user retrieval process.

imjasonh (approved):

imjasonh (approved):

hown3d:
I'm currently stuck regarding an issue building [this dockerfile](https://github.com/GoogleContainerTools/kaniko/pull/2106/files#diff-127f34e0d02e27a896da6bb44240bd2e65cde4e84154e360f89b6080b472f002)
All action in kaniko are executed as root, right? How does it come, that opening a file (as root), which got created in a previous stage, can be denied by the kernel? File permissions on the created file are 420. Isn’t root able to open all files, regardless of permissions and ownership?
I’m running into an *access denied* error from __os.Create__ like https://github.com/GoogleContainerTools/kaniko/runs/6782183403?check_suite_focus=true#step:6:118 all the time, which is thrown here: https://github.com/GoogleContainerTools/kaniko/blob/1395e4682f80bf4286513392220597dcca60ce1c/pkg/util/fs_util.go#L549
Maybe a second pair of eyes can spot the mistake, since I don't know why this error appears.

The error can't be reproduced on my WSL workstation, but appears in the github actions and mac running the integration tests.

msuperina:
@imjasonh Thank you for approving this MR. I am struggling to find info about which release this fix would land in, any chance you could advise ? I can see MR for https://github.com/GoogleContainerTools/kaniko/pull/2117 but this fix is not highlighted there - I am possibly mistaken though...

gabyx:
Great!

PR #2125, opened 2022-06-07
Fix#2122

BobbyNie:
Fixes #<2122>  

**Description**

**BobbyNie@139.com**

Do not call os.Chown when both DoNotChangeUID and DoNotChangeGID  to support noroot container to run some basic Dockerfile

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -57,7 +57,7 @@ COPY --from=0 /usr/local/bin/docker-credential-acr-env /kaniko/docker-credential
 COPY --from=certs /ca-certificates.crt /kaniko/ssl/certs/
 COPY --from=0 /kaniko/.docker /kaniko/.docker
 COPY files/nsswitch.conf /etc/nsswitch.conf
-#ENV HOME /root
+ENV HOME /kaniko
 #ENV USER root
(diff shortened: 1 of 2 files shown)


imjasonh:
cc @hown3d since you're also making changes to UID/GID handling in https://github.com/GoogleContainerTools/kaniko/pull/2106

BobbyNie:
> I don't think that this change will make running kaniko in rootless mode possible. Kaniko extracts the tared base image inside of the kaniko container, where folder destinations like _/bin_ are used for creating symlinks, directories etc.
> 
> By using another user than root, you can't always 100% make sure, that every folder where the image will get extracted is writable by that user. In order to achieve that, kaniko would need to have some general changes on how the base images are extracted into the kaniko container filesystem and how the file permission settings are handled.
> 
> So going rootless requires a lot of steps.
> 
> The integration tests will succeed, since the container is always run without the --user flag and no other user is specified inside the Dockerfile, so the container will be still run as root.

Thanks for review.

I'm runing kaniko in openshift cluster.
actually this attempt not work  when the Dockerfile contained  COPY or RUN  cmd.

finally I  created a service account  and give it anyuid permission. then runAsUser 0  to sove it.

PR #2128, opened 2022-06-08
fix: use refrence should after err handles

hingbong:
when I use kaniko as lib in my code, I don't know and set some options, and it doesn't handle error, just get a nil pointer panic.
should move the usage after err handling.


hingbong:
what do i need to do to fix these checks

imjasonh:
> what do i need to do to fix these checks

There was a tiny lint finding, I commited a fix on your branch (hope that's okay). Once tests pass I'll merge it.

Thanks!

hingbong:
> > what do i need to do to fix these checks
> 
> There was a tiny lint finding, I commited a fix on your branch (hope that's okay). Once tests pass I'll merge it.
> 
> Thanks!

thank you

Issue #2153, opened 2022-06-27
COPY and RUN commands have full access to the filesystem


patrickschur:
**Actual behavior**
If you build an image with Kaniko each `COPY` and `RUN` command has full acccess to the filesystem (even to files outside of the workspace). A (malicious) Dockerfile can be used to read, write or modify arbitrary files on the filesystem. E.g. you can read the credentials from `/kaniko/.docker/config.json` or modify other files to change the behaviour of subsequent builds (in case you build multiple images like we do).

**Expected behavior**
Builds should only have access to files inside `/workspace` and the extracted filesystem and not `/kaniko` or other directories part of Kaniko. Would be possible to extract the filesystem and execute each command in a temporary directory instead of using the root directory? If it's not possible can we then introduce a new flag like `--chroot` to build the image inside of a temporary directory? Of course this would require the `sys_chroot` capability.

**To Reproduce**
1. Create a Dockerfile called `Dockerfile.1`:

```Dockerfile
FROM gcr.io/kaniko-project/executor:v1.8.1-debug

# Replace the ls command with our own implementation
RUN rm -f /busybox/ls && echo -e "#!/bin/sh\necho not ls" > /busybox/ls && chmod +x /busybox/ls
````

2. Create another Dockerfile called `Dockerfile.2`:

```Dockerfile
FROM gcr.io/kaniko-project/executor:v1.8.1-debug

# I would expect this to fail with an error message like 
# "There is no such file or directory called /kaniko/.docker/config.json"
# and ls should list all files in the current directory
RUN cat /kaniko/.docker/config.json && ls
```

3. Run Kaniko via:

```sh
docker run -it --rm \
  -v $(pwd):/workspace \
  -v $HOME/.docker/config.json:/kaniko/.docker/config.json:ro \
  --entrypoint sh gcr.io/kaniko-project/executor:v1.8.1-debug
```

4. Build `Dockerfile.1`:

```sh
executor --dockerfile Dockerfile.1 -d example --no-push --cleanup
```

5. Build `Dockerfile.2`.
```sh
executor --dockerfile Dockerfile.2 -d example --no-push --cleanup
```

Output of the second build log:

```txt
INFO[0000] Retrieving image manifest gcr.io/kaniko-project/executor:v1.8.1-debug
INFO[0000] Retrieving image gcr.io/kaniko-project/executor:v1.8.1-debug from registry gcr.io
INFO[0001] Built cross stage deps: map[]
[...]

Issue #107, opened 2018-04-17
Question: why no chroot()?


anguslees:
From my read of the designdoc and quick look at the code, I think kaniko works by unpacking the image onto the real (inside the container) `/` and just assuming that the kaniko machinery can stay out of the way of the user's container data via an explicit whitelist.

I'm curious:  Why not unpack into a subdir, and chroot() there when it's time to RUN commands?  That way the "working directory" root is cleanly separated from the actual kaniko agent's root filesystem.  From what I can see, kaniko already assumes it is running as uid=0 (correct?) to preserve file permissions, run commands (apt-get) as root, etc so invoking chroot() should work just fine.

(I feel like this is such an obvious approach that I was expecting to see it mentioned in the docs, with an explanation of why it wasn't the approach taken.  I'm hoping to discover that reason with this issue/question...)

dlorenc:
Really good question! chroot can't be run inside a default container in kubernetes or docker without the --privileged flag.

I've been playing around with a version of this that uses proot/ptrace to fake a chroot, I think it can work here and help reduce privileges.

dlorenc:
Actually, it looks like my memory on this was wrong. It's the "mount --bind" that many programs use to populate a chroot before using that causes trouble inside a standard container that doesn't have the `--privileged` flag.

```shell
docker run -it alpine sh
mkdir newroot
cd newroot/
mkdir bin lib
cp -r /bin/* bin/
cp -r /lib/* lib/
chroot /newroot /bin/sh
```

We should probably do this to help separate the unpacking from the base kaniko directories. I don't think it would help much with dropping privileges though, because we'd still need to be root to create the initial rootfs.

priyawadhwa:
I tried using chroot to execute commands in a subdirectory. Since we can't use "mount --bind" I had to copy over relevant directories (`/proc`, `/dev`, `/sys`, `/etc`). Copying basically took forever and didn't really work, so I don't think we'll be able to use chroot.

PR #2169, opened 2022-07-11
feat: chroot isolation

### LINKED ISSUE #2165: Conflicting files while building images with Kanino on Kubernetes
**Actual behavior**
I'm using Tekton+EKS so we use kaniko to build images. 

We started to see an error like this:
```
{"level":"info","msg":"Running: [/bin/sh -c mkdir /root/.ssh]","time":"2022-07-06T14:22:19Z"}
mkdir: can't create directory '/root/.ssh': File exists
```

I checked the base image, and it does *not* have this folder.
Adding a `RUN ls -lah /root && ls -lah /root/.ssh` before this `RUN mkdir /root/.ssh` step, I got:
```
{"level":"info","msg":"Running: [/bin/sh -c ls -lah /root \u0026\u0026 ls -lah /root/.ssh ||:]","time":"2022-07-06T14:22:19Z"}
total 0      
drwx------    3 root     root          18 Jul  6 14:22 .
drwxr-xr-x    1 root     root         272 Jul  6 14:22 ..
drwx------    2 root     root          72 Jul  6 14:22 .ssh
total 12K    
drwx------    2 root     root          72 Jul  6 14:22 .
drwx------    3 root     root          18 Jul  6 14:22 ..
-rw-------    1 root     root         114 Jul  6 14:22 config
-rw-------    1 root     root        3.2K Jul  6 14:22 id_default-github-ssh-key
-rw-------    1 root     root         442 Jul  6 14:22 known_hosts
```

But these files are injected via ServiceAccount into kaniko's step, so it doesn't come from the base image. I tested changing to `mkdir -p /root/.ssh` and it passed, and it didn't add those files (thankfully)

**Expected behavior**
I expected some sort of isolation between what's run for building the layer. So I expected to be able to run `RUN mkdir <path>` if that file path doesn't exist on the base image.

**To Reproduce**
Although I'm using a Kubernetes, Tekton, etc. I believe you can safely test it locally by going to an empty folder and running this script:
```bash
#!/usr/bin/env bash
set -euo pipefail

mkdir -p ./foo

echo -e 'FROM alpine \nRUN mkdir /foo' > Dockerfile | tar -cf - Dockerfile | gzip -9 | docker run \
	--interactive -v $(pwd):/workspace -v $(pwd)/foo:/foo gcr.io/kaniko-project/executor:latest \
	--context tar://stdin \
	--log-format=json \
	--no-push \
	--tarPath=foo.tar \
	--destination=foo
```

hown3d:
This PR implements isolation between build commands with chroot.
Should fix #2165 and #2153.

**Description**

Chroot isolation is implemented by creating a temporary directory inside the HOME directory which is later used as the base for chroot. Before chroot, the context directory, /proc and /dev are bind mounted into the new root directory.
Isolation is wrapped between each build command execution.

Isolation can be toggled with a flag **--isolation** which resolves by default to chroot. Providing anything else will disable isolation.

TODOs before removing Draft:
- [ ] Integration tests are passing
- [ ] New unit tests for chroot

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

User facing changes:
- kaniko adds a new flag `--isolation` to specify the isolation method. The default will be **chroot**.  Removing isolation can be done by setting the flag to **none**


pkg/chroot/root.go
```diff
@@ -0,0 +1,18 @@
+package chroot
+
+import (
+	"errors"
+	"os"
+)
+
+func TmpDirInHome() (string, error) {
```
imjasonh, line 8: We may want [`os.UserHomeDir()`](https://pkg.go.dev/os#UserHomeDir) instead, which uses `$HOME` and other heuristics.

pkg/executor/build.go
```diff
@@ -348,94 +369,137 @@ func (s *stageBuilder) build() error {
...
-		isCacheCommand := func() bool {
-			switch command.(type) {
-			case commands.Cached:
-				return true
-			default:
-				return false
+func (s *stageBuilder) isolate() (exitFunc func() error, err error) {
+	switch s.opts.Isolation {
+	case "chroot":
+		{
```
imjasonh, line 395: These `{}`s are unnecessary:

```
switch s.opts.Isolation {
case "chroot":
  newRoot, err := ...
  ...
case "none":
  ...
}
```

imjasonh (commented):
This looks promising! Thanks for working on this. 👍

imjasonh:
> During my development on this PR, I ran more and more into duplications with established isolation tools like runc and crun. Is it maybe worth considering investing into using those tools that are conform with the OCI spec standard? Reimplementing those isolation techniques is imo not easy for the long run.
> 
> I'll look into the possibilities with runc's libcontainer and how it plays with kaniko.

First of all, thanks for digging into this!

I'm excited to learn what our options are if we want to retain the invariant that Kaniko doesn't require the kind of privilege that `docker build`, Buildah, etc., require, and expects to be run inside a container itself. This is an area I'm not very experienced with, so I'll mostly defer to your expertise and findings here.

hown3d:
> > During my development on this PR, I ran more and more into duplications with established isolation tools like runc and crun. Is it maybe worth considering investing into using those tools that are conform with the OCI spec standard? Reimplementing those isolation techniques is imo not easy for the long run.
> > I'll look into the possibilities with runc's libcontainer and how it plays with kaniko.
> 
> First of all, thanks for digging into this!
> 
> I'm excited to learn what our options are if we want to retain the invariant that Kaniko doesn't require the kind of privilege that `docker build`, Buildah, etc., require, and expects to be run inside a container itself. This is an area I'm not very experienced with, so I'll mostly defer to your expertise and findings here.

I took a look around for other isolation tools (libcontainer, bubblewrap, crun) but all of those tools need privileged containers, because they create cgroups and those need writable /sys dirs, thus --privileged.

I'll try to implement a sort of style that buildah did with their [chroot isolation](https://github.com/containers/buildah/tree/main/chroot).

Integration tests for isolation will be created aswell (breaking out of chroot, accessing "host" files).

natalieparellano:
I apologize for my ignorance here, but would it be possible to explain at a high level how does this PR circumvent the limitations noted here: https://github.com/GoogleContainerTools/kaniko/issues/107#issuecomment-384427886? Would a platform running containers without `--privileged` need to pass `--isolation none`?

hown3d:
Glad you asked, since this is a Draft PR I'm not 100% settled on the final implementation, but the current idea is the following:
1. Unpack Rootfs of image into isolated directory under home directory of user 
2. Create user & mount namespace to map current user to root (future rootless implementation?) and bind mount /proc /sys /dev /etc/resolv.conf and /etc/hosts (networking purposes) into the new root
3. pivot_root into newly created isolated directory where the rootfs lays and set the capabilities to the same ones that docker uses (because unshare creates a full set of capabilities)

Currently, this implementation has to set the seccomp and apparmor profiles to "unconfined" because those profiles are very strict in docker (CRI-O, Podman etc. allow mount syscalls), but it's being discussed in the moby repo: https://github.com/moby/moby/pull/42455

The final goal is to be able to run --isolation chroot without SYS_ADMIN privilege and with a slightly different seccomp/apparmor profile than default (when running with docker).

Also I think it's best to leave the default to --isolation **none** at first, since changing the default could break builds.

gabyx:
Cool stuff: Some noobie question: Why is chroot isolation needed, what does it achieve? I know it solves a security issue...

Lord-Y:
Hello guys, any news about this PR?

gewei2000:
hi guys, is there any progress about this PR?

hown3d:
Hey everyone,
I haven't had the time working on that PR, but I'm also reconsidering if this is still a good idea to implement.
The thing that bothers me:
- Chroot itself is not enough of a isolation
-> escaping from chroot is easily done when running as root, so we need to run kaniko as a non root user.
When running as non-root we would need to implement running in user namespaces to achieve root-like behavior.
This leads us into container in container issues, like having to run with a customized seccomp profile, because docker default seccomp profile blocks *setns* and *unshare* syscalls.

Kaniko is just not designed to provide more isolation inside the container environment. 
If you want to build dockerfiles inside containers in a rootless way with more isolation inside the container, but more privileges on the host, consider using the following options:
- [buildah](https://github.com/containers/buildah/blob/main/docs/buildah-build.1.md) rootless with chroot isolation (needs customized seccomp and selinux disabled).
The _containers_ team provides a container image with instructions on [quay.io](https://quay.io/repository/containers/buildah) 
- [buildkit rootless in container](https://github.com/moby/buildkit#containerizing-buildkit)
Can be configured nearly in the same way as the buildah instructions aboth. Uses [rootlesskit](https://github.com/rootless-containers/rootlesskit) to enter new namespaces inside container

I'm hereby closing this PR, since I won't continue working on that feature.

PR #2110, opened 2022-05-28
refactor: simpler local integration tests

hown3d:
**Description**

This PR tries to simplify the process of running the integration tests on a developer machine.

- remove dependency to **gsutil** and creates a package for interacting with gcs buckets.
- add **pattern matching for dockerfiles** 
  - in order to run only a portion of dockerfiles during integration testing
  - running all files can exhaust the dockerhub pull limit pretty fast
- start a **local registry** during integration-test script execution, if *LOCAL* environment variable is set. 
- dockerignore to prevent rebuilding kaniko during integration test file changes

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [x] Adds integration tests if needed.

_See [the contribution guide](../CONTRIBUTING.md) for more details._

**Reviewer Notes**

- [ ] The code flow looks good. 
- [ ] Unit tests and or integration tests added.

**Release Notes**

Simplify local integration tests

PR #2221, opened 2022-08-24
(fix):Pass full URI path to `bucket.GetNameAndFilepathFromURI`

### LINKED ISSUE #2200: Kaniko executor container image v1.9.0: "Error: error resolving source context: storage: bucket name is empty"
**Actual behavior**
We specify build jobs with the following arguments; 

```
apiVersion: batch/v1
kind: Job
metadata:
  name: {{ .Name }}
  namespace: {{ .Namespace }}
  labels:
    collectionVersion: {{ .Name }}
spec:
  backoffLimit: 1
  template:
    spec:
      restartPolicy: Never
      serviceAccountName: [sa-name-with-gcs-workload-identify]
      containers:
      - name: build
        args:
        - "--dockerfile=[file-path]/dockerfile"
        - "--context=gs://{{ .GCSBucketName }}/{{ .GCSObjectName }}"
        - "--destination={{ .ImageRegistry }}/{{ .ImageName }}:{{ .ImageTag }}"
        - "--cache=true"
```
With v.1.9.0 (the current latest build), this now causes the container to error: 

```
Error: error resolving source context: storage: bucket name is empty
Usage:
  executor [flags]
  executor [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  version     Print the version number of kaniko

apiVersion: batch/v1
Flags:
      --build-arg multi-arg type                  This flag allows you to pass in ARG values at build time. Set it repeatedly for multiple values.
      --cache                                     Use cache when building image
      --cache-copy-layers                         Caches copy layers
      --cache-dir string                          Specify a local directory to use as a cache. (default "/cache")
      --cache-repo string                         Specify a repository to use as a cache, otherwise one will be inferred from the destination provided
      --cache-run-layers                          Caches run layers (default true)
      --cache-ttl duration                        Cache timeout in hours. Defaults to two weeks. (default 336h0m0s)
      --cleanup                                   Clean the filesystem at the end
      --compressed-caching                        Compress the cached layers. Decreases build time, but increases memory usage. (default true)
  -c, --context string                            Path to the dockerfile build context. (default "/workspace/")
      --context-sub-path string                   Sub path within the given context.
      --customPlatform string                     Specify the build platform if different from the current host
  -d, --destination multi-arg type                Registry the final image should be pushed to. Set it repeatedly for multiple destinations. (default )
      --digest-file string                        Specify a file to save the digest of the built image to.
  -f, --dockerfile string                         Path to the dockerfile to be built. (default "Dockerfile")
      --force                                     Force building outside of a container
apiVersion: batch/v1
      --force-build-metadata                      Force add metadata layers to build image
      --git gitoptions                            Branch to clone if build context is a git repository (default branch=,single-branch=false,recurse-submodules=false)
  -h, --help                                      help for executor
      --ignore-path multi-arg type                Ignore these paths when taking a snapshot. Set it repeatedly for multiple paths.
      --ignore-var-run                            Ignore /var/run directory when taking image snapshot. Set it to false to preserve /var/run/ in destination image. (default true)
      --image-fs-extract-retry int                Number of retries for image FS extraction
      --image-name-tag-with-digest-file string    Specify a file to save the image name w/ image tag w/ digest of the built image to.
      --image-name-with-digest-file string        Specify a file to save the image name w/ digest of the built image to.
      --insecure                                  Push to insecure registry using plain HTTP
      --insecure-pull                             Pull from insecure registry using plain HTTP
      --insecure-registry multi-arg type          Insecure registry using plain HTTP to push and pull. Set it repeatedly for multiple registries.
      --kaniko-dir string                         Path to the kaniko directory, this takes precedence over the KANIKO_DIR environment variable. (default "/kaniko")
      --label multi-arg type                      Set metadata for an image. Set it repeatedly for multiple labels.
      --log-format string                         Log format (text, color, json) (default "color")
      --log-timestamp                             Timestamp in log output
      --no-push                                   Do not push the image to the registry
      --no-push-cache                             Do not push the cache layers to the registry
      --oci-layout-path string                    Path to save the OCI image layout of the built image.
      --push-retry int                            Number of retries for the push operation
      --registry-certificate key-value-arg type   Use the provided certificate for TLS communication with the given registry. Expected format is 'my.registry.url=/path/to/the/server/certificate'.
      --registry-mirror multi-arg type            Registry mirror to use as pull-through cache instead of docker.io. Set it repeatedly for multiple mirrors.
      --reproducible                              Strip timestamps out of the image to make it reproducible
      --single-snapshot                           Take a single snapshot at the end of the build.
      --skip-tls-verify                           Push to insecure registry ignoring TLS verify
      --skip-tls-verify-pull                      Pull from insecure registry ignoring TLS verify
      --skip-tls-verify-registry multi-arg type   Insecure registry ignoring TLS verify to push and pull. Set it repeatedly for multiple registries.
      --skip-unused-stages                        Build only used stages if defined to true. Otherwise it builds by default all stages, even the unnecessaries ones until it reaches the target stage / end of Dockerfile
      --snapshotMode string                       Change the file attributes inspected during snapshotting (default "full")
      --tarPath string                            Path to save the image in as a tarball instead of pushing
      --target string                             Set the target build stage to build
      --use-new-run                               Use the experimental run implementation for detecting changes without requiring file system snapshots.
  -v, --verbosity string                          Log level (trace, debug, info, warn, error, fatal, panic) (default "info")

Use "executor [command] --help" for more information about a command.
```

**Expected behavior**
With previous images (e.g. v.1.8.1) this manifest works appropriately and the image is built. With `latest` or `v.1.9.0` the executor command fails. 

**To Reproduce**
1. Run the kubernetes job with the specified command arguments as indicated in the issue description; in this case a gcs bucket holding the dockerfile 
2. View the output of the pod and notice the error posted above. 

**Additional Information**
Simply changing back to `gcr.io/kaniko-project/executor:v1.8.1` resolves the issue for now, but we are now locked in this version. Our workflow does use Workload Identity to access GKE clusters and GCS storage buckets, which continues to funciton prior to `v1.9.0`

 
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [ ] </li></ul>|  (not applicable, the command doesn't accept the syntax)

almg80:
Fixes #2200

**Description**

On version 1.9.0 context fetch from S3 is broken due to `url.Parse` requiring full path: https://github.com/GoogleContainerTools/kaniko/blob/90e426ba3fde4b72efbcec5f10e4f73963313228/pkg/util/bucket/bucket_util.go#L77

Currently when we do `--context=s3://my-bucket/my-path/context.tar.gz` we would pass only `my-bucket/my-path/context.tar.gz` which breaks context fetch. Tested with S3 but #2200 refers `gs://` so passing srcContext for both cases.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- kaniko adds a new flag `--registry-repo` to override registry

```

### DIFF
--- a/pkg/buildcontext/buildcontext.go
+++ b/pkg/buildcontext/buildcontext.go
@@ -50,9 +50,9 @@ func GetBuildContext(srcContext string, opts BuildOptions) (BuildContext, error)
 
 		switch prefix {
 		case constants.GCSBuildContextPrefix:
-			return &GCS{context: context}, nil
+			return &GCS{context: srcContext}, nil
 		case constants.S3BuildContextPrefix:
-			return &S3{context: context}, nil
+			return &S3{context: srcContext}, nil
 		case constants.LocalDirBuildContextPrefix:
 			return &Dir{context: context}, nil
 		case constants.GitBuildContextPrefix:


imjasonh (approved):
Just so I'm sure, this behavior seems to have been in place since it was added in https://github.com/GoogleContainerTools/kaniko/commit/65d7b0a9aa8578ceff9fce88e89c56b0cf8b72e9, 4 years ago.

Does this mean S3 build contexts have just never worked? (Not that surprising since AFAIK there's no test)

Or was there some other more recent change that affected the inputs to `GetBuildContext` or the behavior of the S3 build context that broke this?

almg80:
I believe https://github.com/GoogleContainerTools/kaniko/pull/2110 changed the way the context is being handled. No longer using `util.GetBucketAndItem` and now using `bucket.GetNameAndFilepathFromURI`. 
S3 build context is still working with 1.8.1

imjasonh:
> I believe #2110 changed the way the context is being handled. No longer using `util.GetBucketAndItem` and now using `bucket.GetNameAndFilepathFromURI`. S3 build context is still working with 1.8.1

Ah excellent, thanks for that. This definitely looks like a culprit.

liam-verta:
@imjasonh 
Can we get this out as hotfix sooner rather than later?  #2110 was supposed to be a refactor and instead it introduced this bug.  I don't see any workaround. This feature is broken until this PR is merged (or #2110 is reverted) and released.

imjasonh:
I've merged the PR, and a new commit-tagged image should be built soon (`:f9aaa9fca7bf4077778ed527c1a8a6e09e60c53c`) that includes this change.

However, I'd rather not cut releases without someone at Google around to update image refs if something goes wrong. cc @chuangw6

liam-verta:
@imjasonh Thanks! Grabbed the commit-tagged image and verified we no longer get this error.

liam-verta:
@chuangw6 Can we get a new release cut?

chuangw6:
> @chuangw6 Can we get a new release cut?

Done! https://github.com/GoogleContainerTools/kaniko/releases/tag/v1.9.1. Please see the tags in that release. Thanks!

Issue #1877, opened 2022-01-11
COPY directory and --cache-copy-layers


RoSk0:
I have a multistage Dockerfile which has `COPY` commands in multiple stages copying contents of directories. In one instance the directory contains files as a direct children. In another, it's a mix with files and many directories which also have files and directories.

When files changed in the first directory Kaniko notices it and runs COPY command, but when files change in the seconds, couple levels down from the directory root, Kaniko uses cached layer.

The questions I have are: 
1. Do I use it correctly or not ? 
2. Are my expectations reasonable and this is a bug ?

dradetsky:
Kaniko doesn't properly support `--cache-copy-layers` in multistage builds. I've made a PR to fail fast in this case https://github.com/GoogleContainerTools/kaniko/pull/2227

PR #2227, opened 2022-08-25
feat: disable cache-copy-layers in multistage builds; closes 2065

### LINKED ISSUE #2065: Kaniko should exit immediately on multi-stage builds if `--cache-copy-layers=true`
This is a Feature Request, I guess, but you don't have a template for that.

As I understand it, Kaniko does not properly support caching copy layers in multi-stage, and supporting this is not a priority. For example, #1244 has been open since May 2020, and in `v1.2.0`, Kaniko actually dropped support for caching copy layers, although it later added it back.

I'm not suggesting that supporting this should be a priority. However, given the severity of the issues that can result from this, I think it would be a _really, really good idea_ to just add a startup check where if

1. The Dockerfile we're trying to build is multi-stage
2. `opts.CacheCopyLayers` is true

we immediately exit & print "Kaniko does not support caching copy layers in multi-stage builds" or something.

In my own case, we had tons of developers coming to us saying "I pushed new code, the new code got built, but it's still running the old code. What's going on?" My boss thought that Kaniko must just be a broken, immature project & told me to migrate us to buildkit ASAP. Fortunately, I figured out what was going on, which was good because buildkit had all sorts of other issues for our setup. I'd much rather be using Kaniko with `--cache-copy-layers=false`.

I realize that for a lot of potential users, caching copy layers isn't something that even makes sense in the first place, and they'll probably just leave it off if they notice it in the first place. However, some users who know that in their particular case they have a lot of expensive COPY operations (say, they get Dockerfiles from node.js developers who don't do any kind of bundling of their backend apps) will see that there's an flag `--cache-copy-layers` and assume it just works as advertised. They won't ask themselves if it works in combination with multi-stage builds because those are just a Dockerfile best practice; mentally they will equate "works with multi-stage builds" with "works with Dockerfiles." They'll roll it out and nobody will notice any issues until weeks later, and it'll take even longer to make the connection. This is extremely poor developer UX.

Anyway, it's too late to save me & my organization from this pain, but maybe we can save others.

I'd contribute a PR, but

1. Maybe people want to discuss it
2. Maybe some core dev will just immediately fix & merge it, given how simple it is
3. I'm a terrible go developer

But I might if this is favorably received & I'm not too busy at work

dradetsky:
Fixes #2065 

**Description**

If `CacheCopyLayers` is enabled, and the user is trying to build a multi-stage Dockerfile, exit immediately since this doesn't work properly.

**Submitter Checklist**

review them:

- ~Adds integration tests if needed.~

**Reviewer Notes**

**Release Notes**

Anyone who is currently passing `--cache-copy-layers` with multistage Dockerfiles will have their builds break.

### DIFF
--- a/pkg/dockerfile/dockerfile.go
+++ b/pkg/dockerfile/dockerfile.go
@@ -58,6 +58,10 @@ func ParseStages(opts *config.KanikoOptions) ([]instructions.Stage, []instructio
 		return nil, nil, errors.Wrap(err, "parsing dockerfile")
 	}
 
+	if opts.CacheCopyLayers && len(stages) >= 2 {
+		return nil, nil, errors.New("kaniko does not support caching copy layers in multistage builds")
+	}
+
 	metaArgs, err = expandNested(metaArgs, opts.BuildArgs)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "expanding meta ARGs")
--- a/pkg/dockerfile/dockerfile_test.go
+++ b/pkg/dockerfile/dockerfile_test.go
@@ -52,7 +52,7 @@ func Test_ParseStages_NoMultistageWithCacheCopy(t *testing.T) {
 	}
 
 	opts := &config.KanikoOptions{
-		DockerfilePath: tmpfile.Name(),
+		DockerfilePath:  tmpfile.Name(),
 		CacheCopyLayers: true,
 	}
 
--- a/pkg/dockerfile/dockerfile_test.go
+++ b/pkg/dockerfile/dockerfile_test.go
@@ -29,6 +29,39 @@ import (
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
 )
 
+func Test_ParseStages_NoMultistageWithCacheCopy(t *testing.T) {
+	dockerfile := `
+	FROM scratch as first
+	COPY testfile /
+
+	FROM scratch as second
+	COPY --from=second testfile /
+	`
+	tmpfile, err := ioutil.TempFile("", "Dockerfile.test")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer os.Remove(tmpfile.Name())
+
+	if _, err := tmpfile.Write([]byte(dockerfile)); err != nil {
+		t.Fatal(err)
+	}
+	if err := tmpfile.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	opts := &config.KanikoOptions{
+		DockerfilePath: tmpfile.Name(),
+		CacheCopyLayers: true,
+	}
+
+	_, _, err = ParseStages(opts)
+	if err == nil {
+		t.Fatal("expected ParseStages to fail on MultiStage build if CacheCopyLayers=true")
+	}
+}
+
 func Test_ParseStages_ArgValueWithQuotes(t *testing.T) {
 	dockerfile := `
 	ARG IMAGE="ubuntu:16.04"


pkg/dockerfile/dockerfile.go
```diff
@@ -58,6 +58,10 @@ func ParseStages(opts *config.KanikoOptions) ([]instructions.Stage, []instructio
 		return nil, nil, errors.Wrap(err, "parsing dockerfile")
 	}
 
+	if opts.CacheCopyLayers && len(stages) >= 2 {
```
imjasonh, line 61: Is there anything we can do besides just fail here? I don't know how many builds out there depend on this, and while I agree failing is better than silently doing the wrong thing, both are worse than doing the right thing, if we can do that.

dradetsky: @imjasonh short of properly supporting caching copy layers in multistage builds, no, there's else nothing we can do here. I personally don't know enough to add this support. I believe it would be fairly complicated.

dradetsky: Actually, I guess maybe we could quietly disable CacheCopyLayers here, if that would work. We could print a message about this, but I think that for most user setups, this message wouldn't be seen. I'm dubious about this, but I guess it's an option.

The issue is that I don't know the kaniko code well. So I'm not 100% sure if any parts of the codebase have already read the value of CacheCopyLayers (or whether parts might be added which would have already read it). So this seems like a way to maybe silently reintroduce the errors.

imjasonh:
Looks like a `gofmt` needed: https://github.com/GoogleContainerTools/kaniko/runs/8026358658?check_suite_focus=true#step:4:1362

Otherwise lgtm, thanks for sending this

pkg/dockerfile/dockerfile.go
imjasonh: That could work. A slower correct build is also better IMO than a fast silently-broken one, or a fails-immediately one.

imjasonh: Do you have a test build that demonstrates the broken behavior? You could run it through kaniko with your change to disable `CacheCopyLayers` and see if it comes out as expected or not.

If we can't figure out how to fix it, we can just fail fast like you have before.

dradetsky:
@imjasonh just so there's no misunderstanding

> A slower correct build is also better IMO than a fast silently-broken one, or a fails-immediately one.

This is true, but my opinion is that changing `opts.CacheCopyLayers = false` in this case is not a "slower correct build". The amount of time to change this one value is trivial. The reason I'm dubious about it is that it is non-defensive programming. To do things this way we would have to be sure that no part of the code attempted to read the value of `opts.CacheCopyLayers` either now or as the result of any future changes. This assumption might hold, but it seems much safer to just not have to make it.

Maybe it would be possible to, at this point, restart execution of the entire process from the beginning with `CacheCopyLayers=false`, which seems more like a "slower but correct build", but this seems like too much added complexity to be worth it. Like, I'd assume that this extra complexity would result in breaking something else (it definitely would if I was the one doing it). By contrast, a few users having their builds break when they're already doing something unsupported which probably isn't working properly seems like it's not so bad.

And this is only users who have chosen not to pin their dependencies, such as their version of executor. I always pinned the executor image I used in my argo workflows since I wanted to make sure that upstream updates don't change anything unless I want them to. But if somebody doesn't pin the executor, they're saying they're okay with upstream breaking their build on upgrades. And this is a case where upstream breaking your build is a good thing, since you were doing something very bad, you just didn't understand it.

dradetsky:
I'm choosing to interpret the previous "lgtm" as implying that integration tests are not required, and crossing that off

dradetsky:
Should probably also close #1877 on merge

imjasonh (approved):

PR #2250, opened 2022-09-15
Provide `--cache-repo` as OCI image layout path

natalieparellano:
**Description**

Adds the ability to provide `--cache-repo` as an OCI image layout path to cache layers on disk.

This is useful if you want to avoid providing registry credentials to kaniko (when used together with `--cache-dir`.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

- kaniko adds the ability to provide `--cache-repo` as an OCI image layout path to cache layers on disk

### DIFF
--- a/integration/images.go
+++ b/integration/images.go
@@ -186,9 +186,10 @@ func FindDockerFiles(dir, dockerfilesPattern string) ([]string, error) {
 // keeps track of which files have been built.
 type DockerFileBuilder struct {
 	// Holds all available docker files and whether or not they've been built
-	filesBuilt           map[string]struct{}
-	DockerfilesToIgnore  map[string]struct{}
-	TestCacheDockerfiles map[string]struct{}
+	filesBuilt              map[string]struct{}
+	DockerfilesToIgnore     map[string]struct{}
+	TestCacheDockerfiles    map[string]struct{}
+	TestOCICacheDockerfiles map[string]struct{}
 }
 
 type logger func(string, ...interface{})
@@ -215,6 +216,12 @@ func NewDockerFileBuilder() *DockerFileBuilder {
 		"Dockerfile_test_cache_install": {},
 		"Dockerfile_test_cache_perm":    {},
 		"Dockerfile_test_cache_copy":    {},
+	}
+	d.TestOCICacheDockerfiles = map[string]struct{}{
+		"Dockerfile_test_cache_oci":         {},
+		"Dockerfile_test_cache_install_oci": {},
+		"Dockerfile_test_cache_perm_oci":    {},
+		"Dockerfile_test_cache_copy_oci":    {},
 	}
 	return &d
 }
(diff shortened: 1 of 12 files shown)


integration/images.go
```diff
@@ -355,42 +362,40 @@ func populateVolumeCache() error {
...
-func (d *DockerFileBuilder) buildCachedImages(config *integrationTestConfig, cacheRepo, dockerfilesPath string, version int, args []string) error {
+// buildCachedImage builds the image for testing caching via kaniko where version is the nth time this image has been built
+func (d *DockerFileBuilder) buildCachedImage(config *integrationTestConfig, cacheRepo, dockerfilesPath, dockerfile string, version int, args []string) error {
 	imageRepo, serviceAccount := config.imageRepo, config.serviceAccount
 	_, ex, _, _ := runtime.Caller(0)
 	cwd := filepath.Dir(ex)
 
 	cacheFlag := "--cache=true"
 
-	for dockerfile := range d.TestCacheDockerfiles {
```
natalieparellano, line 366: I may be missing something, but this function is called inside a `for` loop that also ranges over `TestCacheDockerfiles`.

pkg/executor/push_test.go
```diff
@@ -252,74 +252,99 @@ func TestImageNameTagDigestFile(t *testing.T) {
 	testutil.CheckErrorAndDeepEqual(t, false, err, want, got)
 }
 
-var calledExecCommand = []bool{}
```
natalieparellano, line 255: This was never getting mutated anywhere except `setCalledFalse`

pkg/executor/push_test.go
```diff
@@ -252,74 +252,99 @@ func TestImageNameTagDigestFile(t *testing.T) {
...
+				NoPush:       test.NoPush,
+				NoPushCache:  test.NoPushCache,
 			}
 			if test.ExistingConfig {
 				afero.WriteFile(fs, util.DockerConfLocation(), []byte(""), os.FileMode(0644))
 				defer fs.Remove(util.DockerConfLocation())
 			}
 			CheckPushPermissions(&opts)
-			for i, shdCall := range test.ShouldCallExecCommand {
-				if i < len(calledExecCommand) && shdCall != calledExecCommand[i] {
```
natalieparellano, line 319: Because `calledExecCommand` is never mutated, it is always 0 length, and this statement is always false; I wasn't exactly sure what this was originally supposed to be testing, but I added `checkPushPermsCallCount` to give some more coverage.

pkg/executor/build_test.go
```diff
@@ -522,6 +523,55 @@ func TestInitializeConfig(t *testing.T) {
 	}
 }
 
+func Test_newLayerCache(t *testing.T) {
+	tests := []struct {
+		description string
+		opts        *config.KanikoOptions
+		assert      func(layerCache cache.LayerCache) error
```
imjasonh, line 530: Not a huge deal, but this style effectively undoes the point of table-driven tests. I'd consider just writing three `TestNewLayerCache_Foo` methods, rather than squeezing them into a table-driven test with three different checks.

I won't block the PR on it, just something to consider.

pkg/executor/push_test.go
```diff
@@ -252,74 +252,99 @@ func TestImageNameTagDigestFile(t *testing.T) {
...
 }
 
 func TestCheckPushPermissions(t *testing.T) {
 	tests := []struct {
-		description           string
-		Destination           []string
-		ShouldCallExecCommand []bool
-		ExistingConfig        bool
+		description                     string
+		CacheRepo                       string
```
imjasonh, line 269: nit (not blocking): All of these fields should be unexported.

imjasonh (approved):
LGTM, some small nits that you can choose to ignore if you'd rather not.

Thanks for your patience, and sorry this took so long to get around to.

pkg/executor/push_test.go
natalieparellano: Updated in c55854f7

pkg/executor/build_test.go
natalieparellano: Updated in c55854f7

PR #2313, opened 2022-11-08
Add support for zstd compression

LFrobeen:
**Description**

This PR adds support for zstandard compression to kaniko. We have modified  the `go-containerregistry` package to add support for zstd compressed tarball image layers. The PR (google/go-containerregistry#1487) is currently being reviewed.

We have added two additional command line arguments `--compression` and `--compression-level` which can be used to select the compression algorithm (`gzip` or `zstd`) and specify a numeric compression level. To ensure full backwards compatibility, `gzip` will be used by default.

Please note that we had to bump the version of google/go-containerregistry to v0.13.0 and cloud.google.com/go/storage to v1.26.0 which caused a couple of other dependencies to be upgraded as well.

So far, we ran some manual tests but haven't added automated tests yet. If you have any suggestions or pointers regarding test, especially integration tests, please let us know. 

To give some more context:
We are adding zstd support to [Northflank CI/CD & Images](https://northflank.com/). We have built support for multiple build engines including Buildkit, Kaniko and CNBPs. Buildkit already supports zstd, but Kaniko does not. This PR implements zstd support for Kaniko which outperformed gzip in both performance and compression ratio in our tests.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- kaniko adds a new argument `--compression` which can be used to select the compression algorithm for layers (`zstd`, `gzip`). If omitted, `gzip` will be used by default.
- kaniko adds a new argument `--compression-level` which can be used to specify the desired compression level (0 for fastest to 9 for slowest but most compressed output)
```


pkg/config/options.go
```diff
@@ -68,6 +68,8 @@ type KanikoOptions struct {
 	ImageNameDigestFile      string
 	ImageNameTagDigestFile   string
 	OCILayoutPath            string
+	Compression              string
```
imjasonh, line 71: Could this be a `CompressionLevel` type like you have in the go-containerregistry change?

LFrobeen: If we expose the `Compression` type in the go-containerregistry package I'll just use this type. Otherwise I'll create a new one in this codebase.

PiersRBME:
Is there much activity on this project these days? I'd love to see this PR merged.

(Should the README be updated with new flags, by the way?)

LFrobeen:
Yes, I'd like to proceed with this PR. @imjasonh could you trigger the automatic checks please?

LFrobeen:
Thanks @imjasonh for triggering the tests. I'm having a hard time fixing the unit tests on my machine. It seems like the vagrant setup doesn't work on M2 cpus :/ Did anyone have this issue before?
Anyway, I'll try to debug this on another machine for now.

LFrobeen:
Ok, I've now created a proper test setup on debian. I'm pretty sure that i'm doing everything right, but the `make test` step still fails for me. 

I tried running the `make test` step in the `main` branch of this repo without any modifications and i get pretty much the same errors as in this branch. The unit tests seem to be fine, but the gofmt and linter checks fail. I was able to fix the gofmt errors but I'm having trouble with the linter error which looks like this:

```
WARN [runner] Can't run linter unused: buildssa: analysis skipped: errors in package: [/go/src/github.com/GoogleContainerTools/kaniko/pkg/config/args.go:20:2: could not import fmt (/usr/local/go/src/fmt/errors.go:8:2: could not import errors (/usr/local/go/src/errors/wrap.go:8:2: could not import internal/reflectlite (/usr/local/go/src/internal/reflectlite/swapper.go:8:2: could not import internal/goarch (-: could not load export data: cannot import "internal/goarch" (unknown bexport format version -1 ("u\x01\x00\x00\x00\x 
...
```

I tried upgrading the linter to a more recent version and while that seems to resolve the cryptic error message above, it results in new lint errors. 

Could it be that there is still something wrong with my test setup? I don't quite understand why i'm getting lint and gofmt errors on a version of the repo that has already been released.

LFrobeen:
Regarding my previous message: I realized that my changes weren't based on the v1.9.1 release, but on a more recent commit on the main branch which also suffers from the same linting issues. I thought I made my changes on top of v1.9.1 which is why i was so confused.

Earlier today I saw that there was some activity on main that seems to address the issues (thanks everyone who is working on this), so I rebased my changed onto the latest commit on main.

PiersRBME:
Glad to see this picking up again :) 

As and when there's a good time, I'd be glad to test for my use case. 

Not sure if an actual gcr.io image will be generated from this PR -- or even if merged -- but if I'd have to build the image myself, I can work that out if needed.

LFrobeen:
Sorry to be a bother with this, but could you trigger the checks again please, @imjasonh?

LFrobeen:
Thanks @imjasonh for triggering the tests. The results looked much better than the test runs before that. I fixed the issues that popped up in the previous unit test. Would you mind restarting the tests once again? Thanks a lot!

imjasonh (approved):

slonopotamus:
It it true that this PR doesn't add zstd:chunked support?

LFrobeen:
It only adds regular zstd compression. The underlying compression logic can be found here: https://github.com/google/go-containerregistry/pull/1487/files

PR #2384, opened 2023-02-01
fix(executor): make pax tar builds reproducible again

### LINKED ISSUE #2005: Reproducible builds broken in 1.8.0
**Actual behavior**
Consider the Go program main.go and its corresponding Dockerfile (both listed below). Using kaniko in version 1.7.0, two subsequent reproducible builds using the command listed below result – as expected – in two identical Docker images. In version 1.8.0, however, two subsequent builds are no longer identical.

**Expected behavior**

I expect two subsequent reproducible builds to result in identical images.

**To Reproduce**
1. Build an image by running:
```
$ docker run -v $(pwd):/src --network=host gcr.io/kaniko-project/executor:v1.8.0 --reproducible --dockerfile /src/Dockerfile --no-push --tarPath /src/image-file-main-00.tar --destination main:00 --cache=false --context dir:///src/
```
2. Build a second image by running:
```
$ docker run -v $(pwd):/src --network=host gcr.io/kaniko-project/executor:v1.8.0 --reproducible --dockerfile /src/Dockerfile --no-push --tarPath /src/image-file-main-01.tar --destination main:01 --cache=false --context dir:///src/
```
3. Import both images by running:
```
$ cat image-file-main-00.tar | docker load
$ cat image-file-main-01.tar | docker load
```
4. Compare the image IDs:
```
$ docker image ls main
REPOSITORY   TAG       IMAGE ID       CREATED   SIZE
main         00        e65d80240143   N/A       1.75MB
main         01        77fc4150ed91   N/A       1.75MB
```

The Go program is identical in both builds but the surrounding tar archive isn't. I compared the hexdump of the tar archive of both builds and noticed that there are atime and ctime fields that contain a Unix timestamp, which is the reason why the builds differ. Could this regression have been caused by ee95be1e?

**Additional Information**
 - Dockerfile 
```
FROM golang:1.18 as builder
WORKDIR /src
COPY main.go ./
RUN CGO_ENABLED=0 GO111MODULE=off go build -trimpath -o main
FROM scratch as artifact
COPY --from=builder /src/main /bin/
CMD [ "/" ]
```
 - Build Context
```
package main                      
 
import "fmt"

func main() {
	fmt.Println("Hello!")
}
```
 
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [X] </li></ul>|

BronzeDeer:
Fixes #2005 

In v1.8.0 (commit 7410007) kaniko switched to using the pax tar header format for compressing image layers, since this format allows for greater precision in recording timestamps, however this inadvertendly broke the "--reproducible" functionality, due to an bug in the underlying go-containerregistry dependency which did not set the additional timestamps in the pax header when canonicalizing image layers. This oversight has since been fixed in the dependency.

This commit bumps the google/go-containerregistry dependency to the first commit which has fixed the bug
(v0.13.1-0.20230201183932-824efc7772b0). It also bumps the version of cloud.google.com/go/storage to v1.29.0 to be compatible with the higher transitive dependency.

**Submitter Checklist**

review them:

(Since this is a fix in a dependency, no new integration or unit tests were added, do we need a regression or acceptance test for the `--reproducible` functionality?)

**Reviewer Notes**

**Release Notes**

```
- kaniko reproducible builds work correctly again

```


imjasonh (approved):

imjasonh:
Can you also update these lines to `go-version: 1.20`?

https://github.com/GoogleContainerTools/kaniko/blob/fe2413e6e3c8caf943d50cf1d233a561943df1d6/.github/workflows/unit-tests.yaml#L18
https://github.com/GoogleContainerTools/kaniko/blob/fe2413e6e3c8caf943d50cf1d233a561943df1d6/.github/workflows/integration-tests.yaml#L30

BronzeDeer:
I created the requested changes in the workflows and rebased both commits on top of main

imjasonh:
This error now seems to be https://github.com/actions/setup-go/issues/328

.github/workflows/integration-tests.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
     steps:
     - uses: actions/setup-go@b22fbbc2921299758641fab08929b4ac52b32923 # v3
       with:
-        go-version: 1.17
+        go-version: 1.20
```
imjasonh, line 30:
```suggestion
        go-version: '1.20'
```

.github/workflows/unit-tests.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
     steps:
     - uses: actions/setup-go@b22fbbc2921299758641fab08929b4ac52b32923 # v3
       with:
-        go-version: 1.17
+        go-version: 1.20
```
imjasonh, line 18:
```suggestion
        go-version: '1.20'
```

imjasonh:
Sorry about these failures. https://github.com/GoogleContainerTools/kaniko/pull/2386 does the Go update and fixes the gofmt findings.

I'm not sure why the integration tests are failing.

BronzeDeer:
> Sorry about these failures. #2386 does the Go update and fixes the gofmt findings.
> 
> I'm not sure why the integration tests are failing.

Sorry that I wasn't around, had a busy time at work, I can now also look back into this. In the mean time, would you like me to rebase this PR on top of #2386 ?

BronzeDeer:
As a test, I've bumped the version of golangci-linter to the latest version, it can now handle the code for go 1.20, but it starts complaining about a lot of new unresolved linting issues (The jump was 24 minor versions). Maybe we should first spin out the fixes to ci into a separate PR including fixing the linting issues?

imjasonh:
> As a test, I've bumped the version of golangci-linter to the latest version, it can now handle the code for go 1.20, but it starts complaining about a lot of new unresolved linting issues (The jump was 24 minor versions, so new errors are to be expected). Maybe we should first spin out the fixes to the ci into a separate PR including fixing the linting issues?

Unfortunately golangci-lint adds and drops and changes lint requirements pretty aggressively sometimes, even within minor releases.

I'd say if there are any you can fix easily then go ahead, otherwise feel free to disable the noisy linters so this can merge. Nobody has the bandwidth to fix all of kaniko's many lint errors, and keep them up to date as golangci-lint finds more. Best to get the actual bugfix in.

https://github.com/GoogleContainerTools/kaniko/pull/2386 indicates there might also be minikube errors lurking past the Go version bump.

Sorry for all these speed bumps.

BronzeDeer:
I've created a separate PR in #2400 that bumps the linter version and fixes all linter errors and then rebased this PR on top of it. I also squashed the all  test workflow go version bumps into a single commit

BronzeDeer:
@imjasonh Sorry to be here only so sporadicly, but could you unblock the workflows on the 2 PRs so that I can get some feedback on the integration tests, even if there is not enough time to review yet?

BronzeDeer:
I've rebased on top of the new version of the other PR, now the boilerplate test should also pass, the failing setup for minikube is still left blocking the integration tests, I'll see if I can find the problem in there later

BronzeDeer:
Update to the CI situation: I've played around with the setup script a bit here https://github.com/BronzeDeer/kaniko/pull/2

I've managed to restore minikube installation to a working state taking into account all the CNI and CRI changes that happened in k8s 1.24, currently I'm stuck on why the hostPort of the registry-proxy doesn't work. I'm at the point where I'm thinking about just setting up a more lightweight alternative using k3os or similar. 

@imjasonh Is there a list of binaries/capabilities that the integration runner VM definitely needs? I might be  able to explore simpler options

imjasonh:
> I've managed to restore minikube installation to a working state taking into account all the CNI and CRI changes that happened in k8s 1.24, currently I'm stuck on why the hostPort of the registry-proxy doesn't work. I'm at the point where I'm thinking about just setting up a more lightweight alternative using k3os or similar.

Simplifying the integration test setup however you see fit would be very welcome, if you have the time and energy to undertake it. I think KinD could also be a good solution. AFAIK there's no reason to test k8s specifically at all. A test that uses [`run_in_docker.sh`](https://github.com/GoogleContainerTools/kaniko/blob/main/run_in_docker.sh) to build some Dockerfiles would likely be just as good as a test, and require much less infrastructure.

> @imjasonh Is there a list of binaries/capabilities that the integration runner VM definitely needs? I might be able to explore simpler options

Not that I'm aware of.

BronzeDeer:
> > I've managed to restore minikube installation to a working state taking into account all the CNI and CRI changes that happened in k8s 1.24, currently I'm stuck on why the hostPort of the registry-proxy doesn't work. I'm at the point where I'm thinking about just setting up a more lightweight alternative using k3os or similar.
> 
> Simplifying the integration test setup however you see fit would be very welcome, if you have the time and energy to undertake it. I think KinD could also be a good solution. AFAIK there's no reason to test k8s specifically at all. A test that uses [`run_in_docker.sh`](https://github.com/GoogleContainerTools/kaniko/blob/main/run_in_docker.sh) to build some Dockerfiles would likely be just as good as a test, and require much less infrastructure.
> 
> > @imjasonh Is there a list of binaries/capabilities that the integration runner VM definitely needs? I might be able to explore simpler options
> 
> Not that I'm aware of.

Just a quick update, since it's been a few days again. I've reworked most of the integration setup and got it working quicker and almost all green again, just struggling on one weird error around OCI support, will cut a separate PR once its ready. For Details: https://github.com/BronzeDeer/kaniko/pull/2

imjasonh:
You're a champion. 🤯

BronzeDeer:
I opened the integration test refactor PR (#2425)  and rebased this PR on top of it. Currently the fix PR is still waiting on another fix in [container-diff ](https://github.com/GoogleContainerTools/container-diff/issues/389) and once that is integrated into the integration fix PR, we should finally be able to validate and merge this as well

isker:
Thank you for your quite persistent efforts to get this fixed.  I can finally update off 1.7 🌞.

pstoellberger:
Thank you! Very much appreciated!

imjasonh:
> @imjasonh What's the policy regarding when to make a new tag, ie 1.9.2?

For that we beg someone at Google to cut a release. cc @chuangw6 🙏🥹🙏

chuangw6:
> > @imjasonh What's the policy regarding when to make a new tag, ie 1.9.2?
> 
> For that we beg someone at Google to cut a release. cc @chuangw6 🙏🥹🙏

Done! https://github.com/GoogleContainerTools/kaniko/releases/tag/v1.9.2. 🎉

PR #2437, opened 2023-03-21
bump cosign version used to sign images

imjasonh:
It seems that images can't be signed due to using a too-old cosign version. This fixes that.


chuangw6 (approved):
Thanks @imjasonh for the fix!

Issue #1960, opened 2022-03-02
--reproducible flag massively increases build time


roussidis:
I am building one large docker image and I am experiencing a somewhat weird behavior that I would like to address.

So we are talking about a large-ish image (~2.5GB) and when I built it with cache it is being built at ~1 min. 
So far so good, but the problem that I am facing is that this cached image is always creating a new artifact with a new sha256. 

I found out that for that issue the `--reproducible` flag exists and it does exactly that. When I built the image with the `--reproducible` flag it doesn't create a new artifact with a new sha256 but the build time increases from ~1 min to ~7 mins. That's a huge overhead imho and I would like to figure that out.
I got logs (with trace verbosity) when the `--reproducible` flag was enabled and when it was not. The main difference that was found in the logs:

-  `--reproducible` enabled 

```
time="2022-03-02T12:36:05Z" level=debug msg="mapping stage idx 0 to digest sha256: ..."
time="2022-03-02T12:36:05Z" level=debug msg="mapping digest sha256 to cachekey ..."
time="2022-03-02T12:40:47Z" level=info msg="Pushing image to ..."
```

- `--reproducible` disabled

```
 time="2022-03-02T12:29:12Z" level=debug msg="mapping stage idx 0 to digest sha256: ..."
 time="2022-03-02T12:29:12Z" level=debug msg="mapping digest sha256 to cachekey ..."
 time="2022-03-02T12:29:12Z" level=info msg="Pushing image to ..."
```

If you watch closely in the timestamps when `--reproducible` is enabled the stage `mapping digest...` is ~5 mins longer compared when the `-reproducible` is disabled.

 Why is this behavior and why does it add so much overhead in the build process?
 Is there any other way to use cache to build an image but not create a new artifact like docker does. `--reproducible` completely strips down the timestamps which provides a not useful output when running `docker images`.
 
 e.g. In the below code block when using `--reproducible` the CREATED output is **N/A** which is not the desired behavior. 
 ```
 REPOSITORY                           TAG               IMAGE ID        CREATED              SIZE
 random-image                       latest             abcdefghijk         N/A                 2.68GB
 ```

tejal29:
For your second question, see [here](https://github.com/GoogleContainerTools/kaniko#--reproducible) Using the flag, all timestamps are stripped off and no history is available.

Will have to dig deeper into performance issue. Can you clarify your use case a little bit more?
[...]

Issue #862, opened 2019-11-15
Using --reproducible loads entire image into memory


nathan-c:
**Actual behavior**
When running with --reproducible=true the entire image is loaded into memory. 

**Expected behavior**
Memory profile should remain stable regardless of size of image being built.

**To Reproduce**
1. Create following Dockerfile
    ```
    FROM ubuntu:latest
    RUN dd if=/dev/zero of=output.dat  bs=1M  count=1000
    ```
1. Execute kaniko with the following parameters
    ``` 
    --context=/some-empty-directory
    --dockerfile=/some-empty-directory/Dockerfile
    --destination=<any destination>
    --cache=true
    --reproducible=true
    ```
2. Watch as memory consumption climbs dramatically near the end of the build.

**Additional Information**
 - Having done some investigation, this is down to go-containerregistry not kaniko code so maybe this issue belongs there? The offending code is: https://github.com/google/go-containerregistry/blob/master/pkg/v1/mutate/mutate.go#L300
 - Image used: gcr.io/kaniko-project/executor@sha256:fe07c91e9342a097a7ee7bf90d0d8b49b285a725315758e92de03e9f5debbb5c
 
 **Triage Notes for the Maintainers**
 

 
 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li>- [ ] </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li>- [x] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [ ] </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [ ] </li></ul>|

nathan-c:
I should probably mention that this means we cannot use --reproducible for our builds because the containers running the build are being killed by k8s on our dev cluster and increasing the node RAM capacity is not a cost effective solution

macrotex:
We are having this issue as well. We need to use reproducible builds so that our daily builds do not get added to our container registry if there is no "real" change to the docker image.

Phylu:
> @Phylu Will this also be closed in the 1.7.0 release as of #1722?

I am not sure. Did not try this flag with my changes. If you are lucky, it could be that way.

bh-tt:
@zx96 do you plan to create a PR for the linked commit? It would save many people a lot of frustration if our image builds would stop crashing at the end of a build due to being killed.

[...]

PR #2477, opened 2023-04-22
Add `--zero-file-timestamps` flag

zx96:
**Description**

This MR adds a new flag (`--zero-file-timestamps`) to zero timestamps in layer tarballs without making a fully reproducible image. This flag provides a workaround (but not a complete solution) for #862 and #1960.

My use case for this is maintaining a large image with build tooling. I have a multi-stage Dockerfile that generates an image containing several toolchains for cross-compilation, with each toolchain being prepared in a separate stage before being `COPY`'d into the final image. This is a very large image, and while it's incredibly convenient for development, making a change as simple as adding one new tool tends to invalidate caches and force the devs to download another 10+ GB image.

If timestamps were removed from each layer, these images would be mostly unchanged with each minor update, greatly reducing disk space needed for keeping old versions around and time spent downloading updated images.

I wanted to use Kaniko's `--reproducible` flag to help with this, but ran into issues with memory consumption (#862) and build time (#1960). Additionally, I didn't really care about reproducibility - I mainly cared about the layers having identical contents so Docker could skip pulling and storing redundant layers from a registry.

This solution works around these problems by stripping out timestamps as the layer tarballs are built. It removes the need for a separate postprocessing step, and preserves image metadata so we can still see when the image itself was built.

An alternative solution would be to use `mutate.Time` much like Kaniko currently uses `mutate.Canonical` to implement `--reproducible`, but that would not be a satisfactory solution for me until [issue 1168](https://github.com/google/go-containerregistry/issues/1168) is addressed by go-containerregistry. Given my lack of Go experience, I don't feel comfortable tackling that myself, and this seems like a simple and useful workaround in the meantime.

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you review them:

**Reviewer Notes**

**Release Notes**

```
- Add `--zero-file-timestamps` flag to make layer contents reproducible without stripping metadata
```


peter-volkov:
Dear maintainers. Take a look at this PR please

zx96:
It's been a few months and I still would really like this feature; is there anything I can do to help the chances of this getting merged?

This change solves an immediate problem that I (and several others) have, and it seems like a useful feature to me in general (I think that having reproducible layer contents but preserving the image metadata itself would be valuable).

derari:
>  I didn't really care about reproducibility - I mainly cared about the layers having identical contents

Isn't this exactly what "reproducible" means?

But I, too, would prefer this implementation over the existing one

jkalez:
@zx96, has this been tested?  I'm attempting to use your branch and am still getting cache misses in use-cases where `--reproducible` gets cache hits.  Granted the time overhead of `--reproducible` completely negates any cache performance benefits, so a fix like you're proposing is absolutely necessary.

zx96:
@derari In my head, "reproducible" generally means you get the exact same thing out and can verify it with the hashes; this flag doesn't give you the exact same image, just identical layer contents (which is good enough for Docker to avoid storing the same thing multiple times).

@jkalez I did test it back before I initially opened the PR, but my testing was mostly just ensuring I got the same layer hashes out reliably and that they matched the layer hashes I got with `--reproducible`. Could you tell me more about how you're using the cache (what command are you running Kaniko with) so I can look into it?

jkalez:
I'm using commit `bd1dd69b59f164e3702493b66c8edb7ae7059614` from your repository, and built a kaniko image from that commit via `docker build -t kaniko_zero_file_timestamps:debug -f deploy/Dockerfile_debug .`

My kaniko arguments look as follows:
```bash
      /kaniko/executor
      --cache=true
      --cache-copy-layers=true
      --cache-run-layers=true
      --cache-ttl=24h
      --cache-repo="$CI_REGISTRY_IMAGE/container-cache"
      --compressed-caching=false
      --zero-file-timestamps
      --snapshot-mode=redo
      --context="$CI_PROJECT_DIR/images/$IMAGE_NAME"
      --build-arg BASE_IMAGE
      --destination "$CI_REGISTRY_IMAGE/$IMAGE_NAME:$CI_COMMIT_REF_SLUG"
      --registry-certificate "$CI_REGISTRY=$CI_SERVER_TLS_CA_FILE"
```
All the environment variables are appropriately populated.  For this test, I'm just using a local container registry like this: https://docs.docker.com/registry/deploying/

The context directory "$CI_PROJECT_DIR/images/$IMAGE_NAME" contains a `setup.sh` script and a very small Dockerfile
```dockerfile
ARG BASE_IMAGE
FROM $BASE_IMAGE

RUN apt update
RUN apt install -y xvfb libxcb1-dev
COPY script.sh .
ENTRYPOINT ["./script.sh"]
```

If I run kaniko like this twice with a cold cache, I expect the following:
- 1st run: build executes in its entirety and pushes layers to the cache
- 2nd run: build pulls from cached layers and runs significantly faster

However, the 2nd run never actually pulls anything from the cache.  I get logs like the following:
```
INFO[0000] Checking for cached layer <$CI_REGISTRY_IMAGE>/container-cache:fd283800bb67e1fd789bf395e0297aacaee9c8682640ef3c2fdb37175a77745b...
INFO[0000] No cached layer found for cmd RUN apt update
INFO[0000] Unpacking rootfs as cmd RUN apt update requires it.
```

Note in the 1st run, I got the following:
```
INFO[0000] Checking for cached layer <$CI_REGISTRY_IMAGE>/container-cache:bcd9d2b1730856f75ee1324198f7176c6d67ea896ba8a0f9e7a567f16c8f14ee...
INFO[0000] No cached layer found for cmd RUN apt update
INFO[0000] Unpacking rootfs as cmd RUN apt update requires it.
...
INFO[0030] Pushing layer <$CI_REGISTRY_IMAGE>/container-cache:bcd9d2b1730856f75ee1324198f7176c6d67ea896ba8a0f9e7a567f16c8f14ee to cache now
INFO[0030] Pushing image to <$CI_REGISTRY_IMAGE>container-cache:bcd9d2b1730856f75ee1324198f7176c6d67ea896ba8a0f9e7a567f16c8f14ee
```

Obviously those hashes do not match.  Note if I run the exact same test with `--reproducible` instead of `--zero-file-timestamps`, the second run pulls from the cache:
```
INFO[0000] Checking for cached layer <$CI_REGISTRY_IMAGE>/container-cache:352dbd5bf4eb089c4c765d74b1d32b86dc2c2299f2cadf74348fee35beb2b665...
INFO[0000] Using caching version of cmd: RUN apt update
```
And this is the first run when run with `--reproducible`:
```
INFO[0031] Taking snapshot of full filesystem...
INFO[0032] Pushing layer <$CI_REGISTRY_IMAGE>/container-cache:352dbd5bf4eb089c4c765d74b1d32b86dc2c2299f2cadf74348fee35beb2b665 to cache now
INFO[0032] Pushing image to <$CI_REGISTRY_IMAGE>/container-cache:352dbd5bf4eb089c4c765d74b1d32b86dc2c2299f2cadf74348fee35beb2b665
```

Note here that the hashes match.

Let me know if there's anything else I can provide to you to help you repro this!

jkalez:
@zx96 any luck reproducing what I saw in your branch?

zx96:
@jkalez I've been busy and haven't had a chance to look into it yet... :sweat_smile: 

I might get a chance to look at this weekend; I'll update here if I figure something out.

bh-tt:
@zw96 could this perhaps be merged with a note that setting this flag can have some issues with layer caching? That would at least enable the feature, which likely provides a larger benefit to speed than the cache does in the first place.

philippe-granet:
[SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/) is a convention for pinning timestamps to a specific value.
Buildkit use it:
https://www.docker.com/blog/highlights-buildkit-v0-11-release/
https://github.com/moby/buildkit/blob/master/docs/build-repro.md#source_date_epoch

Could Kaniko use this convention?

bh-tt:
Hey, I'd still like this feature very to be in kaniko very much. We have builds taking 3 seconds to initialize, 5 seconds to pull the image, 10 seconds to unpack the rootfs, 2 seconds to run the Dockerfile commands, and then 38 (!) seconds before it starts to push the image, which takes about 10 seconds again. This means a much faster implementation of --reproducible would halve our build time. It seems @zx96 has abandoned this PR, perhaps someone can take over?

PR #2494, opened 2023-05-07
Default User and Group to 0:0 for ADD and COPY

### LINKED ISSUE #1921: COPY does not default UID to 0
**Actual behavior**
Running kaniko in my local docker daemon, files copied with `COPY` still have my UID in the built container. This breaks reproducibility when different users build the same image because the file's UID is included in the layer hash.

**Expected behavior**
Files copied with `COPY` and no `--chown` arg should be defaulted to UID 0 as per the [Dockerfile documentation](https://docs.docker.com/engine/reference/builder/#copy)
> All new files and directories are created with a UID and GID of 0, unless the optional --chown flag specifies...

**To Reproduce**
```sh
#!/bin/sh

echo "Running as $USER (uid=$(id -u))"

cd $(mktemp -d)

cat >run.sh <<'EOF'
#!/bin/sh
stat $0
EOF

chmod 0755 run.sh

cat >Dockerfile <<EOF
FROM alpine:latest

COPY run.sh /app/

ENTRYPOINT ["/app/run.sh"]
EOF

echo "Building with docker"
docker build -t uid-demo:docker .

echo "Building with kaniko"
docker run --rm -v $(pwd):/workspace gcr.io/kaniko-project/executor:v1.7.0 \
    --destination uid-demo:kaniko \
    --no-push \
    --tarPath image.tar

docker load < image.tar

echo "Running docker version"
docker run --rm -it uid-demo:docker
echo "Running kaniko version"
docker run --rm -it uid-demo:kaniko
```

**Additional Information**
 - Dockerfile 
   see above
 - Build Context
   see above 
 gcr.io/kaniko-project/executor:v1.7.0
         "Id": "sha256:4f9b8daa25ddc38e7f91eba9a7b989a555c623450c6f39b91bb51c66f1ec994b",
 
**Workaround**
Explicitly add `--chown` to every `COPY`/`ADD` in each Dockerfile.

KerchumA222:
Default to UID 0 and GID 0 as per the [Dockerfile documentation](https://docs.docker.com/engine/reference/builder/#copy)

Fixes #1921

**Description**
Update to match Dockerfile specifications when using ADD or COPY
Previous functionality was to preserve the user and group from the source, which may not exist in the container.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

ADD and COPY now default to UID 0 and GID 0 instead of maintaining the source ownership

### DIFF
--- a/pkg/util/command_util.go
+++ b/pkg/util/command_util.go
@@ -347,7 +347,7 @@ Loop:
 
 func GetUserGroup(chownStr string, env []string) (int64, int64, error) {
 	if chownStr == "" {
-		return DoNotChangeUID, DoNotChangeGID, nil
+		return DefaultUID, DefaultGID, nil
 	}
 
 	chown, err := ResolveEnvironmentReplacement(chownStr, env, false)
--- a/pkg/util/command_util_test.go
+++ b/pkg/util/command_util_test.go
@@ -561,8 +561,8 @@ func TestGetUserGroup(t *testing.T) {
 			mockIDGetter: func(string, string) (uint32, uint32, error) {
 				return 0, 0, fmt.Errorf("should not be called")
 			},
-			expectedU: -1,
-			expectedG: -1,
+			expectedU: 0,
+			expectedG: 0,
 		},
 	}
 	for _, tc := range tests {
--- a/pkg/util/fs_util.go
+++ b/pkg/util/fs_util.go
@@ -45,6 +45,8 @@ import (
 const (
 	DoNotChangeUID = -1
 	DoNotChangeGID = -1
+	DefaultUID     = 0
+	DefaultGID     = 0
 )
 
 const (


pkg/util/command_util.go
```diff
@@ -352,7 +352,7 @@ Loop:
 
 func GetUserGroup(chownStr string, env []string) (int64, int64, error) {
 	if chownStr == "" {
-		return DoNotChangeUID, DoNotChangeGID, nil
+		return 0, 0, nil
```
JeromeJu, line 355: Hi @KerchumA222 this default change looks good.
For the nit, would you mind also following the previous convention for naming it as DefaultUID?
Thanks!

JeromeJu (approved):

pkg/util/command_util.go
KerchumA222: @JeromeJu I updated the code. Is this what you meant?

JeromeJu (approved):

KerchumA222:
@JeromeJu this seems to be stalled waiting for approval for workflows. Is that something you can help with?

pkg/util/command_util.go
JeromeJu: Thanks @KerchumA222

JeromeJu:
> @JeromeJu this seems to be stalled waiting for approval for workflows. Is that something you can help with?

Thanks @KerchumA222 for the changes, I have re-runned the integration tests but the error seems to have persisted. Would you mind taking a look into it?

KerchumA222:
The issue is that the file ownership can't be changed to root when running as a non-root user. The tests are run as a regular user and therefore fail.
I need to think through this problem a little before proceeding.
Would a virtual filesystem be acceptable (so that we can create and manipulate files owned by user 0 without needing root)?
Do you have any other ideas on how to solve this? I don't want to sacrifice being able to build images without root access (without needing to `--chown` to a regular user).

KerchumA222:
Without running inside a container, I don't know of any other way to make files owned by root that can be interacted with by the RUN command. I thought of using a virtual filesystem to track ownership and properly encode that when building the tar files, but the complexity of that solution is pretty high and leaves some open questions (how to track files created or modified with a RUN command for example).

If the tests are run with `sudo`, they should pass. How about running the tests with `sudo`, and adding another basic test suite that runs without `sudo` to prove root isn't required unless you need root-owned files?

JeromeJu:
Thanks to the offline discussion with @aaron-prindle .
The pointer led to the design of Kaniko with UID for historical reasons. From Kaniko's previous design, we do not have user root id access in containers.
So the changes in this PR is going to be a breaking change for <= v1.20 users since all the copy files will be root after this.

aaron-prindle:
Closing this for reasons @JeromeJu mentioned.  Feel free to continue the discussion around this in #1921, currently IIUC this would break users

Issue #2276, opened 2022-10-04
CVEs on 1.9.1


tooptoop4:
/kaniko/docker-credential-gcr has github.com/docker/docker (v20.10.8) affected by CVE-2021-39293,CVE-2021-41089,CVE-2021-41091,CVE-2021-41103,CVE-2022-1705,CVE-2022-1962,CVE-2022-24769,CVE-2022-27664,CVE-2022-28131,CVE-2022-29162,CVE-2022-30630,CVE-2022-30631,CVE-2022-30632,CVE-2022-30633,CVE-2022-30635,CVE-2022-32148,CVE-2022-32189,CVE-2022-32190,CVE-2022-36109

/kaniko/docker-credential-gcr has github.com/docker/cli (v20.10.8) affected by CVE-2021-39293, CVE-2021-41092

/kaniko/docker-credential-acr-env has golang.org/x/crypto (0.0.0-20201002170205-7f63de1d35b0) affected by CVE-2020-9283, CVE-2022-27191

/kaniko/docker-credential-acr-env has github.com/dgrijalva/jwt-go (3.2.0) affected by CVE-2020-26160

liam-verta:
This kind of blanket reporting of CVEs is not useful - it makes it very hard to track progress and know when the issues are finally closed.

It looks like most of these are closed. 

In v1.9.1, the first three items are addressed:
* Fixed: github.com/docker/docker has been updated to v20.10.14 - [go.mod#L20](https://github.com/GoogleContainerTools/kaniko/blob/f9aaa9fca7bf4077778ed527c1a8a6e09e60c53c/go.mod#L20)
* Fixed: github.com/docker/docker has been updated to v20.10.12 - [go.mod#L88](https://github.com/GoogleContainerTools/kaniko/blob/f9aaa9fca7bf4077778ed527c1a8a6e09e60c53c/go.mod#L88)
* Fixed: golang.org/x/crypto has been updated to v0.0.0-20220112180741-5e0467b6c7ce - [go.mod#L141](https://github.com/GoogleContainerTools/kaniko/blob/f9aaa9fca7bf4077778ed527c1a8a6e09e60c53c/go.mod#L141)

Usually, `go list -m all`will get a list of the actual version of each dependency that is used.  Kaniko "vendors" all their dependencies, so we have to use `go list -m all -mod=mod` to get that list.  It confirms that the above items are fixed. 

[...]

PR #2531, opened 2023-05-26
chore(deps): update docker-credential-* binaries in kaniko images

### LINKED ISSUE #2251: docker-credential-gcr produces Username incompatible with GCR when used by other libraries
**Actual behavior**
 `echo https://gcr.io | docker-credential-gcr get` produces output like this:
```json
{"ServerURL":"https://gcr.io","Username":"_dcgcr__token","Secret":"<redacted_oauth_token>"}
```

When using this helper with other libraries like Jib, they fail to successfully authenticate against GCR because GCR would not accept the username in the format `_dcgcr__token` (note the **2** `_`s between `dcgcr` and `token`).

But when `docker-credential-gcr` is build from the latest source, it produces `Username` in the format `_dcgcr_2_0_4_token`,
which is accepted by GCR when using basic auth.

**Expected behavior**
`docker-credential-gcr get` should produce `Username` that has version between `dcgcr` and `token`, which works with GCR.

**To Reproduce**
Please see the commands above.

**Additional Information**
 - Dockerfile 
   N/A
 - Build Context
   N/A
   `gcr.io/kaniko-project/executor@sha256:a3f8c85c6a0fa490b9a619bdb503c8cb98fb2fd17f96d0e4356a0bd65c1e5056` (`:debug`)
   `gcr.io/kaniko-project/executor@sha256:1f982af0b54be748221d9a35dcfa608660ab3d51229aa56bde5416f75aff7561` (`:latest`)
 
I believe this issue is due to the [version information](https://github.com/GoogleCloudPlatform/docker-credential-gcr/blob/4cdd60d0f2d8a69bc70933f4d7718f9c4e956ff8/config/const.go#L119) somehow not being bundled with the build used (even though it shows up as `v1.5.1-0.20210909233216-4cdd60d0f2d8` on `version` command).

I would suggest to update the binary to the latest version and make sure it outputs `Username` in the right format.

aaron-prindle:
fixes #2251 
Related: https://github.com/GoogleContainerTools/kaniko/issues/2276

Updates:
- docker-credential-gc
- docker-credential-ecr-login
NOTE: docker-credential-acr-env has no newer updates

This reduces the set of CVEs present in the kaniko image but unfortunately these binaries still have CVEs present even in their most up to date releases

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -23,10 +23,10 @@ ENV CGO_ENABLED=0
 ENV GOBIN=/usr/local/bin
 
 # Get GCR credential helper
-RUN go install github.com/GoogleCloudPlatform/docker-credential-gcr@4cdd60d0f2d8a69bc70933f4d7718f9c4e956ff8
+RUN go install github.com/GoogleCloudPlatform/docker-credential-gcr@62afb2723512fd6572c6300024e0c94f734b0ae3 #v2.1.8
 
 # Get Amazon ECR credential helper
-RUN go install github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login@69c85dc22db6511932bbf119e1a0cc5c90c69a7f # v0.6.0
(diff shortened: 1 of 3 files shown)


chuangw6 (approved):
Thank you @aaron-prindle !

PR #2553, opened 2023-06-09
Remove `github.com/docker/docker/builder` as dependency

BobDu:
Remove `github.com/docker/docker/builder` as dependency.

We only used several helper functions in buildargs.go, but introduced a large number of dependency libraries due to the dependency of the docker builder package. 
A very small amount of duplicate code (less than 100 lines) can eliminate many dependencies, and the binary size has decreased by ~20%.

**Description**

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Remove `github.com/docker/docker/builder` as dependency

```


aaron-prindle:
Thank you for the PR here @BobDu,  appreciate your work here.  For the time being I think it makes sense to keep the dependency as is to allow for pulling in any updates values such as `allowedBuildArgs`, `allowedMetaArgs`, etc. directly from the `github.com/docker/docker/builder` package vs having to manually identify and fix any discrepency.  kaniko builder args should be as close to 1:1 as possible so for now it might make sense to keep these despite the number of deps that are pulled in as a result.  Going to close this for now, lmk if you have any additional thoughts or concerns with this decision.  Thanks

PR #2637, opened 2023-07-25
Allows to disable the fallback to the default registry on image pull

### LINKED ISSUE #2445: Avoid registry fallback
Hi all!
I want to use a private registry in Kaniko, but only that registry.

I am implementing an image whitelist in the private registry, but when i return a 403 for a non whitelisted image, Kaniko fallback to dockerhub and the image is pulled from there anyway.
Is there a way to prevent this behavior?

Im using the _--registry-mirror_ argument with the  private registry address

Thanks very much in advance!

fgiannetti:
Fixes #2445

**Description**

When one or more registry mirror(s) are defined with the 'registry-mirror' flag, if none of those mirrors can pull the image, the current behavior is to fallback to the default registry. If a whitelist (or some image restriction) is applied at the mirror side, fallbacking to the default registry makes that restriction useless. This new argument allows to skip the fallback and abort the build if the mirror do not contains or rejects an image. If it is not set, is completely transparent.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

The fallback to default registry can be disabled if you defined a registry-mirror list and none of them can make pull of an image.

```
- kaniko adds a new flag `--skip-default-registry-fallback` to disable the fallback to the default registry if none of the  mirrors (if defined) cannot pull an image

```


aaron-prindle (approved):
@fgiannetti thanks for the PR here!  LGTM!

PR #2652, opened 2023-07-31
Avoid redundant calls to filepath.Clean

aaronlehmann:
`filepath.Clean` shows up in profiles as a hot spot, and there seem to be many redundant calls, particularly in ignorelist handling. We can avoid these redundant calls by pre-cleaning entries in the ignore list, and providing fast paths when we know we're already dealing with a cleaned candidate path.

Before:

     580ms  3.03% 72.35%      590ms  3.08%  path/filepath.(*lazybuf).append (inline)
     390ms  2.03% 74.39%      990ms  5.16%  path/filepath.Clean

After:

     0.13s  0.69% 84.01%      0.17s  0.91%  path/filepath.(*lazybuf).append (inline)
     0.13s  0.69% 84.70%      0.31s  1.65%  path/filepath.Clean

### DIFF
--- a/pkg/commands/copy_test.go
+++ b/pkg/commands/copy_test.go
@@ -153,7 +153,7 @@ func Test_CachingCopyCommand_ExecuteCommand(t *testing.T) {
 				extractedFiles: []string{"/foo.txt"},
 				contextFiles:   []string{"foo.txt"},
 			}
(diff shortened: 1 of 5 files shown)


aaron-prindle (approved):
Awesome, thanks for this @aaronlehmann!  LGTM

PR #2676, opened 2023-08-10
Fix: Change condition for the behaviour when --no-push=true without setting --destinations

### LINKED ISSUE #1613: When using "--no-push" and "--tarPath" error is thrown expecting "--destination" to be set.
**Actual behavior**
Command run 
```
/kaniko/executor --dockerfile=Dockerfile --no-push --tarPath=image.tar \
  --context=dir:///workspace "${opt_args}" --verbosity=debug \
  >/workspace/builder.log 2>&1
```

Returns error in log:
```
error pushing image: must provide at least one destination when tarPath is specified
```

**Expected behavior**
Expect no push to attempt to be executed and `image.tar` to appear in `/workspace`.

**To Reproduce**
1. Run command with both `--tarPath` and `--no-push`

**Additional Information**
 - The issue is with the order of the `opts` checks for `opts.tarPath` and `opts.noPush`. Because it checks for `opts.tarPath` first it throws an error expecting `--destination` to be set, but I specifically don't want to push which is why I'm using `--no-push`. 
Presumably the `opts.noPush` conditional should be at the very top because why run any part of `DoPush` if `--no-push` is set. https://github.com/GoogleContainerTools/kaniko/blob/0477900febfacb98b88abdbfaf8bbaeb05f51060/pkg/executor/push.go#L229-L248
 
 | **Description** | **Yes/Np** |

JeromeJu:
Fixes #1613

**Description**

This commit changes the condition check for the behavior when `--no-push` is set to true while destinations are needed. Prior this change, users would have to set destinations even when noPush option is set to true.

**Submitter Checklist**

review them:

- [n/a] Adds integration tests if needed.

**Reviewer Notes**

### DIFF
--- a/pkg/executor/push_test.go
+++ b/pkg/executor/push_test.go
@@ -222,6 +222,68 @@ func TestImageNameDigestFile(t *testing.T) {
 
 }
 
+func TestDoPushWithOpts(t *testing.T) {
+	tarPath := "image.tar"
+
+	for _, tc := range []struct {
+		name        string
+		opts        config.KanikoOptions
+		expectedErr bool
+	}{
+		{
+			name: "no push with tarPath without destinations",
+			opts: config.KanikoOptions{
+				NoPush:  true,
+				TarPath: tarPath,
+			},
+			expectedErr: false,
+		}, {
+			name: "no push with tarPath with destinations",
+			opts: config.KanikoOptions{
+				NoPush:       true,
+				TarPath:      tarPath,
+				Destinations: []string{"image"},
+			},
+			expectedErr: false,
+		}, {
+			name: "no push with tarPath with destinations empty",
+			opts: config.KanikoOptions{
+				NoPush:       true,
+				TarPath:      tarPath,
+				Destinations: []string{},
+			},
+			expectedErr: false,
+		}, {
+			name: "tarPath with destinations empty",
+			opts: config.KanikoOptions{
+				NoPush:       false,
+				TarPath:      tarPath,
+				Destinations: []string{},
+			},
+			expectedErr: true,
+		}} {
+		t.Run(tc.name, func(t *testing.T) {
+			image, err := random.Image(1024, 4)
+			if err != nil {
+				t.Fatalf("could not create image: %s", err)
+			}
+			defer os.Remove("image.tar")
+
+			err = DoPush(image, &tc.opts)
+			if err != nil {
+				if !tc.expectedErr {
+					t.Errorf("unexpected error with opts: could not push image: %s", err)
+				}
+			} else {
+				if tc.expectedErr {
+					t.Error("expected error with opts not found")
+				}
+			}
+
+		})
+	}
+}
+
 func TestImageNameTagDigestFile(t *testing.T) {
 	image, err := random.Image(1024, 4)
 	if err != nil {
--- a/pkg/executor/push.go
+++ b/pkg/executor/push.go
@@ -53,11 +53,13 @@ type withUserAgent struct {
 
 // for testing
 var (
-	newRetry = transport.NewRetry
+	newRetry          = transport.NewRetry
+	DummyDestinations = []string{DummyDestination}
 )
 
 const (
 	UpstreamClientUaKey = "UPSTREAM_CLIENT_TYPE"
+	DummyDestination    = "docker.io/unset-repo/unset-image-name"
 )
 
 func (w *withUserAgent) RoundTrip(r *http.Request) (*http.Response, error) {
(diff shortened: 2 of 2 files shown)


aaron-prindle:
In testing with this PR, it seems that when `--no-push` and `--tar-path` are set w/ this PR, there is no longer an error message but the tar is not generated.  This likely not the desired outcome for users setting `--no-push` and `--tar-path`, they will expect a tar generated at the path

test command:
```
    docker run \
        -v "$HOME"/.config/gcloud:/root/.config/gcloud \
        -v "$context":/workspace \
        gcr.io/kaniko-project/executor:latest \
        --dockerfile "${dockerfile}" --context dir:///workspace/ \
        --cache="${cache}" --no-push --tarPath=/image.tar
```

output @ HEAD (w/o this PR)
```
INFO[0107] Applying label retention=prune               
error pushing image: must provide at least one destination when tarPath is specified

# ^ error message is printed, kaniko exits with failure and no tar is generated
```

output w/ this PR
```
INFO[0109] LABEL retention="prune"                      
INFO[0109] Applying label retention=prune               
INFO[0109] Skipping push to container registry due to --no-push flag 

# ^ kaniko exits with success BUT NO TAR IS GENERATED <-- undesirable UX
```

This makes sense as with this change, we short circuit `push` logic here:
https://github.com/GoogleContainerTools/kaniko/blob/f0c2ff471f899b2db3018a1319c333feeab2d4c2/pkg/executor/push.go#L210

Which means we don't get to the tarPath logic here:
https://github.com/GoogleContainerTools/kaniko/blob/f0c2ff471f899b2db3018a1319c333feeab2d4c2/pkg/executor/push.go#L213-L227

Fixing this is somewhat non-trivial as currently the tarPath logic requires usage of a map that is empty when no destination is set:
https://github.com/GoogleContainerTools/kaniko/blob/f0c2ff471f899b2db3018a1319c333feeab2d4c2/pkg/executor/push.go#L221

While the logic here could be tweaked to not require a `destRef` mapping, it is additionally complicated as currently the tar file kaniko creates via `--tar-path` has a manifest.json that has `RepoTags` values set to the `--destination` value(s).  Will need to check if this is an optional field or a dummy value could be supplied. 

From this, the simplest fix would likely be to modify the kaniko logic s.t. when `--tar-path` is used w/o `--destination`, a "dummy" destination is set.  This would allow for the currently necessary `tagToImage[destRef] = image` line to work as expected as a dummy destRef would exist in the case none were set

JeromeJu:
Thanks @aaron-prindle , I'm trying to get a better picture of different comb of opts here:

|   opts             |  value         |                 |           |
| ------------  | ---------   | --------- | -----|
| NoPush          |  True         |       True       |
| Destinations  |  empty      |   empty ||
| TarPath          | not empty |  empty | | 
| behavior        | generate a tar at TarPath |    | |

aaron-prindle:
@JeromeJu 

> Thanks @aaron-prindle , I'm trying to get a better understanding of different combinations of opts here, wondering if the following makes sense:
> 
> opts	Case i	Case ii	Case iii
> --no-push	True	True	False
> --destinations	Empty	Empty	Empty
> --tar-path	Not empty	Empty	Either empty or not
> Behaviour	Generate a tar at TarPath, use the dummy destRef workaround	Short circuit in DoPush	Return the previous Error(_error pushing image: must provide at least one destination when tarPath is specified_) where destinations are missing

Yes +1, the behaviour for the three cases you outlined is exactly what is desired

pkg/executor/push_test.go
```diff
@@ -222,6 +222,68 @@ func TestImageNameDigestFile(t *testing.T) {
...
+		}, {
+			name: "no push with tarPath with destinations empty",
+			opts: config.KanikoOptions{
+				NoPush:       true,
+				TarPath:      tarPath,
+				Destinations: []string{},
+			},
+			expectedErr: false,
+		}, {
+			name: "no push with tarPath with destinations empty",
```
aaron-prindle, line 257: nit: this and the above test have the same name which might make debugging a failing test more difficult.  I think this test should likely be named: 
"tarPath with destinations empty"

aaron-prindle:
Just tried it and it works great!  I left one small comment regarading a test case name being a dupe. After that is resolved we should be good to merge! 🎊 

As expected the dummy value is present in the manifest.json
```
aprindle@aprindle-ssd ~/kaniko-tar-path/kaniko cat manifest.json 
[{"Config":"sha256:fcbf14c4f26da0e408cd84d3ff07a09db17eb56dcc9ff3c4deaf7a8f98b90ce7","RepoTags":["docker.io/unset-repo/unset-image-name:latest"],"Layers":["396c31837116ac290458afcb928f68b6cc1c7bdd6963fc72f52f365a2a89c1b5.tar.gz","ddd8d01266dd2012b35dd4adfea327d1dd5cbf1431e02a28d32fa6d709b5eff8.tar.gz"]}]aprindle@aprindle-ssd

```

and as discussed the dummy value existing in the tar should not be an issue for users ommitting the `--destination` flag.  User's who rely on a specific value veing in the `manifest.json` for `RepoPaths` can still set the now optional `--destination` as they did previously

pkg/executor/push_test.go
JeromeJu: Thanks for the catch! 🙏

JeromeJu:
Thanks for the review @aaron-prindle . Updated.

aaron-prindle (approved):
LGTM!  Thanks for the PR @JeromeJu! 🎊

PR #2686, opened 2023-08-18
fix: resolve issue where kaniko would fail if user attempted to create /workspace directory

### LINKED ISSUE #1508: Dockerfile cannot create /workspace directory already exists, is it dangerous to use?
**Actual behavior**
If a dockerfile has a step to create a directory /workspace, kaniko fails.

**Expected behavior**
Parity with the docker behavior, I expected it to pass like docker build

**To Reproduce**
Simple Dockerfile
```
FROM alpine:3.9
RUN mkdir /workspace
```

*running with docker passes*
```
docker build -f Dockerfile-workspace .
Sending build context to Docker daemon  2.816MB
Step 1/2 : FROM alpine:3.9
 ---> 78a2ce922f86
Step 2/2 : RUN mkdir /workspace
 ---> Running in 37d72caee37d
Removing intermediate container 37d72caee37d
 ---> c5c09e97f416
Successfully built c5c09e97f416
```

**running with kaniko fails with mkdir: can't create directory '/workspace': File exists**
```
docker run -v `pwd`:/kaniko-workspace/ gcr.io/kaniko-project/executor:latest --no-push --dockerfile Dockerfile-workspace  --context /kaniko-workspace/ --verbosity debug --destination=image:tag --tarPath=/kaniko-workspace/image.tar
DEBU[0000] Copying file /kaniko-workspace/Dockerfile-workspace to /kaniko/Dockerfile
DEBU[0000] Skip resolving path /kaniko/Dockerfile
DEBU[0000] Skip resolving path /kaniko-workspace/
DEBU[0000] Skip resolving path /cache
DEBU[0000] Skip resolving path /kaniko-workspace/image.tar
DEBU[0000] Skip resolving path
DEBU[0000] Skip resolving path
DEBU[0000] Built stage name to index map: map[]
INFO[0000] Retrieving image manifest alpine:3.9
INFO[0000] Retrieving image alpine:3.9
DEBU[0002] No file found for cache key sha256:65b3a80ebe7471beecbc090c5b2cdd0aafeaefa0715f8f12e40dc918a3a70e32 stat /cache/sha256:65b3a80ebe7471beecbc090c5b2cdd0aafeaefa0715f8f12e40dc918a3a70e32: no such file or directory
DEBU[0002] Image alpine:3.9 not found in cache
INFO[0002] Retrieving image manifest alpine:3.9
INFO[0002] Retrieving image alpine:3.9
INFO[0003] Built cross stage deps: map[]
INFO[0003] Retrieving image manifest alpine:3.9
INFO[0003] Retrieving image alpine:3.9
DEBU[0004] No file found for cache key sha256:65b3a80ebe7471beecbc090c5b2cdd0aafeaefa0715f8f12e40dc918a3a70e32 stat /cache/sha256:65b3a80ebe7471beecbc090c5b2cdd0aafeaefa0715f8f12e40dc918a3a70e32: no such file or directory
DEBU[0004] Image alpine:3.9 not found in cache
INFO[0004] Retrieving image manifest alpine:3.9
INFO[0004] Retrieving image alpine:3.9
INFO[0005] Executing 0 build triggers
INFO[0005] Unpacking rootfs as cmd RUN mkdir /workspace requires it.
DEBU[0005] Mounted directories: [{/kaniko false} {/etc/mtab false} {/tmp/apt-key-gpghome true} {/var/run false} {/proc false} {/dev false} {/dev/pts false} {/sys false} {/sys/fs/cgroup false} {/sys/fs/cgroup/cpuset false} {/sys/fs/cgroup/cpu false} {/sys/fs/cgroup/cpuacct false} {/sys/fs/cgroup/blkio false} {/sys/fs/cgroup/memory false} {/sys/fs/cgroup/devices false} {/sys/fs/cgroup/freezer false} {/sys/fs/cgroup/net_cls false} {/sys/fs/cgroup/perf_event false} {/sys/fs/cgroup/net_prio false} {/sys/fs/cgroup/hugetlb false} {/sys/fs/cgroup/pids false} {/sys/fs/cgroup/rdma false} {/sys/fs/cgroup/systemd false} {/dev/mqueue false} {/dev/shm false} {/kaniko-workspace false} {/etc/resolv.conf false} {/etc/hostname false} {/etc/hosts false} {/proc/bus false} {/proc/fs false} {/proc/irq false} {/proc/sys false} {/proc/sysrq-trigger false} {/proc/acpi false} {/proc/kcore false} {/proc/keys false} {/proc/timer_list false} {/proc/sched_debug false} {/sys/firmware false}]
DEBU[0006] Not adding /dev because it is ignored
DEBU[0006] Not adding /etc/hostname because it is ignored
DEBU[0006] Not adding /etc/hosts because it is ignored
DEBU[0006] Not adding /etc/mtab because it is ignored
DEBU[0006] Not adding /proc because it is ignored
DEBU[0006] Not adding /sys because it is ignored
DEBU[0006] Not adding /var/run because it is ignored
INFO[0006] RUN mkdir /workspace
INFO[0006] Taking snapshot of full filesystem...
INFO[0006] cmd: /bin/sh
INFO[0006] args: [-c mkdir /workspace]
INFO[0006] Running: [/bin/sh -c mkdir /workspace]
mkdir: can't create directory '/workspace': File exists
error building image: error building stage: failed to execute command: waiting for process to exit: exit status 1
```

**Additional Information**
- I stumbled on this issue and couldn't find any documentation or report in kaniko for this. It is easy to workaround the issue by putting `mkdir -p /workspace` but I'm worried that kaniko uses that directory internally and it could conflict with what docker is trying to do. Can one of the maintainers confirm? I can imagine more complex examples where permissions and user chown to that directory could potentially be needed for the docker container and mess things up? Is it bad practice to have a dockerfile use /workspace? If the answer is yes, in that case this should be documented, or I would prefer the kaniko code be changed so that users do not inavertently use internal directory structures needed by kaniko.

```
        "Id": "sha256:d4478ae513eed31b69e375a681fee25592cd1e3618ce28f9633af841cba24410",
        "RepoTags": [
            "gcr.io/kaniko-project/executor:latest"
        ],
```
 
aaron-prindle:
Fixes #1508 

Context: 
https://github.com/GoogleContainerTools/kaniko/issues/1508#issuecomment-1683187462

Currently a `/workspace` folder is created when kaniko is run as it is the WORKDIR for the kaniko image.  This `/workspace` folder is deprecated though currently, this directory was renamed `/kaniko` and it is unclear if the WORKDIR should be set at all for users IIUC.  This change removes a WORKDIR from being set which fixes issues where kaniko could not create a `/workspace` directory as it already existed from it being set as the image's WORKDIR

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -83,7 +83,6 @@ COPY --from=builder /kaniko/.docker /kaniko/.docker
 
 ENV DOCKER_CONFIG /kaniko/.docker/
 ENV DOCKER_CREDENTIAL_GCR_CONFIG /kaniko/.config/gcloud/docker_credential_gcr_config.json
-WORKDIR /workspace
 
 ### FINAL STAGES ###
 


JeromeJu:
It seems from the integration tests that we might want to add /workspace for https://github.com/GoogleContainerTools/kaniko/blob/main/integration/dockerfiles/Dockerfile_relative_copy after the change IIUC.

aaron-prindle:
@JeromeJu thanks for the info here.  I am hesitant to make this change currently as this test outlines a potential issue where changing the Kaniko image to no longer set the WORKDIR to /workspace might break some users/Dockerfiles that rely on the current functionality. 

Additionally this test `Dockerfile_relative_copy` is essentially testing that WORKDIR is auto-set for relatvie path COPY and would not make sense to change to `/workspace` (as the test would no longer test `relative_copy`.  Still debating the best thing to do here, it might make sense to simply remove this as it no longer makes sense or have a special kaniko image with a WORKDIR set to test this functionality even if the default image doesn't set it

aaron-prindle:
Going to close this for now until I have more definitive information on the potential user breakage vs the fix upside (Dockerfiles created /workspace work)

PR #2790, opened 2023-10-11
fix: remove log line from listpullreqs.go and additional release.sh fixes

aaron-prindle:


JeromeJu (approved):
Thanks for the fix!
LGTM

Issue #1598, opened 2021-03-11
Can we put `context.ExcludesFile` before `os.Lstat` when copyDir?


keshin:
Hi team,

We have a Jenkins job which runs two stages in parallel, 
1. stage A might delete some files under the folder defined in `.dockerignore`
2. stage B runs kaniko executor to build/push docker images with the `.dockerignore`

If a file under the ignore folder is deleted after 

https://github.com/GoogleContainerTools/kaniko/blob/eda57e5400e4cdecccb2a3d9db445e57e81a9c4f/pkg/util/fs_util.go#L591-L594

Then an error would be raise when calling `os.Lstat`, even the file is meant to be ignored.

https://github.com/GoogleContainerTools/kaniko/blob/eda57e5400e4cdecccb2a3d9db445e57e81a9c4f/pkg/util/fs_util.go#L597-L606

Questions:
1. May I know why we need to put `os.Lstat` before `context.ExcludesFile?`?
2. Is there a way that we can prevent such issue happens?

jimmy-guzman:
Just sharing that we're seeing this behavior as well in our pipelines and we also came to the same conclusion.

PR #2812, opened 2023-10-23
fix: fix COPY command error due to missing but ignored files

QuanZhang-William:
**Description**
Fixes https://github.com/GoogleContainerTools/kaniko/issues/1598.

This commit puts `context.ExcludesFile` before `os.Lstat` to avoid the `COPY` command error due to missing but ignored files.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
Ignore file missing error in `COPY` command if the file is to be ignored
```

### DIFF
--- a/pkg/commands/copy_test.go
+++ b/pkg/commands/copy_test.go
@@ -20,6 +20,7 @@ import (
 	"archive/tar"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/commands/copy_test.go
```diff
@@ -461,6 +462,60 @@ func TestCopyCommand_ExecuteCommand_Extended(t *testing.T) {
...
+		}
+
+		cfg := &v1.Config{
+			Cmd:        nil,
+			Env:        []string{},
+			WorkingDir: testDir,
+		}
+
+		err = cmd.ExecuteCommand(cfg, dockerfile.NewBuildArgs([]string{}))
+		if err != nil {
```
JeromeJu, line 497: Could you help me to understand what's the error as described in #1598 could have happened without this fix?

pkg/commands/copy_test.go
```diff
@@ -461,6 +462,60 @@ func TestCopyCommand_ExecuteCommand_Extended(t *testing.T) {
...
+			expected[sf.Name()] = sf
+		}
+
+		cmd := CopyCommand{
+			cmd: &instructions.CopyCommand{
+				SourcesAndDest: instructions.SourcesAndDest{SourcePaths: []string{srcDir}, DestPath: "dest"},
+			},
+			fileContext: util.FileContext{
+				Root:          testDir,
+				ExcludedFiles: []string{filepath.Join(srcDir, "bam.txt")}},
```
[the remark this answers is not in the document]
QuanZhang-William: Good catch, fixed!

pkg/commands/copy_test.go
QuanZhang-William: Hi @JeromeJu. IIUC #1598 describes a racing condition between 2 process. Say process A 1) finds all the files to be copied; and 2) copies the files one by one. If process B kicks in and deletes some file between 1) and 2), it will throw an error. However, if the files to be copied are supposed to be ignored, it should not fail since the files will not be copied anyways. That's all swapping the 2 function calls solve the problem.

Regarding the tests, I didn't find a way to mock the race condition in unit test, so I just added a test case with ignored files. Please let me know if you have better way to test it 😁

aaron-prindle: One way of doing this is via "monkey patching" (done elsewhere in the code) where we can mock out `RemainingFiles` or `os.Lstat` to immediately delete an excluded file after processing and use this mock for a repro test.  I have done this here - https://gist.github.com/aaron-prindle/7f7fd1f43dc5c3fe4e9b14f69ecdd9e7

I don't think it is necessary for the code in this case though, I have manually tested with the above and things seem to be working.  Looks good!

aaron-prindle (approved):
LGTM!  Left one comment related to adding a test that repro's the old issue exactly.  I have manually tested with such a test and it is working as expected

pkg/commands/copy_test.go
```diff
@@ -461,6 +462,60 @@ func TestCopyCommand_ExecuteCommand_Extended(t *testing.T) {
 		}
 	})
 
+	t.Run("copy dir to another dir - with ignored files", func(t *testing.T) {
```
aaron-prindle, line 465: Thanks for adding this!

Issue #1333, opened 2020-06-29
Image build process Freezes on `Taking snapshot of full filesystem...`


abhi1git:
**Actual behavior**
While building image using **gcr.io/kaniko-project/executor:debug** in **gitlab CI runner** hosted on **kubernetes** using helm chart the image build process freezes on **Taking snapshot of full filesystem...** for the time till the runner timeouts(1 hr)
This behaviour is intermittent as for the same project image build stage works sometimes

Issue arises in multistage as well as single stage Dockerfile.

**Expected behavior**
Image build should not freeze at `Taking snapshot of full filesystem...` and should be successful everytime.

**To Reproduce**
As the behaviour is intermittent not sure how it can be reproduced
 
 | **Description** | **Yes/No** |
 |----------------|---------------|
 | Please check if this a new feature you are proposing        | <ul><li>- [ ] </li></ul>|
 | Please check if the build works in docker but not in kaniko | <ul><li>- [Yes ] </li></ul>| 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [ ] </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [ ] </li></ul>| 
 
@tejal29

abhi1git:
Can you please elaborate snapshot for which filesystem is being taken while building image so that we can see if filesystem size is causing this issue. we are using kaniko to build images in gitlab cicd and runner is deployed on kubernetes using helm chart.
Preiously this issue used to arise randomly but all of our kaniko build image jobs get freeze on `Taking snapshot of full filesystem...`
@tejal29

tejal29:
@abhi1git can you try the newer snapshot modes `--snapshotMode=redo`?

tejal29:
@abhi1git please switch to using `--snapshotMode=redo`. See comments here https://github.com/GoogleContainerTools/kaniko/issues/1305#issuecomment-672752902

Kiddinglife:
**@abhi1git please switch to using --snapshotMode=redo. See comments here #1305 (comment)**
I suffered from the same issue and --snapshotMode=redo did not resolve the issue.

tejal29:
@Kiddinglife Can you provide your dockerfile or some stats on number of files in your repo?

leoschet:
I am experience this problem while building an image with less than a gb. Interesting that it fails silently. GitLab CI job will be marked as successfull but no image is actually pushed.

[...]

PR #2816, opened 2023-10-26
snapshotter: use syncfs system call

zhouhaibing089:
`sync` system call triggers a full page cache sync which may not always work, especially in kubernetes environment where it is very easy to be interfered by others. I have seen several cases where a broken nfs mount is blocking kaniko from doing its job.

With `syncfs`, it only writes cache back to disk for the current filesystem that is used by kaniko which is supposed to be more reliable.

Here is the details that I captured recently. From kaniko executor logs:

```console
/workspace # executor --no-push
INFO[0000] Retrieving image manifest ubuntu:22.04
INFO[0000] Retrieving image ubuntu:22.04 from registry index.docker.io
INFO[0001] Built cross stage deps: map[]
INFO[0001] Retrieving image manifest ubuntu:22.04
INFO[0001] Returning cached image manifest
INFO[0001] Executing 0 build triggers
INFO[0001] Building stage 'ubuntu:22.04' [idx: '0', base-idx: '-1']
INFO[0001] Unpacking rootfs as cmd RUN apt-get update requires it.
INFO[0003] RUN apt-get update
INFO[0003] Initializing snapshotter ...
INFO[0003] Taking snapshot of full filesystem...
```

It hangs there forever in `Taking snapshot of full filesystem` - see https://github.com/GoogleContainerTools/kaniko/issues/1333 as well.

I looked the stack of kaniko process and see this:

```console
root@<my-host-name>:/proc/180716/task# cat 182585/stack
[<0>] sync_inodes_sb+0x10b/0x2b0
[<0>] sync_inodes_one_sb+0x15/0x20
[<0>] iterate_supers+0xa0/0x100
[<0>] ksys_sync+0x42/0xb0
[<0>] __do_sys_sync+0xe/0x20
[<0>] do_syscall_64+0x59/0xc0
[<0>] entry_SYSCALL_64_after_hwframe+0x44/0xae
```

where `180716` is the pid of kaniko executor, and one of its subtask is waiting in `ksys_sync` forever.

At that point of time, running `sync` (command) on the host was also stuck forever. We further checked and then identified the system call was waiting for one of nfs mount to flush its superblock (but due to connectivity issues, it can't complete).

```console
root@<my-host-name>:/# dmesg -T | grep nfs
[Thu Oct 26 04:05:18 2023] FS-Cache: Netfs 'nfs' registered for caching
[Thu Oct 26 04:12:00 2023]  ? nfs_ctx_key_to_expire+0x69/0xf0 [nfs]
[Thu Oct 26 04:12:00 2023]  nfs_wb_all+0x28/0xf0 [nfs]
[Thu Oct 26 04:12:00 2023]  nfs4_file_flush+0x73/0xb0 [nfsv4]
[Thu Oct 26 04:12:28 2023] nfs: server <nfs-server-host-name> not responding, still trying
[Thu Oct 26 04:12:43 2023] nfs: server <nfs-server-host-name> not responding, still trying
[Thu Oct 26 04:14:00 2023]  ? nfs_ctx_key_to_expire+0x69/0xf0 [nfs]
[Thu Oct 26 04:14:00 2023]  nfs_wb_all+0x28/0xf0 [nfs]
[Thu Oct 26 04:14:00 2023]  nfs4_file_flush+0x73/0xb0 [nfsv4]
```

While this is technically not related to kaniko, but by switching to `syncfs` with fd from current directory, it is going to only sync the current file system which holds the provided fd, and this is not affected by external filesystems.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
snapshotter: use syncfs system call rather than sync
```

### DIFF
--- a/pkg/snapshot/snapshot.go
+++ b/pkg/snapshot/snapshot.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"syscall"
 
@@ -31,6 +32,7 @@ import (
 	"github.com/GoogleContainerTools/kaniko/pkg/util"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 // For testing
@@ -155,7 +157,20 @@ func (s *Snapshotter) scanFullFilesystem() ([]string, []string, error) {
 	// for example the hashing function that determines if files are equal uses the mtime of the files,
 	// which can lag if sync is not called. Unfortunately there can still be lag if too much data needs
 	// to be flushed or the disk does its own caching/buffering.
-	syscall.Sync()
+	if runtime.GOOS == "linux" {
+		dir, err := os.Open(s.directory)
+		if err != nil {
+			return nil, nil, err
+		}
+		defer dir.Close()
+		_, _, errno := syscall.Syscall(unix.SYS_SYNCFS, dir.Fd(), 0, 0)
+		if errno != 0 {
+			return nil, nil, errno
+		}
+	} else {
+		// fallback to full page cache sync
+		syscall.Sync()
+	}
 
 	s.l.Snapshot()
 


pkg/snapshot/snapshot.go
```diff
@@ -155,7 +155,16 @@ func (s *Snapshotter) scanFullFilesystem() ([]string, []string, error) {
...
 	// which can lag if sync is not called. Unfortunately there can still be lag if too much data needs
 	// to be flushed or the disk does its own caching/buffering.
-	syscall.Sync()
+	dir, err := os.Open(s.directory)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer dir.Close()
+	// 306 is the number for sys_syncfs
+	_, _, errno := syscall.Syscall(306, dir.Fd(), 0, 0)
```
aaron-prindle, line 164: nit: Can you add some OS related checks to ensure that this is only done for operating systems that support sys_syncfs.  Perhaps look at `runtime.GOOS` and only do the sys_syncfs logic for linux OS?  Admittedly kaniko is mostly run in a container on top of linux containerization but just for completeness/future-proofing this check might make sense

aaron-prindle:
Thanks for the PR here @zhouhaibing089, appreciate the fix!  Left one nit comment, once that is addressed we can get this merged

pkg/snapshot/snapshot.go
zhouhaibing089: Sounds good. Thanks for the suggestion.

zhouhaibing089: Ah, `306` is also amd64 specific. I think I can leverage golang.org/x/sys/unix.

aaron-prindle (approved):
LGTM!  Thanks @zhouhaibing089!

rayunstop:
great job, help a lot. Thanks.

PR #2843, opened 2023-11-11
refactor: rm bool param detectFilesystem in `InitIgnoreList`

### LINKED ISSUE #2840: [refactor] Cleanup `InitIgnoreList(detectFilesystem bool)`
Currently `func InitIgnoreList(detectFilesystem bool) error {` takes ini a bool parameter but in all code, the value is only ever set to true.  This issue tracks removing the bool parameter as it is unused:

Related code:
https://github.com/GoogleContainerTools/kaniko/blob/main/pkg/util/fs_util.go#L1035

tal66:
hi, this closes #2840.

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -101,7 +101,7 @@ func newStageBuilder(args *dockerfile.BuildArgs, opts *config.KanikoOptions, sta
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


JeromeJu (approved):
Thanks for the cleanup! 🎉

Issue #2033, opened 2022-04-05
Multi stage build ends with "no space left on device", no matter the free space you provide


bmalynovytch:
**Actual behavior**
When trying to build a multi stage container, Kaniko drains *all* the local storage available (source images weight between 1Gi to 2Gi, Kaniko drains up to 500Gi of storage) and crashes with `no space left on device`.

**Logs**
```
INFO[0007] Resolved base name docker.io/bitnami/redis:6.2 to redis 
INFO[0007] Resolved base name docker.io/bitnami/memcached:1.6.13 to memcached 
INFO[0007] Resolved base name docker.io/zulip/zulip-postgresql:10 to postgresql 
INFO[0007] Resolved base name docker.io/zulip/docker-zulip:4.10-0 to zulip 
INFO[0007] Retrieving image manifest docker.io/bitnami/redis:6.2 
INFO[0007] Retrieving image docker.io/bitnami/redis:6.2 from registry index.docker.io 
INFO[0008] Retrieving image manifest docker.io/bitnami/memcached:1.6.13 
INFO[0008] Retrieving image docker.io/bitnami/memcached:1.6.13 from registry index.docker.io 
INFO[0009] Retrieving image manifest docker.io/zulip/zulip-postgresql:10 
INFO[0009] Retrieving image docker.io/zulip/zulip-postgresql:10 from registry index.docker.io 
INFO[0010] Retrieving image manifest docker.io/zulip/docker-zulip:4.10-0 
INFO[0010] Retrieving image docker.io/zulip/docker-zulip:4.10-0 from registry index.docker.io 
INFO[0011] Retrieving image manifest docker.io/bitnami/rabbitmq:3.9.13 
INFO[0011] Retrieving image docker.io/bitnami/rabbitmq:3.9.13 from registry index.docker.io 
INFO[0012] Built cross stage deps: map[0:[/] 1:[/] 2:[/] 3:[/]] 
INFO[0012] Retrieving image manifest docker.io/bitnami/redis:6.2 
INFO[0012] Returning cached image manifest              
INFO[0012] Executing 0 build triggers                   
INFO[0016] Saving file . for later use                  
error building image: could not save file: write /kaniko/0/dev/full: no space left on device
```

**Dockerfile** (extract)
```
FROM docker.io/bitnami/redis:6.2         AS redis
FROM docker.io/bitnami/memcached:1.6.13  AS memcached
[...]

PR #2863, opened 2023-11-21
Skip the /kaniko directory when copying root

JeromeJu:
Prior to this change, the files under the `/kaniko` dir is also attempted to be copied. This commit adds the skip option for otiai10.Copy to skip the /kaniko directory when the root is being copied. The files under /kaniko dir should be ignored and thus this shall not cause any loss of information.
fixes: #2033

**Description**

**Submitter Checklist**

review them:

**Reviewer Notes**

### DIFF
--- a/integration/dockerfiles/Dockerfile_test_copy_root_multistage
+++ b/integration/dockerfiles/Dockerfile_test_copy_root_multistage
@@ -1,0 +1,5 @@
+FROM busybox:1.31
+COPY context/foo foo
+
+FROM alpine@sha256:5ce5f501c457015c4b91f91a15ac69157d9b06f1a75cf9107bf2b62e0843983a
+COPY / /foo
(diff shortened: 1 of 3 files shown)


integration/dockerfiles/Dockerfile_test_copy_root_multistage
```diff
@@ -0,0 +1,5 @@
+FROM busybox:1.31
+COPY context/foo foo
+
+FROM alpine@sha256:5ce5f501c457015c4b91f91a15ac69157d9b06f1a75cf9107bf2b62e0843983a
```
QuanZhang-William, line 4: I don't have full context, but could you help explain how does this dockerfile tests the logic change?

QuanZhang-William: Also, is the release note related to the PR?

JeromeJu: > I don't have full context, but could you help explain how does this dockerfile tests the logic change?

Prior to this change, when the source "/" is specified, we will see errors described at https://github.com/GoogleContainerTools/kaniko/issues/2033#issuecomment-1766604713 with recursive `/kaniko/0` in the names. This test case test copying from root and the `kaniko/0` folder shall not be apparent.

I also added a sentence for the previous situation in the message. Thanks for pointing this out.

aaron-prindle (approved):
LGTM!  Just testing this locally and it works great, thanks for fixing this!

PR #2935, opened 2024-01-07
Add flag to remap registries for any registry mirror

babs:
**Description**

The purpose of this PR is to add an option to remap registries, a kind of generalized `--registry-mirror`. This is helpful for air-gapped environments and/or when local registry mirrors are available (not limited to docker.io). This allows user to reference any images without having to change their location. It also permit to separate infra related configuration (the mirrors) from CI/CD pipeline definition by using an environment variable for example (the reason behind the early return if flag provided but empty). Therefore you can have a pipeline calling kaniko with `--registry-map=$REGISTRY_MAP` and have the `REGISTRY_MAP` populated via the runner's env by another team, and the absence of env wouldn't trigger a failure, it makes the pipeline env independent.

~I've also considered the option of environment variables directly but it doesn't seems to be in kaniko's philosophy.~ (see [update 1](#update1))

~This makes quite some duplicated code :/ One option to keep the mirror flag and behavior would be to use only one codebase and convert `--registry-mirror=VALUE` to `--registry-map=index.docker.io=VALUE` internally.~(see [update 1](#update1))
Suggestions welcome!

<a name="update1"></a>**[UPDATE 1]**
As there is a `KANIKO_REGISTRY_MIRROR`, I added `KANIKO_REGISTRY_MAP`.
I've also replaced the mirror logic with map, keeping the original behavior.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- kaniko adds a new flag `--registry-map` to provide a map of mirrors to use as pull-through cache instead. Expected format is 'orignal.registry=new.registry;other-original.registry=other-remap.registry'

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -987,6 +987,8 @@ You can mention several remap in a single flag too, separated by semi-colon.
 If an image is not found on the first mirror, Kaniko will try
 the next mirror(s), and at the end fallback on the original registry.
 
+Registry maps can also be defined through `KANIKO_REGISTRY_MAP` environment variable.
+
 Expected format is `original-registry=remapped-registry[;another-reg=another-remap[;...]]` for example.
 
 Note that you can't specify a URL with scheme for this flag. Some valid options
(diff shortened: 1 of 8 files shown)


babs:
Still open to suggestions ;-)

babs:
I've finally decided to replace registry mirror with map logic and allow use of environment variable as `KANIKO_REGISTRY_MIRROR` existed in the code

babs:
Related issue #2990

aaron-prindle (approved):
Appreciate your patience here @babs.  Thanks for the PR, LGTM

babs:
Np, happy to see it merged.
If I can help ;)

PR #2943, opened 2024-01-09
Prevent extra snapshot with --use-new-run

### LINKED ISSUE #2800: Redundant snapshot with --use-new-run
**Actual behavior**

When I pass `--use-new-run` I see the following toward the end of the logs:

```
INFO[0000] Initializing snapshotter ...                 
INFO[0000] Taking snapshot of full filesystem...
```

But I am not sure why a snapshot would be required, or if it makes sense that the behavior differs when running without `--use-new-run`.

**Expected behavior**

I would expect not to see any snapshotting.  There is no snapshot if I omit `--use-new-run`.  In some cases we are seeing a pretty long delay for the snapshot (12 seconds in one case) so avoiding it would speed up our builds.

**To Reproduce**

1. Create an empty directory.
2. Create a Dockerfile in that directory with only `FROM codercom/oss-dogfood` (or any other image, I believe).
3. Change into that directory and run something like:
```
docker run --rm -v $PWD:/workspace gcr.io/kaniko-project/executor:latest --no-push --use-new-run
```

**Additional Information**

I can prevent the snapshot by removing the `s.opts.RunV2` check found here: 

https://github.com/GoogleContainerTools/kaniko/blob/b6f14ae676b92ba5e27943c5a873c476d3e9a0f8/pkg/executor/build.go#L347

However, I am unsure if there are negative consequences to doing so.  I see [the commit](https://github.com/GoogleContainerTools/kaniko/commit/58276274e835d026e0550331bab1a08b41d5a693#diff-815db7a7d24db4a24dfcaa06524762a4d2631187778eff8cdc6834a50532f493R322) that added the check but it did not shed any light for me.

I tested my own build and ran `go test` without seeing any related issues though.  

If this does seem like something that can be removed, please let me know if it would be helpful for me to send a PR or if this is tiny enough that it is easier for someone else to just push it.
 
Thank you!

code-asher:
Fixes #2800

**Description**

This removes an extra snapshot that can occur when using `--use-new-run`.

**Submitter Checklist**

**Reviewer Notes**

**Release Notes**

```
--use-new-run will no longer always take a snapshot even if a snapshot was not necessary 
```

### DIFF
--- a/pkg/executor/build.go
+++ b/pkg/executor/build.go
@@ -344,7 +344,7 @@ func (s *stageBuilder) build() error {
 	}
 
 	initSnapshotTaken := false
-	if s.opts.SingleSnapshot || s.opts.RunV2 {
+	if s.opts.SingleSnapshot {
 		if err := s.initSnapshotWithTimings(); err != nil {
 			return err
 		}
--- a/pkg/executor/build_test.go
+++ b/pkg/executor/build_test.go
@@ -616,7 +616,7 @@ func Test_stageBuilder_optimize(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			cf := &v1.ConfigFile{}
-			snap := fakeSnapShotter{}
+			snap := &fakeSnapShotter{}
 			lc := &fakeLayerCache{retrieve: tc.retrieve}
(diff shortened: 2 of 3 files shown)


aaron-prindle (approved):
Sorry for the delay and thanks for the PR here @code-asher!  Just did some manually testing from the PR here and I agree this change should not change any critical functionality for kaniko w/ `--use-new-run`.  LGTM!

PR #2983, opened 2024-02-02
feat: add support for no push environment variable

### LINKED ISSUE #2385: Feature request: Allow `--no-push` configuration through environment variable
Hello everyone,

I won't use the bug issue template because this is not a bug report but a request for new functionality.

We run our container builds using shell scripts and would prefer to be able to set an environment variable for skipping push when building the images in merge/pull requests for example. 
That would allow to keep build script as they are and only set a variable from the job configuration.

My idea is to do the same as for the `KANIKO_REGISTRY_MIRROR` variable.
The variable could be called `KANIKO_NO_PUSH` for example.

I am not a Go developper so I have no idea how to parse environment variable but I guess it might quite similar to what is done [here](https://github.com/GoogleContainerTools/kaniko/blob/61312a95ae1a305f5154f8d88dc58ee1e3259ae8/cmd/executor/cmd/root.go#L69-L71).
If someone can confirm, I could open a PR to add what's missing. If it is more complicated, I won't be able to help.

Thanks in advance.
Cheers

sachasmart:
Supports #2385 and #144 _in case of a bug fix, this should point to a bug and any other related issue(s)_

**Description**

This feature adds support for checking the `KANIKO_NO_PUSH` environment variable. If the environment variable is declared, it sets the value to `opts.NoPush`. 

Side note - fully aware there may be an opportunity to use [viper](https://github.com/spf13/viper) here. Happy to help implement it, but I was not sure if this was a path the project was interested in. 

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
Add KANIKO_NO_PUSH environment variable to set --no-push flag
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ specify the location of your build context:
 | Source             | Prefix                                                                | Example                                                                       |
 | ------------------ | --------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
 | Local Directory    | dir://[path to a directory in the kaniko container]                   | `dir:///workspace`                                                            |
-| Local Tar Gz       | tar://[path to a .tar.gz in the kaniko container]                     | `tar:///path/to/context.tar.gz`                                                |
+| Local Tar Gz       | tar://[path to a .tar.gz in the kaniko container]                     | `tar:///path/to/context.tar.gz`                                               |
 | Standard Input     | tar://[stdin]                                                         | `tar://stdin`                                                                 |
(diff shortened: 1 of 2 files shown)


JeromeJu (approved):
Thanks for this @sachasmart , would you mind also help add this to the Readme.md?

sachasmart:
Perfect - updated with:
```
#### Flag `--no-push`

Set this flag if you only want to build the image, without pushing to a
registry. This can also be defined through `KANIKO_NO_PUSH` environment
variable.
```

JeromeJu (approved):

PR #3049, opened 2024-03-06
test: fix test breakage caused by external dependency update

aaron-prindle:


JeromeJu (approved):
Thanks @aaron-prindle , LGTM for the patch.

PR #3053, opened 2024-03-07
Fix #3032: Remove query parameters in ADD command when the destinatio…

### LINKED ISSUE #3032: ADD using an url with GET parameters and a directory destination includes the GET parameters in the file name
**Actual behavior**
The GET parameters of the url are included in the filename when the destination is a directory.

**Expected behavior**
The filename should have the GET parameters removed.

**To Reproduce**
1. Create a Dockerfile with the following contents
```
FROM ubuntu:jammy
ADD https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf?ext=.txt test-dir/
RUN ls -l test-dir
```
3. Build using docker to observe the output from line 3 is 
```
total 16
-rw------- 1 root root 13264 Aug 27  2007 dummy.pdf
```
4. Build using kaniko to observe the output from line 3 is
```
total 16
-rw------- 1 root root 13264 Aug 27  2007 dummy.pdf?ext=.txt
```

**Additional Information**
`gcr.io/kaniko-project/executor@sha256:ac169723b2076f9d5804f4bc05c98397e286da6fdcdd5a09fdc179f06ccb3be1`
 - Location of issue: https://github.com/GoogleContainerTools/kaniko/blob/main/pkg/util/command_util.go#L218
 - Caused by: [filepath.Base()](https://pkg.go.dev/path/filepath#Base) returns the last element of path and does not account for urls with GET parameters
 
prima101112:
Fixes #3032 

**problem**

`ADD https://something?asd dir/` filename returned something?asd

**expected : and fixed**

the filename should be `something`

**Description**

fix filename.Base still returning query params

**Submitter Checklist**

review them:

- [  ] Adds integration tests if needed.

**Reviewer Notes**

**Release Notes**

```
- fix #3032 

```

### DIFF
--- a/pkg/util/command_util.go
+++ b/pkg/util/command_util.go
@@ -216,11 +216,17 @@ func URLDestinationFilepath(rawurl, dest, cwd string, envs []string) (string, er
 		}
 		return dest, nil
 	}
-	urlBase := extractFilename(rawurl)
-	urlBase, err := ResolveEnvironmentReplacement(urlBase, envs, true)
+
+	urlBase, err := ResolveEnvironmentReplacement(rawurl, envs, true)
 	if err != nil {
 		return "", err
 	}
+
+	urlBase, err = extractFilename(urlBase)
+	if err != nil {
+		return "", err
+	}
+
 	destPath := filepath.Join(dest, urlBase)
 
 	if !filepath.IsAbs(dest) {
@@ -472,9 +478,11 @@ func getUID(userStr string) (uint32, error) {
 }
 
 // ExtractFilename extracts the filename from a URL without its query url
-func extractFilename(rawURL string) string {
-	// not necessarily return an error, due to parsing ambiguities https://pkg.go.dev/net/url#Parse
-	parsedURL, _ := url.Parse(rawURL)
+func extractFilename(rawURL string) (string, error) {
+	parsedURL, err := url.Parse(rawURL)
+	if err != nil {
+		return "", err
+	}
 	filename := filepath.Base(parsedURL.Path)
-	return filename
+	return filename, nil
 }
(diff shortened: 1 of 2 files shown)


prima101112:
@JeromeJu is this enough? or should i open new PR?

prima101112:
@JeromeJu sorry there was error linter fixed that the test should passed

pkg/util/command_util.go
```diff
@@ -469,3 +469,11 @@ func getUID(userStr string) (uint32, error) {
 	}
 	return uint32(uid), nil
 }
+
+// ExtractFilename extracts the filename from a URL without its query url
+func extractFilename(rawURL string) string {
+	// not necessarily return an error, due to parsing ambiguities https://pkg.go.dev/net/url#Parse
+	parsedURL, _ := url.Parse(rawURL)
```
aaron-prindle, line 476: I see the comment above but can you explain why we don't surface or bubble up this error?   From the url.Parse docs

```
// The url may be relative (a path, without a host) or absolute
// (starting with a scheme). Trying to parse a hostname and path
// without a scheme is invalid but may not necessarily return an
// error, due to parsing ambiguities.
```

I believe we should bubble up this error, let me know if that makes sense or there is some issue that occurs when doing this.  I believe the COPY value when in this method `URLDestinationFilepath` should have a scheme so we should respect the error

aaron-prindle:
Thanks for the PR @prima101112!  Left one comment related to bubbling up the `err` from `url.Parse`

pkg/util/command_util.go
prima101112: Hi @aaron-prindle 

Thanks, I thought that the filepath example.com/thefile.txt would return an error since it doesn't have the scheme, but it actually doesn't. Thanks for the comment. I've already fixed it and switched the execution order, with ResolveEnvironmentReplacement being executed before extractFilename, i think thats better to replace the env first.

kindly check

aaron-prindle (approved):
LGTm, thanks for the PR @prima101112!

Issue #3124, opened 2024-04-19
[TEST FLAKE] Intermittently seeing - `No such image: localhost:5000/<test-image>`


aaron-prindle:
Currently multiple PRs are seeing the below CI/CD test flake error:
```
        integration_test.go:609: Couldn't get details from image reference for (localhost:5000/docker-dockerfile_test_copyadd_chmod): Couldn't get reference to image localhost:5000/docker-dockerfile_test_copyadd_chmod from daemon: Error response from daemon: No such image: localhost:5000/docker-dockerfile_test_copyadd_chmod:latest
```

where the actual image reference is subject to change.  

I believe this is due to some changes on the GitHub action runner VM or VM image.  Currently this is a flake as some PRs are still able to get all green tests while some seem hit by this.  Still root causing this rcurrently 

Previous runs with this flake (this same PR would succeed [green tests] with no changes in later attempts):
https://github.com/GoogleContainerTools/kaniko/actions/runs/8682504136/job/23807123764#step:7:73

mschneider82:
The integration_test did not find the image which was build with `docker` command. The `BuildDockerImage()` call ignored the returned error. I opened a PR to fix this. After that lets see whats wrong with the build on the chmod issue.

mschneider82:
Solved by #3131 Close this

aaron-prindle:
This was related to DOCKER_BUILDIT not being set in some test envMaps.   In the future, we will likely need to update the test envMap for previous and new tests:

https://github.com/GoogleContainerTools/kaniko/pull/3119#issuecomment-2068932322

Closing, will open a new issue to track that work

PR #3119, opened 2024-04-18
Added --chmod for ADD and COPY commands. Fixes #2850 and #1751

mschneider82:
Fixes #2850 and #1751

**Description**

I have added --chmod feature for ADD and COPY command.

```
[36mINFO[0m[0004] ADD --chmod=644 https://xxx/test.gpg /etc/apt/trusted.gpg.d/ 
[36mINFO[0m[0004] RUN ls -la /etc/apt/trusted.gpg.d/test.gpg 
[36mINFO[0m[0007] Cmd: /bin/sh                                 
[36mINFO[0m[0007] Args: [-c ls -la /etc/apt/trusted.gpg.d/test.gpg] 
[36mINFO[0m[0007] Running: [/bin/sh -c ls -la /etc/apt/trusted.gpg.d/test.gpg] 
-rw-r--r-- 1 root root 1545 May  5  2023 /etc/apt/trusted.gpg.d/test.gpg
```

**Submitter Checklist**

These are the criteria that every PR should meet, please check them off as you
review them:

- [X] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [X] Adds integration tests if needed.

**Reviewer Notes**

- [ ] The code flow looks good. 
- [ ] Unit tests and or integration tests added.

**Release Notes**

```
- kaniko now supports setting --chmod on ADD and COPY commands.
```

PR #3131, opened 2024-04-22
fail on error on build with docker

mschneider82:
@aaron-prindle  i think this will show the error from build #3119

The build with docker command did not exit on error.


mschneider82:
should fix #3124

PR #3157, opened 2024-05-10
feat: add --skip-snapshot arg which skips all build snapshots

### LINKED ISSUE #1615: Suggestion: --snapshotMode=skip
I expect that my use-case is *well* out of scope for Kaniko, but it actually works great for it and I wanted to propose a small improvement which would be of help with it.

Basically what I've done is prepared a virtualization appliance (ovf) which on startup pulls and unpacks a container image onto its own filesystem, and then reboots into it as a VM with no containerization layer. This has the benefit of being simple for users who expect/need a VM, while also allowing lots of flexibility in how to prepare and use this image.

The core of the implementation is shipping the kaniko binary, plus a trivial Dockerfile which starts `FROM` the image I want, and then afterward has some `RUN` lines to install and set up the kernel, bootloader, etc. So then it's just an init script which basically does:

```
/kaniko/executor --dockerfile Dockerfile --no-push
/usr/bin/rm /kaniko
[kexec/reboot into new kernel]
```

But despite `--no-push`, this still takes the time afterward to crawl the whole filesystem looking for changes:

```
INFO[0489] Taking snapshot of full filesystem...        
INFO[0517] No files were changed, appending empty layer to config. No layer added to image. 
INFO[0517] Skipping push to container registry due to --no-push flag 
```

Assuming there's a valid use-case for Kaniko to create the image when passed `--no-push`, could there either be a no-op snapshot mode, or perhaps a similar `--no-snapshot` flag? Alternatively, I'd be happy to hear if there are suggestions for other, possibly lower-level tools which would handle the "Pull multi-layer image from registry, unpack it to `/`" piece of the puzzle.

wussh:
Fixes #1615 

**Description**

This PR introduces new flags, --skip-snapshot, enhancing the build process. The --skip-snapshot flag allows users to skip all snapshots during the build, potentially speeding up the process.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

```
- Added flags `--skip-snapshot` to enhance build process.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ _If you are interested in contributing to kaniko, see
       - [Flag `--skip-default-registry-fallback`](#flag---skip-default-registry-fallback)
       - [Flag `--reproducible`](#flag---reproducible)
       - [Flag `--single-snapshot`](#flag---single-snapshot)
+      - [Flag `--skip-snapshot`](#flag---skip-snapshot)
       - [Flag `--skip-push-permission-check`](#flag---skip-push-permission-check)
       - [Flag `--skip-tls-verify`](#flag---skip-tls-verify)
       - [Flag `--skip-tls-verify-pull`](#flag---skip-tls-verify-pull)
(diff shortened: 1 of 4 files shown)


zppinho:
Looking forward for this optimization

PR #3181, opened 2024-05-28
Enable pushing cache with --no-push

### LINKED ISSUE #3180: Support pushing cache even when using `--no-push`
**Actual behavior**
When using `--no-push` with `--cache=true` (and `--cache-repo=my-reg`) Kaniko will not push any cached layers to the registry.
```
INFO[0007] Pushing layer my-reg/cache/52445215:35656a0e1... to cache now 
INFO[0007] Skipping push to container registry due to --no-push flag 
```
This is undocumented and a bit unexpected given the presence of the `--no-push-cache` flag.

**Expected behavior**
Be able to not push the final image with `--no-push` but still benefit from cache being pushed.

**Additional Information**
Our use-case and workflow is that we do not push images immediately after Kaniko builds them, but instead save them as `tar`s (`--tar-path`) and then scan them with a container vulnerability scanner in the CI pipeline. Once that's passed the `tar` image is pushed to the registry. 

It seems like this used to work before https://github.com/GoogleContainerTools/kaniko/pull/3132/files was merged. It seems to fix an issue (#2752) that probably should have used the `--no-push-cache` option instead of the `--no-push` option.

Ideally the cache should be pushed as long as the `--no-push-cache` flag is not provided, regardless of the `--no-push` being present or not.
 
 | Please check if this a new feature you are proposing        | <ul><li>- [X] </li></ul>|
 | Please check if the build works in docker but not in kaniko |  N/A | 
 | Please check if this error is seen when you use `--cache` flag | <ul><li>- [X] </li></ul>|
 | Please check if your dockerfile is a multistage dockerfile | <ul><li>- [X] </li></ul>|

leosunmo:
Make sure we check --no-cache-push instead of --no-push when deciding whether to push cache or not.

Fixes #3180

**Description**
Allows for pushing to cache even if `--no-push` is enabled. This is useful in scenarios where you want to benefit (and contribute) to cache but ultimately save the image to a tar file, for example when container scanning in CI pipelines. 

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

Enable pushing cache even when `--no-push` is provided. To disable pushing cache, use `--no-push-cache`.

### DIFF
--- a/pkg/executor/push.go
+++ b/pkg/executor/push.go
@@ -398,8 +398,8 @@ func pushLayerToCache(opts *config.KanikoOptions, cacheKey string, tarPath strin
 		return errors.Wrap(err, "appending layer onto empty image")
 	}
 	cacheOpts := *opts
-	cacheOpts.TarPath = ""         // tarPath doesn't make sense for Docker layers
-	cacheOpts.NoPush = opts.NoPush // we don't want to push cached layers if no push is specified
+	cacheOpts.TarPath = ""              // tarPath doesn't make sense for Docker layers
+	cacheOpts.NoPush = opts.NoPushCache // we do not want to push cache if --no-push-cache is set.
 	cacheOpts.Destinations = []string{cache}
 	cacheOpts.InsecureRegistries = opts.InsecureRegistries
 	cacheOpts.SkipTLSVerifyRegistries = opts.SkipTLSVerifyRegistries


aaron-prindle (approved):
Thanks @leosunmo for the fix and explanation here.  LGTM

PR #3341, opened 2024-10-13
WORKDIR learned to cache it's potential output layer

### LINKED ISSUE #3340: layer created implicitly by WORKDIR is not cached
**Actual behavior**
When `WORKDIR` is called on a non-existent directory, kaniko is kind enough to create that directory for you, resulting in a layer being added. However, kaniko does not cache that layer, which means that on every invocation a completely new image is emitted from that point onwards. Inside the same stage this is non-obvious as caching mechanism still pulls, so you get a 100% cache hitrate thereafter, but the image is completely new. In multistage builds or builds that depend on the newly emitted image, this is catastrophic, as they do consider the entire image's sha when determining whether a cache is hit or not, so this will invalidate the entire cache.

Workaround is simply but silly, we just need to make sure that the directory exists before calling `WORKDIR`

```dockerfile
RUN mkdir /app
WORKDIR /app
```

**Expected behavior**

When `WORKDIR` is called on a non-existent directory, that directory is implicitly created **and the resulting layer is pushed into cache**.

**To Reproduce**

```dockerfile
FROM ubuntu:latest
WORKDIR /app
RUN echo "whatever"
```

run this build multiple times, with cache, and observe that the layers change every time

```shell
docker inspect <image>
```

 
mzihlmann:
Fixes #3340

**Description**

 When WORKDIR is called on a non-existent directory, kaniko is kind enough to create that directory for you, resulting in a layer being added. However, kaniko does not cache that layer, which means that on every invocation a completely new image is emitted from that point onwards. Inside the same stage this is non-obvious as caching mechanism still pulls, so you get a 100% cache hitrate thereafter, but the image is completely new. In multistage builds or builds that depend on the newly emitted image, this is catastrophic, as they do consider the entire image's sha when determining whether a cache is hit or not, so this will invalidate the entire cache.

So far the workaround was to ensure that the directory exists before calling `WORKDIR` to avoid creating it implicitly, as `RUN` statements can be cached:

```dockerfile 
RUN mkdir /app
WORKDIR /app
```

With this change the layer potentially created by WORKDIR is cached too in similar vein to how `RUN` statements are cached.

There is some optimization potential left on the table here, as we do sometimes know a-priori whether a layer should be created at all and always know which directory. Currently I copied the code from `RUN` to make it work, but this is suboptimal, as this code assumes no a-priori knowledge. I'm open for suggestions.

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**

- kaniko learned to cache layers created by `WORKDIR`

### DIFF
--- a/pkg/commands/commands.go
+++ b/pkg/commands/commands.go
@@ -78,7 +78,7 @@ func GetCommand(cmd instructions.Command, fileContext util.FileContext, useNewRu
 	case *instructions.EnvCommand:
 		return &EnvCommand{cmd: c}, nil
 	case *instructions.WorkdirCommand:
-		return &WorkdirCommand{cmd: c}, nil
+		return &WorkdirCommand{cmd: c, shdCache: cacheRun}, nil
 	case *instructions.AddCommand:
 		return &AddCommand{cmd: c, fileContext: fileContext}, nil
 	case *instructions.CmdCommand:
--- a/pkg/commands/workdir.go
+++ b/pkg/commands/workdir.go
@@ -17,9 +17,11 @@ limitations under the License.
 package commands
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
+	kConfig "github.com/GoogleContainerTools/kaniko/pkg/config"
 	"github.com/GoogleContainerTools/kaniko/pkg/dockerfile"
 	"github.com/pkg/errors"
 
@@ -33,6 +35,7 @@ type WorkdirCommand struct {
 	BaseCommand
 	cmd           *instructions.WorkdirCommand
 	snapshotFiles []string
+	shdCache      bool
 }
 
 // For testing
@@ -81,7 +84,11 @@ func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile
 
 // FilesToSnapshot returns the workingdir, which should have been created if it didn't already exist
 func (w *WorkdirCommand) FilesToSnapshot() []string {
-	return w.snapshotFiles
+	return nil
+}
+
+func (r *WorkdirCommand) ProvidesFilesToSnapshot() bool {
+	return false
 }
 
 // String returns some information about the command for the image config history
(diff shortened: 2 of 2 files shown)


pkg/commands/workdir.go
```diff
@@ -81,14 +84,116 @@ func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile
 
 // FilesToSnapshot returns the workingdir, which should have been created if it didn't already exist
 func (w *WorkdirCommand) FilesToSnapshot() []string {
-	return w.snapshotFiles
+	return nil
```
mzihlmann, line 87: `WORKDIR /` for example would result in an empty list, this was not cacheable on my trials. As it is not cacheable it causes cache misses every time requiring the entire file system to be unrolled. This is copied from `RUN` command, as a run command can result in no files being created ie. `RUN echo "Hello World"`, somehow if I return `nil` instead of an empty list the caching mechanism is handled graciously, I'm currently investigating why.

pkg/commands/workdir.go
```diff
@@ -81,14 +84,116 @@ func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile
...
+		extractFn: util.ExtractFile,
+	}
+}
+
 func (w *WorkdirCommand) MetadataOnly() bool {
 	return false
 }
+
+func (r *WorkdirCommand) RequiresUnpackedFS() bool {
+	return true
```
mzihlmann, line 114: `WORKDIR` required me to unpack the filesystem as otherwise the user is not known.

pkg/commands/workdir.go
```diff
@@ -81,14 +84,116 @@ func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile
...
+}
+
+func (w *WorkdirCommand) ShouldCacheOutput() bool {
+	return w.shdCache
```
mzihlmann, line 118: I was thinking whether we could optimize this instruction here to not cache the output if no files were created, I'm not sure how to tell from the cache consumer side whether a cache entry is missing or not there on purpose.

pkg/commands/workdir.go
```diff
@@ -81,14 +84,116 @@ func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile
...
+	caching
+	img            v1.Image
+	extractedFiles []string
+	cmd            *instructions.WorkdirCommand
+	extractFn      util.ExtractFunction
+}
+
+func (wr *CachingWorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile.BuildArgs) error {
+	var err error
+	logrus.Info("Cmd: workdir")
```
mzihlmann, line 132: I copied that entire block of code here, it is to ensure that even if we hit the cache we still do the metadata operation thingy and actually change the workdir. I could of course put that in a function etc. but I would actually prefer to pass the resolved directory into the cache, s.t. I can have a single line here. Not only reusing the functionality, but reusing the result.

```go
cfg.WorkingDir = wr.resolvedWorkingDir
```

pkg/commands/workdir.go
```diff
@@ -81,14 +84,116 @@ func (w *WorkdirCommand) ExecuteCommand(config *v1.Config, buildArgs *dockerfile
...
+
+	return &CachingWorkdirCommand{
+		img:       img,
+		cmd:       w.cmd,
+		extractFn: util.ExtractFile,
+	}
+}
+
 func (w *WorkdirCommand) MetadataOnly() bool {
 	return false
```
mzihlmann, line 110: sometimes calling `WORKDIR` is metadata only, there might be some optimization potential here, depending on whether we can know this a-priori.

ie. calling 
```
WORKDIR /
```
without any further context is guaranteed to be metadata only in all images to my understanding.

pkg/commands/commands.go
```diff
@@ -78,7 +78,7 @@ func GetCommand(cmd instructions.Command, fileContext util.FileContext, useNewRu
 	case *instructions.EnvCommand:
 		return &EnvCommand{cmd: c}, nil
 	case *instructions.WorkdirCommand:
-		return &WorkdirCommand{cmd: c}, nil
+		return &WorkdirCommand{cmd: c, shdCache: cacheRun}, nil
```
mzihlmann, line 81: I piggy-backed on the flag for `RUN` instructions as I don't think it's sensible to have a separate flag for each instruction.

mzihlmann: cache copy layers probably got its own flag because depending on the context the files might be huge and invalidate the purpose of a cache, which is speeding things up by downloading a layer instead of executing commands. This should not be a problem in our simple case here, so piggy-backing should be fine.

pkg/commands/workdir.go
mzihlmann: passing the result is not possible, as either the regular or the cached ExecuteCommand function is called, never both.

pkg/commands/workdir.go
mzihlmann: I was able to resolve this by allowing the cache to contain empty images, just to indicate that the command did not change any files and we are aware of this. With this modification we are able to report back an empty list and no longer need to use the FS snapshot function from RUN command.

pkg/commands/workdir.go
mzihlmann: hmmm... passing the result would be possible but requires some creative thinking. We do store the layer as an image when we push to the cache, so we could change the WORKDIR on that image and with that store that variable too. Come to think of it, labels on the cached image could be used to pass arbitrary data between cache creation and reusing stage.

thearyanahmed:
Thanks for getting this done. Do we know why it wasn't reviewed yet? We are also facing this issue. 😓

mzihlmann:
kaniko project seems to be in limbo at the moment. The workaround, as silly as it is, works flawlessly. Just make sure the folder exists before changing workdir and you're good:
```dockerfile
RUN mkdir /app
WORKDIR /app
```

mzihlmann:
If you're still looking for a solution to this you could give my fork a try https://github.com/mzihlmann/kaniko/releases/
It fixes this issue and a few more, mostly related to caching, if you have other issues you would like to see resolved please let me know. I know that this is not ideal and I hope we can get the changes merged here eventually but for now that's the best I can offer.
If you like what you see you can support me with a star, thank you 🙇

PR #3347, opened 2024-10-22
Speedup --reproducible layer stripping

### LINKED ISSUE #862: Using --reproducible loads entire image into memory

bh-tt:
Fixes #862, may mitigate #1960

The layerTime function returns a layer backed by an in-memory buffer, which means during the stripping of timestamps the entire image is loaded into memory. Additionally, it runs gzip after the layer is created, resulting in even an even larger in-memory blob.

This commit changes this method to use a temporary file instead of an in-memory buffer, and to use gzip compression while writing to this layer file, instead of compressing during read.

Fixes #862, may mitigate #1960

**Description**

This commit changes the layerTime method to use a temporary file instead of an in-memory buffer, and to use gzip compression while writing to this layer file, instead of compressing during read.

**Submitter Checklist**

review them:

Covered by existing unit tests, single method change. Integration tests not necessary.

**Reviewer Notes**

**Release Notes**

```
- improve --reproducible by not loading the entire image into memory, instead writing it to a temporary file

```

### DIFF
--- a/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/mutate.go
+++ b/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/mutate.go
@@ -16,16 +16,17 @@ package mutate
 
 import (
 	"archive/tar"
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
-	"github.com/google/go-containerregistry/internal/gzip"
+	stdgzip "compress/gzip"
+
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/match"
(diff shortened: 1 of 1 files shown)


bh-tt:
Note: the temporary files created by kaniko are not cleaned up, I'm not sure if this is necessary given that kaniko is designed to run in a container. If it is, is there some kind of lifecycle method that can be called to clean the files, preferably through the v1.Layer struct?

bh-tt:
I'm a developer working at a company so it may take some time before I can get the CLA signed...

bh-tt:
Hmm should I submit this here or in go-containerregistry?

PR #3354, opened 2024-10-30
BOOTC Fix

kfox1111:
**Description**

When extending an image intended for use with BOOTC, the image fails to install because the hardlinks have a '/' prefix. Remove it like we remove it for other files.

**Submitter Checklist**

So trivial, does it need a unit test?

**Reviewer Notes**

**Release Notes**

Fix images intended to be booted with BOOTC

### DIFF
--- a/pkg/util/tar_util.go
+++ b/pkg/util/tar_util.go
@@ -122,7 +122,8 @@ func (t *Tar) AddFileToTar(p string) error {
 
 	hardlink, linkDst := t.checkHardlink(p, i)
 	if hardlink {
-		hdr.Linkname = linkDst
+		// Docker uses no leading / in the tarball
+		hdr.Linkname = strings.TrimLeft(linkDst, "/")
 		hdr.Typeflag = tar.TypeLink
 		hdr.Size = 0
 	}


cgwalters:
I agree that not having absolute paths in the tar stream is a general best practice.

However to be clear we view this as a bug in bootc, we will ship a fix for it. Tracked here https://github.com/containers/bootc/issues/856

kfox1111:
Ping

ProbstDJakob:
I used the the following script within a pipeline:

```sh
kanikoDir="$(mktemp -d)"
git clone --branch v1.23.2 https://github.com/GoogleContainerTools/kaniko.git "$kanikoDir"
git -C "$kanikoDir" remote add fork https://github.com/kfox1111/kaniko.git
git -C "$kanikoDir" fetch fork
git -C "$kanikoDir" config user.email invalid@invalid.invalid
git -C "$kanikoDir" config user.name invalid
git -C "$kanikoDir" cherry-pick b187e9b61e2de97368580be7a2a0e9ab2a2688bf
CGO_ENABLED=0 go -C "$kanikoDir" install "$kanikoDir/cmd/executor"

[...]

"$kanikoDir/executor" --ignore-path=/var/mail --ignore-path=/var/spool/mail --context "$CI_PROJECT_DIR" --dockerfile "$CI_PROJECT_DIR/bootc/Containerfile" --destination "$CI_REGISTRY_IMAGE:$CI_COMMIT_SHA"
```

But when deploying the image with anaconda it still fails. Building the same `Containerfile` with `podman` works. In the failed anaconda session, when going to the shell and execute the command which has failed, the following is the output:

```console
# ostree container image deploy --sysroot=/mnt/sysimage --image=<image>
ERROR Performing deployment: Importing: Parsing layer blob sha256:<sha256>: : Processing tar: Filtering tar stream: Copying entry: paths in archives must have at least one component when setting path for /
```

When downloading the image, saving it as `oci-dir` and uncompressing each layer blob with `tar -tz`, the layers build with kaniko give the warning: ``tar: Removing leading `/' from member names``

BUT, `tar` does not warn about hard links any more, so the PR fixes what is written in the description, but still does not fix the whole problem `bootc` has (only a partial _BOOTC Fix_).

I guess what is missing is to remove the `/` entry and, probably optionally, remove the leading slash from all entries.

kfox1111:
Interesting. Was this a FROM scratch image? Not sure why there would be a '/' entry otherwise. Were there any layers with '/' on filenames other then the '/' entry?

ProbstDJakob:
Pretty sure there were before applying your patch, but now there are only the `/` entries but no entries with a prepended slash. Sorry, I did not look in the output again and just assumed things based on the warnings.

So to correct myself, the only thing which is missing is to remove the `/` entries.

ProbstDJakob:
`kaniko` is explicitly allowing the `/` entry in order to set permissions. So the question is, whether the OCI image allows a `/` entry or not. If the specifications allow such entries, `bootc`/`ostree-ext` should probably be patch (if not already done) otherwise `kaniko`.

https://github.com/GoogleContainerTools/kaniko/blob/b187e9b61e2de97368580be7a2a0e9ab2a2688bf/pkg/util/tar_util.go#L105-L108

PR #3386, opened 2025-01-13
Resolve a couple of issues around azure blob storage support

### LINKED ISSUE #3044: Azure Blob Storage Context invalid path
It seems that the path to get the context is wrong when using azure blob storage container

ex :

```
apiVersion: v1
kind: Pod
metadata:
  name: kaniko
spec:
  containers:
  - name: kaniko
    args: ["--dockerfile=./Dockerfile",
            "--context=https://kanikoi0skbb6wot.blob.core.windows.net/kaniko/context.tar.gz",
            "--destination=herveleclerc/kaniko-hello:0.0.1", "--verbosity=trace"]
    volumeMounts:
    - name: kaniko-secret
      mountPath: /kaniko/.docker
    env:
    - name: AZURE_STORAGE_ACCESS_KEY
      valueFrom:
        secretKeyRef:
          name: azure-storage-access-key
          key: azure-storage-access-key
  volumes:
  - name: azure-storage-access-key
    secret:
      secretName: azure-storage-access-key
  - name: kaniko-secret
    secret:
      secretName: dockerhub
      items:
      - key: .dockerconfigjson
        path: config.json
```

generate this error log

```
k logs kaniko
DEBU[0000] Getting source context from https://kanikoi0skbb6wot.blob.core.windows.net/kaniko/context.tar.gz
DEBU[0000] BaseDir /kaniko/buildcontext for file /kaniko/buildcontext/context.tar.gz does not exist. Creating.
Error: error resolving source context: HEAD https://kanikoi0skbb6wot.blob.core.windows.net/kaniko/context.tar.gz/kaniko/context.tar.gz
--------------------------------------------------------------------------------
RESPONSE 404: 404 The specified blob does not exist.
ERROR CODE: BlobNotFound
--------------------------------------------------------------------------------
Response contained no body
--------------------------------------------------------------------------------
```

the path name is in double

### LINKED ISSUE #3167: Use Azure Default Credential to avoid using keys
Azure azidentity package has [DefaultCredentials](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-defaultazurecredential) process to manage different login methods:

login

Using this, we avoid to force the use of `AZURE_STORAGE_ACCESS_KEY` env and we can use different ways of auth, including workload identity.

okhaliavka:
Fixes #3044 
Resolves #3167 

**Description**
There's currently a number of issues with Azure support in kaniko. 

#3044 is a bug that makes kaniko completely unable to download build context from Azure Blob Storage, which was probably introduced in https://github.com/GoogleContainerTools/kaniko/commit/16ed6b24287dfc9808c58b44aad324bf3eed4f5a. 
When constructing a blob storage client, kaniko is passing an entire context url (e.g. `https://accname.blob.core.windows.net/bucket/context.tar.gz`) where just an account url (`https://accname.blob.core.windows.net/`) is expected. As a result, the `bucket/context.tar.gz` part gets duplicated in the final URL and the blob doesn't get found. 

#3167 is rather a usability concern about kaniko requiring users to provide a storage access key as an env var, making it impossible to use better auth methods, such as managed workload identity. Looks like it'd be better to offload authentication to the `azidentity` library and use the [default azure credential chain](https://learn.microsoft.com/ms-my/azure/developer/go/sdk/authentication/authentication-overview#use-defaultazurecredential-in-an-application), letting users decide on the appropriate way to provide storage credentials for kaniko.
To keep things backward-compatible, the existing auth method will be kept and still attempted first. But when an access key is not provided, kaniko will fallback to the default credential chain, instead of immediately failing with an error. 

I'm aware that this project is pretty much abandoned, but I still have a tiny glimmer of hope that this contribution will be reviewed 😅 . 

**Submitter Checklist**

Azure blob storage isn't covered by unit/integration tests. Fixing that would require more time that I'm willing to put without any certainty that this PR has a chance to be reviewed and merged.
 
**Reviewer Notes**


okhaliavka:
to be split into separate prs

PR #3454, opened 2025-04-15
chore: upgrade all the dependencies to their latest versions

plumpy:
Basically this is just: `go get -u ./... && go mod tidy && go mod vendor`

But I split those out into three commits for simpler reviewing.

The I updated the version of Go used to build it and fixed one small thing for an API change in buildkit 0.15.0 (from July 2024).


JeromeJu (approved):

JeromeJu:
I've had a similar attempt for `go mod tidy` + `go mod vendor` at https://github.com/GoogleContainerTools/kaniko/pull/3440.
Currently trying to merge in many dependencies update as possible for decoupling.

One question I had was do we want to pull out the buildkit fix in its own PR?

JeromeJu:
With https://github.com/GoogleContainerTools/kaniko/pull/3450, we should have the github action workflows unblocked.

